PR #1, opened 2019-11-08
Fixing the remaining rubocop problems and getting tests in the main repo to run

sfisher:
This fixing the rubocop complaints when it looks at the whole app.  I also changed some timeouts in the tests since it seems tomake it more reliable.

The main app tests should run (though not the tests for all engines yet).

Also to get this running, all I had to do was check it out and do `symlink_config.sh` to bring in the config repo.  It assumes that config repo is on the same directory level as dryad-app.

Sorry, I haven't automated tests through travis.ci yet or fixed the deploy yet.  Those are coming after I be sure the engine tests all still run.


ryscher:
These look good, Scott. Thanks!

Also, I did get the new repo to run on my vm.

PR #2, opened 2019-11-13
Getting travis.ci working against the two sets of tests

sfisher:
This was a bit hard to get working and it seemed like no matter how the Gemfile/.gemspec dependencies were set up, the stash tests would only run against the main set of libraries.  It turns out this works ok if I add just a couple of extra libraries to the main Gemfile and makes the stash tests run faster, anyway.  I'm not sure why this is, but maybe because travis is using RVM.

I finally got tests to pass on travis (my local environment acted entirely different).


ryscher:
Thanks, Scott! This looks great, and it's wonderful to have tests running in the new setup.

PR #3, opened 2019-11-13
Changing deploy to use stash-app repo and removing extra methods

sfisher:
Please look at the earlier pull request for testing before this one since it has that code already as part of this.  If there are problems I can correct that one and re-merge it in.

This now only has to deploy one repository and checkout and symlink one config repository, so the capistrano deploy is simpler.  Also upgraded capistrano and capistrano-rails.

I tested deploying to dryad-dev with the branch change-deploy like.

`bundle exec cap development deploy BRANCH="change-deploy"`

Everything works as expected and the application seems to run fine.

We do get some messages like:

```
      02 fatal: not a git repository (or any parent up to mount point /apps)
      02 Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

People online say this is because some gems have some depenedency on .git and cause this to happen.  It doesn't seem to make a difference.


ryscher:
Ok, looks good. I was able to deploy the dev server from dryad-app. 

Thanks for getting all of this working!

PR #4, opened 2019-11-14
Updating the gems that it would allow

sfisher:
We have a bunch of security warnings on github so I'm updating to try and resolve what I can.

I've also relaxed the requirements in many of the gemfiles since they were locked to pretty specific versions (often quite old).

I still couldn't get Nokogiri to update even after relaxing that in all our own gems/engines.  Maybe it is some outside gems causing us to not be able to update.

It took some time to `bundle update` and bundle install our our engines and get things all compatible with bundler.

Got tests to pass and did a test deploy of the branch and tested especially search, file upload and submission since those were areas I thought might be more likely to break.

I left datacite-mapping at 0.3 because 0.4 requires code changes to work correctly.


ryscher:
Looks good. I'll update my other PR to work with this...

PR #5, opened 2019-11-14
Journal notification tweak

ryscher:
Modify the logic that prevents journal notifications from incorrectly changing curation status. The curators found another example today of an item that was still in_progress (not submitted by the user), yet we received a journal notification that changed the status of the dataset. So I'm reversing the logic to disallow the automatic updates unless the dataset is either submitted or peer_review.  

Also includes the updates to Gemfile.lock that were required to run the tests for stash_api on their own.


ryscher:
ok, this is now updated so it doesn't conflict with Scott's recent dependency updates

PR #6, opened 2019-11-15
minor changes for repopulating database with stats

sfisher:
These make the script work (dash instead of underscore) and repopulate the database with stats and don't erase stats that get put into the database.


ryscher:
This looks good. Hooray for stats!

PR #7, opened 2019-11-15
Fixes logging of incomplete items

sfisher:
Doen't log if missing required fields.  Also escapes control characters in text fields to eliminate things like \t, \n, \r, bell, backspace, etc.

Added tests which didn't exist before and were a little painful to add.


sfisher:
I'll look into why this is failing.  Weird.

ryscher:
Ok, this looks good now. Thanks!

PR #9, opened 2019-11-19
Changing publisher to journal name

sfisher:
As in https://github.com/CDL-Dryad/dryad-product-roadmap/issues/521

Also added basic test to test that it is showing publisher name since there were not tests for this before, I think.


ryscher:
Looks good, tests fine on dev.

PR #12, opened 2019-11-22
Getting the UI library working again

sfisher:
It added some things to the UI library when I installed with npm.

I also added to rubocop to ignore the node modules since it was going haywire trying to fix them.

ps.  Got the ui library working from stash/stash_engine with these things or something similar.

```
npm install -g bower
npm install
```


ryscher:
Looks good. Thanks, Scott!

PR #14, opened 2019-12-02
Fix misc internal errors

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/558

These internal errors are typically caused by crawlers making poorly-formed requests. Nevertheless, they result in emails to the developers, which are undesirable.

Some sample calls that are fixed with this PR:

`curl -i -X GET https://datadryad.org/stash/dataset/doi:10.6071/M3GT0P -H "Accept: application/json"`

`curl -i -X GET https://datadryad.org/stash_datacite/geolocation_boxes/boxes_coordinates?resource_id=31849 -H "Accept: text/html"`

`curl -i -X GET https://datadryad.org/stash_datacite/affiliations/autocomplete?term=Qatar+- -H "Accept: application/json"`

### DIFF
--- a/stash/stash_datacite/app/controllers/stash_datacite/affiliations_controller.rb
+++ b/stash/stash_datacite/app/controllers/stash_datacite/affiliations_controller.rb
@@ -10,7 +10,7 @@ module StashDatacite
       partial_term = params['term']
       return if partial_term.blank?
       # clean the partial_term of unwanted characters so it doesn't cause errors when calling the ROR API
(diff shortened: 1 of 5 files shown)


ryscher:
Note that this PR does *not* fix the InvalidURIError caused by malformed URLs like `https://datadryad.org/resource/doi:%2010.5061/dryad.8682` -- this issue appears to be within either Rails or Passenger, not directly within our code.

PR #15, opened 2019-12-03
Updating Google analytics tag and only insert it if set in config

sfisher:
I've updated the google analytics to have the latest code from google and only load it if the config is set with the analytics ID.

BTW, also added a PR to the config repo since needed to add IDs for google in the config.


ryscher:
Ok, looks great, and I was able to see myself using it with the Analytics "realtime" mode.

PR #17, opened 2019-12-05
Change headings within included files of dataset page

JoelCDL:


JoelCDL:
Note that the headings under the heading "Usage Notes" on the dataset page should be <h3> (they're currently <h4>) but I couldn't find where they're located in the Dryad app.

PR #18, opened 2019-12-06
Adding scripts and information so we can use it again later if needed

sfisher:
Just adding these one-off scripts to our repository so we can play with them again later if we need to and not have to re-invent them if we find them useful later.


ryscher:
looks like get_timings has a lot of rubocop issues....

ryscher:
Thanks!

PR #20, opened 2019-12-06
A11y upload alert

JoelCDL:
So far, this alerts a user using a screen reader "Upload complete", shortly after it appears in the UI. Not able to alert a user that a file has been removed yet because this functionality doesn't exist ...

There needs to be a new <div> created for these status notes when a user uploads or removes files, including logic that renders text that says, "File successfully removed", or something similar. This would require a refactor.


ryscher:
This looks fine and doesn't change any existing functionality. Thanks Joel!

PR #21, opened 2019-12-06
add focus style to choose files button

JoelCDL:


ryscher:
Looks good. The new focus outline is much better than not having one.

PR #22, opened 2019-12-10
Dash datacite targets

sfisher:
See ticket 600.  This queries to get items embargoed or published and from non-Dryad collections (ie Dash) and then updates the metadata for them with DataCite/EZID.  It will update the target URLs and also the ROR ids that Ryan added.

It retries up to 10 times and doesn't update unless there is an embargoed/published item that shows metadata.


marisastrong:
Suggestion to monitor the access logs to ensure traffic is going through the old Dash production server.  Then we can shut down the server.

ryscher:
Looks good. Thanks for adding the retry code -- I'm sure that will be needed.

PR #23, opened 2019-12-12
Counter log fixes

sfisher:
Simple changes not to output if missing either publication date or publication year.  Counter requires them.


ryscher:
Looks good. Always great to protect against bad data.

PR #26, opened 2019-12-13
Dryad doi updater

sfisher:
This added a slightly different query and more error handling inside the task to continue after errors and logging of the errors.  This only updates the embargoed/public DOIs from the Dryad tenant.


ryscher:
Thanks Scott! It's great to get all of these updated.

PR #31, opened 2019-12-20
Allow non-secure logins for docker

sfisher:
This is to allow Terry's docker container to login insecurely.  Modifying the hack we had to put in to get https working correctly because of issues with the Apache reverse proxy in normal operation with login/omniauth .


ryscher:
This looks good. Can't wait to see docker in action.

PR #36, opened 2019-12-20
Accessibility fixes for "Describe" page

JoelCDL:


sfisher:
Thanks for looking at this Ryan.  I was just going to look.

sfisher:
It looks like this is failing because of the weird date things for caching the week's stats.

Probably the easiest thing to do is merge pull request 39 into master which has a fix for that testing problem if you think it's ready.  Then merge master into this branch and I bet it will get tests passing.

PR #37, opened 2019-12-23
Hacky blacklist

sfisher:
Hey Ryan,

This is a hack into our before-filter for setting up downloads.  We had an IP address from India overloading the server today with huge download requests and hitting the same items many times (in excess of 64 times for the full version download of 400+ GB).  I needed to add another IP address to blacklist.  We really need some way to block people when they become a problem without changing code (or puppet or AWS) when we can't control those settings well or quickly right now.

If there is the file `uploads/blacklist.txt` then it reads it and splits the lines.  It ignores any blank lines or ones that start with `#`.  If the remote_ip starts with something on a line in the file then it blocks that IP address and returns a head response of 429 instead of doing a download.

I tested and downloads still work for me and it seems to have blocked the Indian IP Address that some other people besides us also reported as problematic or malicious, also, that I added to the blacklist file.


ryscher:
This looks fine, and has been working well in production.

PR #39, opened 2020-01-04
Limit simultaneous downloads by ip address

sfisher:
See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/613 .

I've added testing where I can but this is a bit difficult to test because of concurrency.

This is a good dataset for testing against since it has a large file/dataset that takes a while to download  https://dryad-dev.cdlib.org/stash/dataset/doi:10.7959/dryad.8cz8w9j1 .

Essentially this logs downloads to the stash_engine_download_histories table.  Create a record when a download starts.  state is downloading or finished.  Change to finished when completed.  The difference between created_at and updated_at times is the download length.  Write ip_address and user_agent to table and I hope this will make it easier for us to see the download activity and see what is going on instead of searching through logs.

Also added rake task and the other config repo adds it to daily cron to cleanup older downloads than 60 days and fix suspicious states that never go to finished (which mostly shouldn't happen unless there is an ungraceful shutdown of a crashed server or something).

Added tests .


ryscher:
Thanks, Scott! This looks great, and works well in my testing.

I corrected one minor rubocop issue.

ryscher:
Not merging yet, since there is still a testing issue with counter_stat_spec.

sfisher:
Woot! Got tests to pass.  It was some weird thing where it was calculating the year and week strangely (2020-week 52) instead of (2019-52).

PR #40, opened 2020-01-04
In-progress items: email reminders

ryscher:
For step 2 of https://github.com/CDL-Dryad/dryad-product-roadmap/issues/568

(Note this subsumes PR #38)

1. Adds a rake task that will email users a reminder about items that have been left in_progress. The email will only be sent once for each resource, regardless of how many times the task is run, using the text of the associated Curation Activity to determine whether the email has been previously sent.
2. The Curation Activity for this task, along with the other Curation Activities that are created by cron tasks, have been set to user_id = 0, which now displays in the UI as "System", to indicate that the activity was not performed by a human user.

### DIFF
--- a/stash/stash_engine/app/mailers/stash_engine/user_mailer.rb
+++ b/stash/stash_engine/app/mailers/stash_engine/user_mailer.rb
@@ -62,7 +62,7 @@ module StashEngine
     end
 
     def in_progress_reminder(resource)
-      warn("Unable to send in_progress_reminder; nil resource") unless resource.present?
+      warn('Unable to send in_progress_reminder; nil resource') unless resource.present?
       return unless resource.present?
       user = resource.authors.first || resource.user
       return unless user.present? && user_email(user).present?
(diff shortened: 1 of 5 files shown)


stash/stash_engine/app/mailers/stash_engine/user_mailer.rb
```diff
@@ -61,6 +61,17 @@ def error_report(resource, error)
            subject: "#{rails_env} Submitting dataset \"#{@resource.title}\" (doi:#{@resource.identifier_value}) failed")
     end
 
+    def in_progress_reminder(resource)
+      warn('Unable to send in_progress_reminder; nil resource') unless resource.present?
```
sfisher: I'm not sure if this is supposed to be `logger.warn('...message...')`.  It looks like there is a kernel method called `warn`, also, but not sure if it logs it.

ryscher: You're right. It doesn't log the same way; it sends the message to stderr rather than a log. There were similar calls elsewhere in user_mailer. But I don't see much use in forcing these messages to stderr when nearly everything else goes to the log, so I'll change them all to use the logger.

PR #42, opened 2020-01-10
Minor changes for weekly processing of Counter files (see also dryad-config)

sfisher:
Very small changes here.  Sync output, comment, task to clear rows from cache table

PS.  You can test on stage by running the bash script in the other repo.


ryscher:
Ok, great.

PR #44, opened 2020-01-10
Gem updates hoping to improve security 2020 01

sfisher:
I manually set bundler to use a later version of Nokogiri, but it reset the versions back to older ones because we have a number of old gems that we use that require older, specific versions.  It doesn't really seem like there is any way around some of these problems until some of the gems we use update to use the newer versions of nokogiri, also.

In any case, we got a few minor updates of other gems.  :-/


ryscher:
Whee version changes!

PR #48, opened 2020-01-17
Curator search allows better control and fixes broken items

sfisher:
These are the changes here and how to test. (I'm adding a cron in the other repository, also.)

- When searching for keywords it orders by relevance ( tf/idf ) after a search is done until another sort order is chosen.
- Added *All terms* checkbox which makes every word typed in the box required (AND query).  It also makes matches exactly so useful for things like DOI search so it doesn't tokenize at the slash, dot or other weird places and match a lot of items.
- *All terms* also allows MySQL specialized search syntax and reverts to that if people use some special characters like the plus sign. (See note at bottom of the page)
- Clear search and reset filters fixed.
- Filter by for tenant fixed and works.
- Journal name autocomplete/filter fixed.  It allows limiting to a subset of journals within the current results as Brian implemented it.


ryscher:
Looks good, especially the promotion of "special people" to have full control over the search. Thanks!!!

PR #49, opened 2020-01-17
Autocomplete: non-controlled journal names

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/627 and the first (journal) part of https://github.com/CDL-Dryad/dryad-product-roadmap/issues/622.

This improves the behavior when a user types a value in the submission system's journal autocomplete box and *doesn't* select a value from the dropdown list. Previously, the typed name would be lost. Now, the name is saved to the database, with an asterisk appended. When the page is later viewed, that same name is displayed. However, if the user had previously selected a controlled value from the list, which was internally stored with the ISSN, the name of that journal takes precedence.

### DIFF
--- a/stash/stash_datacite/app/assets/javascripts/stash_datacite/publications.js
+++ b/stash/stash_datacite/app/assets/javascripts/stash_datacite/publications.js
@@ -37,6 +37,10 @@ function loadPublications() {
 		    });
 		},
                 source: function (request, response) {
+		    // save the user's typed request in the database with an asterisk, in case they don't click on an autocomplete result
+		    document.getElementById("internal_datum_publication_name").value = request.term + "*";
+		    var form = $(this.form);
+                    $(form).trigger('submit.rails');
 		    $.ajax({
 			url: "/stash_datacite/publications/autocomplete",
                         dataType: "json",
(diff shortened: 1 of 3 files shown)


stash/stash_datacite/app/assets/javascripts/stash_datacite/publications.js
```diff
@@ -19,24 +19,30 @@ function loadPublications() {
...
-                    $.ajax({
-                        url: "/stash_datacite/publications/issn/"+ document.getElementById("internal_datum_publication").value,
-                        dataType: "json",
-                        success: function( data ) {
-                            document.getElementById("internal_datum_publication").value = ""
-                            if (data.fullName != null) {
-				document.getElementById("internal_datum_publication").value = data.fullName;
+		    if(document.getElementById("internal_datum_publication_issn").value){
+			$.ajax({
+                            url: "/stash_datacite/publications/issn/"+ document.getElementById("internal_datum_publication_issn").value,
```
sfisher: I think we're always loading jQuery, so you can probably use stuff like `$('#<id>')` or `$('.<class>')` to select elements in a more succinct way.  I think also when you get them that way, they have more jquery methods attached and may be more cross-browser compatible since one of the things jQuery does is make calls more regular and less quirky across browsers.  Though this works and is fine for a simple case like this and getElementById probably works everywhere.

PR #50, opened 2020-01-22
Changes to pre-populate usage by script and only use from the database

sfisher:
This is part of https://github.com/CDL-Dryad/dryad-product-roadmap/issues/617 .

I'm currently pre-populating stats for development and this script will populate the stats into our database.

It adds some fields to the database such as a timestamp for the last time citations were updated.  Also adds in a foreign key for citations and the identifier table so they are easier to link together.

Script will run through and pre-populate all the citation stuff from external APIs which are already embedded in some of our method calls that formerly did it live and cached.

Before we do it on production, will need to deploy code (or at least update the database migration there) because of database additions.


ryscher:
Ok, looks good, works for me. Thanks!!!

PR #51, opened 2020-01-24
Showing stats in the curation UI

sfisher:
This is for the second part of that ticket.  Showing the stats with ajax in the curation place when curators click the little graph.

The capybara tests went haywire and wouldn't pass headless until I redid the whole configuration.  Fun.

This builds on the previous code from PR 50.


ryscher:
This is great! Thanks for getting all the silly tests working.

PR #52, opened 2020-01-24
Output CSV stats with views, downloads and citation numbers

sfisher:
This is basically the last half of this ticket https://github.com/CDL-Dryad/dryad-product-roadmap/issues/617 .

This adds views, downloads and citation count to the CSV from underneath the curation table and query should be optimized so it doesn't take forever to run.  This builds on the previous 2 PRs code, so maybe accept them in order.


ryscher:
Looks good, and works well on dev.

PR #53, opened 2020-01-25
Clear the DB of counter stats by zeroing rather than deleting

sfisher:
We should leave the rest of the record intact since it has citation numbers in it.

The counter stats need to be built up from zero again each week because it's difficult to track what has gone into them already so start from zero and munch those json files.

This has code from the previous 3 pull requests 50-52 since they build on each other.


ryscher:
Yes, great.

PR #54, opened 2020-01-27
Autocomplete: non-controlled affiliations

ryscher:
For the second (affiliation) part of CDL-Dryad/dryad-product-roadmap#622.

When a user types in an affiliation name, but does not select a name from the autocomplete, save the typed name with an asterisk appended. This allows curators to identify the non-controlled values and act accordingly.

### DIFF
--- a/stash/stash_datacite/app/models/stash_datacite/affiliation.rb
+++ b/stash/stash_datacite/app/models/stash_datacite/affiliation.rb
@@ -43,12 +43,12 @@ module StashDatacite
     # asterisk on the name, so we know it is not associated with ROR.
     def self.from_long_name(long_name)
       return nil if long_name.blank?
-      
+
       affil = find_or_initialize_by_long_name(long_name)
       return affil if affil.ror_id.present?
 
       # The record didn't exist in ROR so return it with the asterisk
-      affil.long_name = affil.long_name + "*"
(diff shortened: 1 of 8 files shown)


stash/stash_datacite/app/models/stash_datacite/affiliation.rb
```diff
@@ -38,28 +38,31 @@ def fee_waiver_countries
...
 
-      ror_org = find_by_ror_long_name(long_name)
-      # The record didn't exist in ROR so just return as is
-      return affil if ror_org.blank?
-
-      # Otherwise use the ROR id and long_name
-      affil = find_or_initialize_by_long_name(ror_org[:name])
-      affil.ror_id = ror_org[:id]
-      affil
+      Affiliation.new(long_name: long_name + '*')
```
sfisher: I don't know if `long_name` has a chance of being nil, but if so you can keep it from having some errors with code like this `Affiliation.new(long_name: "#{long_name}*")`  Because the interpolation version will substitute an empty string for nil.  While the + method will throw an error when concatenating nil with a string.

ryscher: Good idea. Even though there is a test for `long_name` being nil above, I prefer this construction.

PR #56, opened 2020-01-28
Allow some datasets to avoid payment for funders

sfisher:
Allow some funders to take on payment.

There is a corresponding commit to add Chan Zuck to the private config and it can be tested manually with that one.  Also added UI tests.

### DIFF
--- a/stash/stash_engine/app/models/stash_engine/identifier.rb
+++ b/stash/stash_engine/app/models/stash_engine/identifier.rb
@@ -222,8 +222,9 @@ module StashEngine
         (!submitter_affiliation.present? || !submitter_affiliation.fee_waivered?)
     end
 
+    # rubocop:disable Metrics/CyclomaticComplexity
     def record_payment
-      return if payment_type.present?
+      return if payment_type.present? && payment_type != 'unknown'
       if journal_will_pay?
         self.payment_type = 'journal-' + publication_data('paymentPlanType')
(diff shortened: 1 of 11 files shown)


stash/stash_engine/app/models/stash_engine/identifier.rb
```diff
@@ -234,11 +234,16 @@ def record_payment
       elsif submitter_affiliation&.fee_waivered?
         self.payment_type = 'waiver'
         self.payment_id = submitter_affiliation.country_name
+      elsif funder_will_pay?
+        contrib = funder_payment_info
+        self.payment_type = "funder:#{contrib.contributor_name}"
```
ryscher: To keep these payment types more parallel, and make the reports easier to process, I think it would be better to keep the `payment_type` as just "funder", and fill the `payment_id` with `#{contrib.contributor_name}:#{contrib.award_number}`

sfisher: I made the payment type just funder.

ryscher:
Wooo, it passed!

ryscher:
I've reviewed the payment code a bit more and I'm convinced that point 2 above is not an issue. However, just to be doubly-certain for the future, I made a slight improvement to the payment logic in #57.

sfisher:
Thank you, @ryscher

PR #57, opened 2020-01-29
Improve logic for sending an invoice

ryscher:
Updating the logic for determining whether to send an invoice. This shouldn't change any existing behavior, but it's a little more future-proof based on to the way usage of `payment_id` and `payment_type` has evolved.


ryscher:
This is not urgent... it can wait for a future sprint.

PR #64, opened 2020-02-13
Don't send whitespace to crossref

ryscher:
People sometimes add whitespace into the DOI lookup field on the submission form. Filter this out before it goes through Serrano so it doesn't cause an error.


sfisher:
Should work fine as long as the doi isn't nil.  Thanks, Ryan.

PR #66, opened 2020-02-15
Apply journal's regex when saving a manuscript number to internal_data

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/507 and https://github.com/CDL-Dryad/dryad-product-roadmap/issues/621

When the user enters a manuscript number in the submission form, first look up the regex for the associated journal, and apply it to get the "canonical" manuscript number, removing version numbers and other meaningless metadata that journals add into their manuscript numbers. 

This allows communication with the journal module to always be based on the canonical manuscript numbers, ensuring that the numbers match more frequently.

To test, enter a journal name and manuscript number in the submission system, then verify that the manuscript number has no extraneous information when saved in `stash_engine_internal_data`. Some sample journals and manuscript numbers:
- Royal Society Open Science, RSOS-191815.R1
- eLife, 11-09-2019-RA-eLife-51810R1 (should remove the date, "RA" article type, and the version number)

### DIFF
--- a/stash/stash_datacite/app/controllers/stash_datacite/publications_controller.rb
+++ b/stash/stash_datacite/app/controllers/stash_datacite/publications_controller.rb
@@ -69,6 +69,7 @@ module StashDatacite
       regex = @se_id.journal_manuscript_regex
       return msid if regex.blank?
       logger.debug("- found regex /#{regex}/")
+      return msid if msid.match(regex).blank?
       logger.debug("- after regex applied: #{msid.match(regex)[1]}")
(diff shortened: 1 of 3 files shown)


sfisher:
Hi Ryan.  I pulled this down to my local machine and tested with your examples and it wouldn't fill things in.  Is that expected or what am I doing wrong?

I think things look pretty good and thanks for the tests.

ryscher:
Good catch! This should fix it.

PR #67, opened 2020-02-17
Force WebMock to be enabled any time mock_ror is called

ryscher:
This may address the intermittent testing failures for ROR functionality.


ryscher:
Grr, it didn't help. Closing.

PR #72, opened 2020-02-19
Updating GEM bundles in all engines and app

sfisher:
I'm hoping to mitigate some security issues, but it seems like some gems we rely on (DataCite) may not do their security updates and lock gems at old versions.


ryscher:
Ok, looks good. Thanks for re-running the tests.

PR #77, opened 2020-02-25
Merritt file download

sfisher:
This is a basic pull request for downloading files from Merritt to the local (EFS on servers) file system before we begin packaging and sending.

It's initialized with a resource and then you tell it to download a file.  It returns hash of success: and error: (if error).  I think we should check status after each operation and I don't care too much why something fails but mostly want to be able to log an error so we can write it somewhere to check and maybe resubmit things later.  The errors may range from network/dns errors to 404 or 500 errors.

It downloads file in chunks to avoid high memory usage or problems if reading the whole thing into memory before writing.

I added tests for the class.  To really test against development, it needs to be done on our development server because Merritt stage is firewalled off which makes it annoying.

I'll add some more PRs for downloading all files before doing the upload to zenodo bits.


ryscher:
This looks good. And it's great to have all of the error checking. Thanks!

PR #78, opened 2020-02-26
Allow affiliation long_name to be duplicated

ryscher:
Addresses https://github.com/CDL-Dryad/dryad-product-roadmap/issues/652

Since ROR doesn't guarantee the uniqueness of long_name, we cannot enforce it to be unique within our database.

### DIFF
--- a/stash/stash_datacite/app/models/stash_datacite/affiliation.rb
+++ b/stash/stash_datacite/app/models/stash_datacite/affiliation.rb
@@ -11,7 +11,7 @@ module StashDatacite
     has_and_belongs_to_many :authors, class_name: 'StashEngine::Author', join_table: 'dcs_affiliations_authors'
     has_and_belongs_to_many :contributors, class_name: 'StashDatacite::Contributor'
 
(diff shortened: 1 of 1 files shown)


sfisher:
Hi Ryan,

Do you know why this was set to unique?  Is the hope to only have one entry in our dcs_affiliations table for each institution?  IDK which of us or @briri set this to unique but it seems like it's no longer a good constraint because of duplicates in ROR itself.  I thought about ror_id being unique but also probably not good if people can just type in things and are not really required to have a ROR id.

I'm a little paranoid about this since I figure there was some intent behind this constraint by someone, such as reducing the number of duplicate entries or garbage in this table and reusing item through the join table (dcs_affiliations_authors).

As long as this isn't going to result in some unforeseen or undesirable consequence, I'm fine removing this constraint.

ryscher:
The original design was for the affiliations to be unique and be associated with authors in a many-to-many relationship, but Brian changed this to be many-to-one because we don't have a UI to support many-to-many. The `create` code still enforces uniqueness of the ror_id, but it's not enforced by a validation.

PR #80, opened 2020-02-28
Fixed leap-year calculation problems

sfisher:
Just adding 366 days for a date over a year out instead of the ugly date creating stuff that was there before.


ryscher:
Looks good, thanks for tracking it down!

PR #91, opened 2020-03-20
Patch for CVE-2020-5267

sfisher:
There is a possible XSS vulnerability in ActionView's JavaScript literal
escape helpers.  Views that use the `j` or `escape_javascript` methods
may be susceptible to XSS attacks.

Versions Affected:  All.
Not affected:       None.
Fixed Versions:     6.0.2.2, 5.2.4.2

### Impact

There is a possible XSS vulnerability in the `j` and `escape_javascript`
methods in ActionView.  These methods are used for escaping JavaScript string
literals.  Impacted code will look something like this:

```erb
<script>let a = `<%= j unknown_input %>`</script>
```

or

```erb
<script>let a = `<%= escape_javascript unknown_input %>`</script>
```


ryscher:
Looks good. We need to start thinking about a Rails upgrade....

PR #93, opened 2020-03-24
Reporting updates

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/663

Add more fields to existing reports. For the administrative screen report, adds the Journal Name. For the shopping cart report, adds:
- Date Submitted
- Dataset Size
- Institution Name


stash/stash_engine/app/models/stash_engine/resource.rb
```diff
@@ -355,6 +355,15 @@ def current_curation_status
       current_curation_activity.status
     end
 
+    # Date on which the user first submitted this dataset
+    # (for peer_review datasets, the date at which it came out of peer_review)
+    def submitted_date
+      curation_activities.order(:id).each do |ca|
```
sfisher: Couldn't you just turn this into a query instead of having to iterate through? `curation_activities.order(:id, :desc).where(status: 'submitted').first` ?  I think this might return a nil if none are present, otherwise you could directly get the record and return the date and have sql find the record for you.

ryscher:
Ok, updated with the new query and added an extra test.

sfisher:
Looks good.  Thanks, Ryan.

PR #95, opened 2020-03-26
Zenodo third copy

sfisher:
This isn't hooked into anything yet to happen automatically and it will eventually need some more fixes once Zenodo changes things on their end.

You can test manually with a rails console (from the dev server because Merritt Express is firewalled off).

```
require 'stash/zenodo_replicate'
@resource = StashEngine::Resource.find(499)
StashEngine::ZenodoThirdCopy.create(state: 'enqueued', identifier_id: @resource.identifier_id, resource_id: @resource.id)
zr = Stash::ZenodoReplicate::Resource.new(resource: @resource)
zr.add_to_zenodo
```

Afterward you can check the deposition_id in the zenodo_third_copy table for the resource_id and look at it on Zenodo if you wish.  That table has the deposition_Id in zenodo.

Api urls like `https://sandbox.zenodo.org/api/deposit/depositions/<deposition_id>?access_token=<token>`

URLs like `https://sandbox.zenodo.org/record/<deposition_id>`


ryscher:
This looks good overall. A couple of issues:
1) The use of the term "third copy" may become confusing as the architecture of Dryad changes.
2) Large files may become a problem. I tried to transfer a 5GB file. After it had processed for about 30 minutes, I received "error writing to socket: Connection timed out". Not sure whether this was a network issue or Zenodo's size limit.

sfisher:
Thanks Ryan.  Changed ZenodoThirdCopy to ZenodoCopy.  We'll have to revisit and change the download to use the presigned urls and also add some more process and a few other things once Zenodo has made the changes for both size and also to allow us to reopen and edit again.

I can up our timeout for the connection if needed, fi that is what is causing the problem.

ryscher:
Great, thanks for the updates!

PR #97, opened 2020-03-30
This changes the download for individual files to use Merritt presigned urls

sfisher:
When you click the download in the UI or go to a link in the API it obtains and gives a redirect to the presigned URL.


ryscher:
Looks good, thanks Scott!

PR #100, opened 2020-04-01
Make Merritt URLs download from presigned URLs

sfisher:
This should be pretty much the same procedure as the previous PR if you want to test manually ( https://github.com/CDL-Dryad/dryad-app/pull/95 ), but now it's using S3 instead.  It adds another query (Merritt) in addition to S3 to download.

I also had to refactor and redo lots of mocks/stubs and testing.


ryscher:
Looks good, and works when I test on dev. Thanks for all the updates!

PR #104, opened 2020-04-06
Add troubleshooting note about timeouts on review page

ryscher:


ryscher:
Yes, restarting the journal module is the best solution.

PR #105, opened 2020-04-06
Prevent out-of-order submission and hook into publishing: ticket #688

sfisher:
This hooks copying to zenodo into publication and also prevents out of order submission (errors).  This also includes code from the previous branch (PR) since this is building on top of it. It also fixes a bunch of tests that broke with the publication callback.

To test:

Start the delayed_job daemon.

```RAILS_ENV=local_dev bin/delayed_job start```

Create a new dataset, add files and then after it is submitted go to the curation area.  Choose to publish the dataset.  At that point it will be enqueued in delayed_job for replication.

You can monitor the state of it in the zenodo_copies table.   After it finishes you can look it up on the zenodo sandbox site such as urls like https://sandbox.zenodo.org/record/520032 (just substitute your deposition_id from the database at the end of the URL).

### DIFF
--- a/stash/stash_engine/spec/db/stash_engine/curation_activity_spec.rb
+++ b/stash/stash_engine/spec/db/stash_engine/curation_activity_spec.rb
@@ -76,7 +76,6 @@ module StashEngine
         allow_any_instance_of(Stash::Payments::Invoicer).to receive(:charge_user_via_invoice).and_return(true)
         allow_any_instance_of(StashEngine::Resource).to receive(:contributors).and_return([])
         allow_any_instance_of(StashEngine::CurationActivity).to receive(:copy_to_zenodo).and_return(true)
-        allow_any_instance_of(StashEngine::CurationActivity).to receive(:copy_to_zenodo).and_return(true)
       end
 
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: stash/stash_engine/spec/db/stash_engine/curation_activity_spec.rb [hunk 1]
```diff
@@ -73,6 +75,8 @@ module StashEngine
         allow_any_instance_of(Stash::Payments::Invoicer).to receive(:new).and_return(true)
         allow_any_instance_of(Stash::Payments::Invoicer).to receive(:charge_user_via_invoice).and_return(true)
         allow_any_instance_of(StashEngine::Resource).to receive(:contributors).and_return([])
+        allow_any_instance_of(StashEngine::CurationActivity).to receive(:copy_to_zenodo).and_return(true)
+        allow_any_instance_of(StashEngine::CurationActivity).to receive(:copy_to_zenodo).and_return(true)
```
ryscher, line 79: These lines are duplicated.

ryscher:

PR #107, opened 2020-04-08
Zenodo "versioning" for re-editing zenodo copy if published multiple times

sfisher:
This builds on previous pull requests for zenodo.  It allows a version to be published more than once and to make the changes to re-edit the zenodo copy as it changes.  See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/695 for cases you can test.

To test, basically create a dataset and then go into curation and publish it.  Then make more changes after publication and publish it again.

This requires the delayed_job daemon to be running.  I'll see if I can deploy this and start this all on dev so it can be tested there.

I also had to make changes to a bunch of tests to ignore the callbacks for this publication event for the 3rd copy when they were tests for unrelated things.

### DIFF
--- a/stash/stash_engine/lib/stash/zenodo_replicate/metadata_generator.rb
+++ b/stash/stash_engine/lib/stash/zenodo_replicate/metadata_generator.rb
@@ -12,7 +12,7 @@ module Stash
       def metadata
         out_hash = {}.with_indifferent_access
         # took out the DOI since I don't want to use it in zenodo because of versioning problems
-        %i[upload_type publication_date title creators description access_right license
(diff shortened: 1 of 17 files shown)


stash/stash_engine/lib/stash/zenodo_replicate/metadata_generator.rb
```diff
@@ -12,7 +12,7 @@ def initialize(resource:)
       def metadata
         out_hash = {}.with_indifferent_access
         # took out the DOI since I don't want to use it in zenodo because of versioning problems
```
ryscher, line 14: Looks like this comment is no longer relevant

ryscher:

PR #108, opened 2020-04-08
For ticket #690  -- retry a few times over a few days on errors

sfisher:
This builds on the previous pull requests and goes along with one in the config repository.

This adds a rake task for retrying those things whose retry limit is 3 or less (in database table).

Added to the config repository to run daily in case we were just unlucky or there were temporary problems so we don't have to spend a long time babysitting errors and I hope some will be sorted out themselves.


ryscher:

PR #109, opened 2020-04-10
Changes to support draining and resuming long jobs

sfisher:
I have some scripts for dealing with long-running jobs and we can put them on the servers to call these rake tasks.

I have some of my sample scripts in the other repo, but Mark or Ashley will need to puppetize and deploy them.


ryscher:
These look good. Thanks for adding them!

PR #111, opened 2020-04-14
Move journal metadata into the DB

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/697

This PR includes:
1) Creation of a new model, StashEngine::Journal
2) Provides a rake task that reads journal metadata from the Dryad v1 API, and uses this information to populate `stash_engine_journals`
3) Refactors most of the journal-related functionality from the Identifier model into the Journal model.

The Journal model itself is very simplistic. Although it is technically writable, it is currently treated as if it were still a read-only API. The only thing that writes values into the model is the rake task that initializes it. The "writable" fields are still in `stash_engine_internal_data`, associated with Identifier objects. 

Now that the metadata is directly accessible by Rails, I was able to remove a lot of test framework that was stubbing out the API and testing trivial GETs. These are now handled the ActiveRecord automatic getters, so I didn't add new tests to replace them.

To migrate the journal metadata:
1) Run the database migration
2) Copy the file `https://v1.datadryad.org/themes/Mirage/lib/journalISSNs.txt` into the directory where you normally run rake tasks.
3) Run the journal-migration task: `rake dryad_migration:migrate_journal_metadata`
You should see all of the relevant metadata appear in the database, and the updated code now uses these database entries instead of the API. 

There is no new functionality, but all journal functions should still operate as normal, including:
- autocomplete in the submission system
- metadata import (note: this still uses the v1 API)
- payment

### DIFF
--- a/stash/stash_datacite/app/assets/javascripts/stash_datacite/publications.js
+++ b/stash/stash_datacite/app/assets/javascripts/stash_datacite/publications.js
@@ -29,8 +29,8 @@ function loadPublications() {
                             dataType: "json",
                             success: function( data ) {
 				$("#internal_datum_publication").val("")
-				if (data.fullName != null) {
-				    $("#internal_datum_publication").val(data.fullName);
+				if (data.title != null) {
+				    $("#internal_datum_publication").val(data.title);
 				}
 			    }
 			});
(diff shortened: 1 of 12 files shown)


ryscher:
I know this introduces another "bubble_up_exact_matches", and it is time to refactor them out into a utility class.... will do that in a future PR.

sfisher:
This looks good and I tested out the rake task.  Thanks Ryan.

PR #114, opened 2020-04-15
Update journal documentation after journal metadata moved to the database

ryscher:


marisastrong:
Hi @ryscher i was confused by the  note - Journal metadata is still available through the v1 journal module's API,
but **this feature is deprecated**. 

is it available or is it deprecated?  thanks for clarifying!

ryscher:
The v1 API is available -- you can still get metadata from https://v1.datadryad.org/api/v1/journals/1557-7015. AND it is deprecated -- it is no longer the preferred way to access this metadata; no new code should use it.

PR #116, opened 2020-04-22
Functioning UI for software uploads

sfisher:
This PR is for duplicating our upload and URL validation UI to work with software uploads.  It adds a new tab for software uploads.  These go into a separate software upload table that duplicates the structure of the current file uploads table.

I tried to get the two separate tabs to use the same code as much as I could so we don't have to maintain so many separate things which needed some trickiness. I was hoping not to have to go modify every query that uses the current file upload table and it keeps these two kinds of files very separate.  When saved to disk, one goes into a `#{resource_id}` directory and the other `#{resource_id}_sfw`.

I added software and file validation upload tests since we didn't have any before.


stash/stash_engine/app/models/stash_engine/software_upload.rb
```diff
@@ -0,0 +1,20 @@
+module StashEngine
+  # class SoftwareUpload < ActiveRecord::Base
```
ryscher, line 2: comments to cleanup?

stash/stash_engine/app/controllers/stash_engine/resources_controller.rb
```diff
@@ -101,13 +103,42 @@ def submission; end
...
+      @resource_assoc = :file_uploads
+    end
+
+    # Upload files view for resource
+    def up_code
+      @file_model = StashEngine::SoftwareUpload
+      @resource_assoc = :software_uploads
+
+      @file_model = StashEngine::FileUpload
+      @resource_assoc = :file_uploads
```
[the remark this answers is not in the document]
sfisher: Thanks for catching this, Ryan. This could've been a serious problem.  Not sure why it didn't break.

stash/stash_engine/app/models/stash_engine/software_upload.rb
sfisher: Got rid of this comment.

PR #117, opened 2020-04-23
super small commit for the sidebar so it suppresses the section for f…

sfisher:
…ile downloads for DataONE since only full version downloads work.  No reason to display an option that doesn't work.


ryscher:
Great, thanks for fixing this annoying issue!

PR #121, opened 2020-04-27
License selection and showing Zenodo files on review page

sfisher:
These changes show the zenodo files and the logos on the review page and also show license if software is present.


ryscher:
Looks good, and works on my test machine. Thanks for making it look so good. And I'm really glad you simplified the list of licenses; that will help to reduce confusion.

PR #125, opened 2020-04-29
Changing UI According to tickets 712, 713

sfisher:
https://github.com/CDL-Dryad/dryad-product-roadmap/projects/2#card-37125964

and

https://github.com/CDL-Dryad/dryad-product-roadmap/projects/2#card-37126234

These are a bunch of text rearrangements, but the biggest change is from ticket 713 since Daniella mentioned to show both options and hide the rest of the page (after step 1) in some circumstances, but in other circumstances it's inappropriate.  So needed to change labels, rework javascript and also fix a bunch of tests.

You can play with it in the UI and see things appear and disappear in different circumstances.  I think it does something reasonable, but it's more complicated that it seems.

### DIFF
--- a/spec/features/stash_engine/ui_file_upload_spec.rb
+++ b/spec/features/stash_engine/ui_file_upload_spec.rb
@@ -2,7 +2,7 @@ require 'rails_helper'
 require 'pry-remote'
 require 'fileutils'
 # binding.remote_pry
-RSpec.feature 'UiFileUpload', type: :feature do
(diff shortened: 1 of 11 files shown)


ryscher:
Thanks, Scott! This looks great.

I have one minor concern: the phrase "the URL for where your data live". To me, this implies that I can give the URL of a directory, and Dryad will grab all of the files from that directory. But perhaps other users won't see it that way.

PR #127, opened 2020-05-05
Zenodo code upload

sfisher:
This continues on from the previous pull requests relating to this topic and merges into the zenodo-code branch.  Probably the differences just in this PR will be clearer if you accept and merge the previous PRs first.

The main feature of this is to ensure all files that we need to upload to zenodo are present in our software directory (so download HTTP).  It checks presence and any digests if they are set and lots of tests of that.

There are also a couple of incidental changes structuring things better for moving forward.

1. The stash_engine_zenodo_copies table can handle tracking status of both data and software copies to zenodo.  This also changes so the resource uses  `has many` instead of `has one` for the zenodo_copies table.  That also necessitated a big bunch of minor code and testing changes (basically pluralize, filter to the upload type and take the first) since the relationship changed in ActiveRecord and the database relation.

2. There is a stub for zenodo_software_job (nothing much here yet).

3. I found the previous model pretty ugly since software upload class inherited from the data upload class and simply undefined a bunch of the methods from it.  Instead I put the common methods into a concern and then both classes pull those methods in.   This feels less icky to me and then the common methods are in the concern file and each class has methods specific to it in it's own file and it's clearer what is unique and what is common.


ryscher:
Looks great! Thanks for fixing all the test stuff.

PR #129, opened 2020-05-06
Changes to get tests working reliably on Travis and gem updates

sfisher:
1. For travis, it updates to Ubuntu 18.04 instead of 14.04 .  14.04 and the version of Chrome released on May 5th made tests stall out and randomly fail.  This seems to fix this.

2. Updating tests to use Puma instead of Webrick since neither are the server we're using for production, but Puma is more of a production-grade server.  Passenger doesn't seem to work correctly with capybara tests, though.  (I tried and it wasn't working correctly.)  Plus I'd argue we may want to move to Puma in the future since I haven't been thrilled with Passenger and it's not a common server to use these days.

3. Updating Ruby to 2.4.10 instead of 2.4.4.  Not sure of all the changes but probably some security updates and it's a minor update.

4. Updating the gems that would update (bundle update) for all our internal gems/engines and also the main app.  I'm a little disappointed that we can't get Nokogiri to update to 1.10 series since it has had some security updates, but maybe we can swing it when we go to rails 5 and even if I required this later version bundler seemed to ignore it since it seems like maybe it doesn't fulfill some dependencies of other gems we're using.  IDK why it will not update.


ryscher:
Yay! Thanks, this is incredibly useful.

PR #131, opened 2020-05-07
Allow geolocation in Zenodo according to Alex's info

sfisher:
This adds geolocation.  It seems we already had the methods in there.


ryscher:
Looks good. Thanks, Scott!

PR #133, opened 2020-05-07
Minor change for domain name for notifier.

sfisher:
This is a very small change to use a merritt-oai machine CNAME they can use on new machines without changing the domain every time the hardware changes (such as the linux 2 upgrade).


ryscher:
Looks good, and it's working properly on dev.

PR #135, opened 2020-05-15
Test of presigned versions 

sfisher:
(must fill in the user/pass for this Proof of Concept to work).

This is just a standalone Ruby script that can test/demo the presigned URL version generation and polling.

I also added anticipated way that we can avoid recreating new objects all the time if one is already assembled in a flowchart.


ryscher:
Looks great, works for me. Thanks for getting this working!

PR #137, opened 2020-05-18
Zenodo funding in notes

sfisher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/715 .  This adds the funding information into the output for notes.

Has tests and looks like these when submitted to Zenodo.  https://sandbox.zenodo.org/record/541797 and https://sandbox.zenodo.org/api/deposit/depositions/541797 .


ryscher:
Looks good. This seems like a good way to fit this into Zenodo, and the information would be parseable if we ever needed it again.

PR #138, opened 2020-05-18
Changes to notifier for new server name.

sfisher:
This is an extra pr.  The branch I'm merging into is a branch with the same contents as our last release tag.  We have to redeploy mid-release for this since Ashley wants to do her changes with Merritt tomorrow.

I'll retag off this branch so we don't pull anything from master until our next real release.


ryscher:
Looks great. Thanks!

PR #140, opened 2020-05-20
Zenodo software up to stage of uploading metadata

sfisher:
This does workflow up to updating metadata, though to really test the complexities of versioning states I'll need files working, too to test multiple versions and publishing.

- Refactoring File (ugly name collisions) to FileDownload
- Checking for more unusual states when starting a workflow.
- Slightly different workflows for versioning/publishing/re-editing existing
- Metadata update allows specifying DOI since metadata updates should use the same DOI in Metadata (or new doi for new version).
- Tracking DOI in the table.
- A mixin to share some common methods between zenodo software and 3rd copy resource workflow.

I'll add some more tests and cleanup/simply code more in the resource class once I see it works with file uploads to zenodo and it becomes easier for me to test many versioning cases against Zenodo to be sure stuff doesn't break.  I don't really want to start mocking up the Zenodo API service until I see that all my assumptions are right about how it treats things, since mocks that don't accurately represent an external system's behavior aren't helpful for testing.


ryscher:
This all seems good. I agree, it's much easier to deal with FileDownload than File.

PR #142, opened 2020-05-22
Fixes our failing dependency checks

sfisher:
The OAI server moved and the Notifier had blank lines at the end of the log, so the last line always being a date isn't great assumption.

if there is whitespace at the end, strip the string of all the blank space.  Also optimizes for only reading in the last 512 bytes instead of loading a possibly huge entire huge log file into memory and an array just check the last line.


ryscher:
Looks good, and fixes the problem on dev. Thanks for tracking this down!

PR #143, opened 2020-05-22
Zenodo 3rd production ready

sfisher:
This gets us ready to submit 3rd copies to zenodo.  In both dryad-app and dryad-config.

- Documentation updates
- Adds community and test to metadata
- Re-enables callback on publication to replicate/update the zenodo copy
- Removes is-identical-to relationship on Zenodo.

### DIFF
--- a/documentation/zenodo_integration/delayed_jobs.md
+++ b/documentation/zenodo_integration/delayed_jobs.md
@@ -1,5 +1,41 @@
-ActiveJob / Delayed Job
-=========================
+# Zenodo extra copies
+
+## It's not processing? Why? Look at this stuff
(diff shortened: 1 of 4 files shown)


ryscher:
Ok, this looks good now, and tests ok for me on dev. 

Note that there is currently a problem with DataCite on dev, and if you set the `skip_datacite_update`, updates to Zenodo are also disabled. This means we currently have to manually push things to Zenodo on dev.

sfisher:
Thanks Ryscher.  I guess I didn't encounter that since I probably mostly used new records that didn't have "skip datacite update" set.

I had also turned off the delayed_job daemon on our development server at some point in order to test and run it in my local environment.

PR #149, opened 2020-06-05
Add dataone_dash to the notifier to check on

sfisher:
Even though we really get no new datasets, we still need to check for updates on existing.


ryscher:
Great, thanks for tracking this down!

PR #150, opened 2020-06-05
This allows downloads from presigned versions

sfisher:
There is a ticket for this at https://github.com/CDL-Dryad/dryad-product-roadmap/issues/738 with a huge checklist of stuff that is happening.  It also has some URLs I was testing locally (or on dev environment for testing).  The main things to test are:

Download link works with:
- [x] Cached dataset
- [x] Uncached dataset and small size
- [x] Uncached dataset and large size
- [x] Works with both AJAX (in the UI) and also when doing a GET curl to the same URL /stash/downloads/download_resource/{resource_id}.

The expired cache key is hard to test, but it's a 410 response from Merritt and I have tests for it and it pretty much works the same way as a cache key that isn't found at all (404 on their side for status).  I guess we could manually ask someone at Merritt to expire something on their side if rather than waiting days for it.

You can clear out the cache by deleting all rows from the table stash_engine_download_tokens in the database.

Sorry this is such a large PR, but the tests ended up being pretty big and it turned out to be a lot more complicated than I thought to handle all the cases (AJAX, non-ajax, sharing links, etc).


ryscher:
The above use-cases work, but the API download seems to be returning the HTML of the landing page:
```
curl -L -O -J -X GET https://dryad-dev.cdlib.org/api/v2/datasets/doi%3A10.5061%2Fdryad.79q870b/download
```

ryscher:
Ahh, this problem only occurs for large files, which were already failing with the API call. Let's make a new ticket to fix this.

ryscher:
Wow, this new feature is really nice! Thanks, for putting it in, and especially thanks for adding all the detailed tests.

PR #151, opened 2020-06-08
Journal API -- Editing links

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/702

Adds a link for editing a dataset, of the form
```
https://datadryad.org/stash/edit/<DOI>
```

The URL format mirrors the sharingLink URLs rather than the API URLs, since it redirects to a submission in progress. The URL is included in API responses when the requesting user has permission to edit the dataset.

This PR also includes support for persisting a "targetPage" in the user's session object. This allows someone who is resolving the editLink to go through the ORCID login process and still be redirected to the correct page once they have completed login.


ryscher:
For future reference, I had to add this line to the tests:
```
visit stash_url_helpers.dashboard_path
```
This forces Capybara to visit a page, and create a session, before attempting to visit the editURL. Otherwise, it doesn't create the session on the initial request (though normal browsers don't have this problem).

And I was unable to get Capybara to persist a session through the mocked ORCID login process, although again, normal browsers don't have this problem.

PR #152, opened 2020-06-08
Change embargoEndDate to publicationDate in API

ryscher:
(this PR is rooted at the branch of #151, since that isn't merged yet)

We were asked to add the publicationDate to API responses. Apparently, it was already in the API, but it was labeled as `embargoEndDate`. Along with the name change, I'm moving into the main metadata.rb, so it can appear for requests of both `api/v2/datasets` and `api/v2/versions`.


ryscher:
Not sure why git is confused about this... the last commit should be the only difference.

PR #153, opened 2020-06-08
Just a small sql fix

sfisher:
Using the ? syntax instead


ryscher:
Thanks for finding this issue!

PR #154, opened 2020-06-09
Require manuscript number (or DOI) in submission

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/719

If the user enters a journal name, require either the manuscript number or the article DOI. To make this work, I added some machinery to save the DOI in the `issupplementto` relation. Previously, the DOI was only saved if the user hit the "Import Metadata" button.

Also corrected the validation text for emails to match the logic of what is actually being tested. The curators have complained about this; we should discuss whether the logic should be changed.


sfisher:
Looks like rubocop tests are failing on this.  :-)

stash/stash_datacite/app/controllers/stash_datacite/publications_controller.rb
```diff
@@ -59,6 +58,18 @@ def save_form_to_internal_data
...
+      save_doi
+    end
+
+    def save_doi
+      form_doi = params[:internal_datum][:doi]
+      related_doi = @se_id.publication_article_doi
+      return unless form_doi.present? && related_doi.blank?
+      bare_form_doi = Stash::Import::Crossref.bare_doi(doi_string: form_doi)
+      res = @se_id.latest_resource
+      res.related_identifiers.find_or_initialize_by(related_identifier: bare_form_doi,
```
sfisher, line 70: Hi Ryan,

I'm not sure if this is a problem or not, but I've had problems with find_or_initialize_by doing something I didn't expect in the past. From quick Googling, it seems like "find_or_initialize_by" doesn't save the record but just creates an unsaved record with the data you have.  I didn't see it being saved to the database afterward, but maybe it is.  I think find_or_create_by will create and save it to the database.

The other thing I got bit by (and can't totally remember details) is either the way it saved the data or did the find for the record, so it's just a vague feeling of unease about the method.

Anyway, I don't know for sure if anything is really wrong here, but could you just look over this code to be sure it is really doing what you expect?

sfisher:
I think this looks ok to me so please go ahead and merge into master if this is ready.

PR #157, opened 2020-06-11
Fix path mangling

sfisher:
This prevents http.rb from normalizing the paths in URLs by setting the normalizer.  The one in here is pretty much the same as the one built into the http.rb code but instead of changing the path it just uses the one that the user put in. `uri.path` instead of `uri.normalized_path` or something like that.

*Yes, I really wanted that path in the URL and I didn't want you to screw with it, http.rb*

I had to dig through code to find the option since I didn't really see it documented anywhere.


ryscher:
Looks good. I created a file with a lot of crazy diacritics in the name, and it works perfectly with this update. Thanks for tracking it down! Character encoding issues are never fun.

PR #162, opened 2020-06-15
Update schema.org metadata

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/720

Improve the schema.org metadata to follow current guidelines from Google. 
- Change the download URL to use the API instead of Merritt
- Change the `citation` and `identifier` to use the recommended values
- Add ORCID
- Add ROR
- Add `name`, because Google prefers it instead of the more specific `familyName` and `givenName`
- Change `fileFormat` to `encodingFormat`


ryscher:
Ugh. Tests are failing here because the stash_datacite rspec can't access stash_api.

sfisher:
Looks like something in the stash_datacite tests in the engine is failing.  If you run it locally you can examine the output file and get a more useful error rather than than the "colorize stack level too deep" one.

ryscher:
@sfisher The tests are now passing. I've moved the relevant test into the main spec directory, along with the data files it depends on, and some data files that were extraneous. I'll take a look at moving the other tests in stash_datacite.

PR #165, opened 2020-06-23
To fix the text/html errors

sfisher:
Merritt is returning text/html instead of json sometimes through the API requests.  IDK what type of errors.  This converts it to a standard 404 status for API use and also raises exceptions for 500 internal server errors from Merritt since we probably would like to know if Merritt is flakey.

Also added a couple of testing fixes from some tests that got moved over.  Compare times in UTC and also only generate factories for the datacite metadata we support in our database through the enum (because otherwise it causes test errors).


ryscher:
Thanks, Scott. This looks good. Hope it fixes the problem!

PR #166, opened 2020-06-23
No longer disallow DataONE individual file downloads.

sfisher:
This removes the file download block for DataONE.


ryscher:
Looks good. Thanks, Scott!

PR #170, opened 2020-06-26
Merging changes for workflow, tests and complete software replication process

sfisher:
Sorry, this is a large PR, but it's all functionally related to getting submissions for the workflow and the subtasks of it all working for zenodo software.  I've added a lot of tests, but essentially it should check state and do all the slightly different things that zenodo needs for different cases.

- Metadata isn't submitted to zenodo unless files have been uploaded for zenodo before
- For uploaded before
  - Metadata isn't submitted to published (closed) zenodo datasets on a normal submission
  - Metadata is updated to zenodo if the dataset is being edited and isn't published
- Metadata is always updated on a Dryad publication event if files are in zenodo. And then it's also published in Zenodo.
- Software changes (additions/deletions) always update zenodo and could be:
  - New dataset (not submitted before)
  - Open dataset (just change the files)
  - Closed (published) dataset, in which case it creates a new linked DOI at zenodo for editing.
- Tracking the zenodo DOIs for software
- Some metadata differences for software (license, related identifiers, doi minting from zenodo instead of bringing our own)
- ActiveJob class for working with ZenodoSoftware
- Table for tracking zenodo status has both software and dataset submissions
- Refactoring of relationship between resource to Zenodo table to 1 to many.
- Refactor of "Resource" class in the Zenodo areas to be called "Copier" to avoid annoying overloading of terminology and make things clearer.
- Lots of tests around workflow since versioning is pretty mismatched with our model, so there were plenty of opportunities for things to go wrong if we didn't have the rules correct.
- I believe I added more prerequisite validation for zenodo software copying.


ryscher:
This looks good. Thanks for all the extra testing, I know it's a pain but it will definitely help with this complex feature.

PR #172, opened 2020-06-26
Logos for claremont colleges

sfisher:
See 
[logo_screenshots.zip](https://github.com/CDL-Dryad/dryad-app/files/4839571/logo_screenshots.zip) for the results.


ryscher:
Looks good, and all logos display properly.

PR #175, opened 2020-06-30
Security updates

sfisher:
I went down a rabbit hole with trying to get the updates to work on my system and eventually upgraded to Ruby 2.6.6 and also updated Rubocop to a more recent version and fixed a few small things that didn't work right in Ruby 2.4 to 2.6.

We can discuss if this is something we really want to merge in right now, but I believe it is all working.  All the tests pass for both sets of tests (the full application and the tests in all the engines and gems).

This would be a little annoying to deploy (we'll need to upgrade Ruby on the server manually when we take it out of rotation for the deploy).  And it may be something we want to test a bit.

But on the other hand since it is working, I don't see a big advantage in delaying the Ruby upgrade until later since we'll need to do it sometime since Ruby 2.4 isn't getting security updates anymore.  I believe the end-of-life for Ruby 2.6 will be in about 2 years (2022) based on the pattern of previous releases (2.4 was end of life a few months ago).

Most of the big changes were made from Rubocop autocorrections based on a newer version and Ruby 2.6.

We can discuss in our standup tomorrow.


stash/stash_engine/app/helpers/stash_engine/application_helper.rb
```diff
@@ -17,17 +17,19 @@ def log_in_out
     def filesize(bytes, decimal_points = 2)
       return '' if bytes.nil?
       return "#{bytes} B" if bytes < 1000
+
       size_str = ::Filesize.new(bytes, Filesize::SI).pretty
       return size_str.gsub('.00', '') if decimal_points == 2
 
-      matches = size_str.match(/^([0-9\.]+) (\D+)/)
+      matches = size_str.match(/^([0-9.]+) (\D+)/)
```
ryscher, line 24: I don't think it's correct to remove the escaping on the period here -- it looks like it's changing the meaning.

stash/stash_datacite/app/controllers/stash_datacite/affiliations_controller.rb
```diff
@@ -12,7 +12,7 @@ def autocomplete
         render json: nil
       else
         # clean the partial_term of unwanted characters so it doesn't cause errors when calling the ROR API
-        partial_term.gsub!(%r{[\/\-\\\(\)~!@%&"\[\]\^\:]}, ' ')
+        partial_term.gsub!(%r{[/\-\\()~!@%&"\[\]\^:]}, ' ')
```
ryscher, line 15: I don't think it is correct to remove escaping on the parens here -- the intention is that parens are explicitly listed as characters that are filtered out, they should not have their normal meaning in a regex.

stash/stash_datacite/app/controllers/stash_datacite/contributors_controller.rb
```diff
@@ -53,17 +53,19 @@ def delete
...
     def autocomplete
       return if params.blank?
+
       partial_term = params['term']
       if partial_term.blank?
         render json: nil
       else
         # clean the partial_term of unwanted characters so it doesn't cause errors when calling the CrossRef API
-        partial_term.gsub!(%r{[\/\-\\\(\)~!@%&"\[\]\^\:]}, ' ')
+        partial_term.gsub!(%r{[/\-\\()~!@%&"\[\]\^:]}, ' ')
```
ryscher, line 62: I don't think it is correct to remove escaping on the parens here -- the intention is that parens are explicitly listed as characters that are filtered out, they should not have their normal meaning in a regex.

stash/stash_datacite/app/controllers/stash_datacite/publications_controller.rb
```diff
@@ -36,7 +36,7 @@ def autocomplete
         render json: nil
       else
         # clean the partial_term of unwanted characters so it doesn't cause errors
-        partial_term.gsub!(%r{[\/\-\\\(\)~!@%&\"\[\]\^\:]}, ' ')
+        partial_term.gsub!(%r{[/\-\\()~!@%&"\[\]\^:]}, ' ')
```
ryscher, line 39: I don't think it is correct to remove escaping on the parens here -- the intention is that parens are explicitly listed as characters that are filtered out, they should not have their normal meaning in a regex.

ryscher:
I was able to get this running on my Mac; I added some steps to `documentation/dryad_install.md`.

@sfisher --- after some testing, it appears that the changes that rubocop made to the regular expressions are *correct*. That is, character escaping doesn't work the normal way when it is within a [ ] character class. I don't believe this is typical of other languages, but I guess if it's the Ruby way, it's ok. What do you think?

sfisher:
Yeah, I think there are two different notations in Ruby, the `/regex/` notation and the `%r{regex}` notation.  I think the `%r` notation specifically allows forward slashes unescaped so it makes putting in things like URLs much easier.  From stackoverflow it seems like it just uses whichever character you use after the %r for the delimiter instead so that it doesn't need to be escaped if you have a lot of that character in your regex.  https://stackoverflow.com/questions/12384704/the-ruby-r-expression

I don't remember all the intricacies of regular expressions and it may be that certain characters (like the parenthesis for capture groups) don't need to be escaped if they are in a `[]` set of characters, but they do in other places.  Seems like it might be the same with the dot character.  I think that is what is going on with some of these simplifications maybe.

Anyway, I'm fine if you want to just disable that cop or it seems you determined it doesn't break the regular expressions it is changing and I'm also fine with that.  I really don't feel strongly about this piece of style guidance (as I don't with a lot of the really small changes rubocop makes).  I often end up writing my code in the easiest way and not too messy and let rubocop fix things to the exact formatting it loves so much.  🤷   I don't think I've ever seen it break things or maybe only a couple times over years.

I guess the consistency in formatting is nice.  I've also seen it attack libraries written by the Rails core team by mistake and tell me things were wrong.  🤔   I assume they know how to do things and I find it too hyperactive, honestly.

Feel free to loosen up the settings a lot.  Another thing we probably need to do is simplify the config for it some because there are a couple of rubocop configurations and it would be awesome to unify into one.

ryscher:
This PR is now working on both my Mac and my Linux VM, and I've run through some tests in the running system. So we're all good to go. 

Thanks for getting it set up, Scott!

PR #176, opened 2020-07-02
Small update to cap version and instructions for Ruby version upgrade

sfisher:
Has changes to version of capistrano and also instructions on how I got Ruby update to work on the servers.


ryscher:
Wow, thanks for the detailed instructions!

PR #178, opened 2020-07-06
Bundler problems fixed for deployment on stage

sfisher:
I updated to bundler 2 in order to get this to work on stage.  For some reason, bundler 1 couldn't handle it and caused me lots of errors.  I really don't know why, but this fixes it.  I also re-updated gems to be sure they were compatible.


ryscher:
I'm getting this warning when I run `bundle install`:
```
rails (4.2.11) has dependency bundler (>= 1.3.0, < 2.0), which is unsatisfied by the current bundler version 2.1.4, so the dependency is being ignored
```

Everything seems to be running ok, so I think this is ok for now, especially since we're moving to Rails 5 shortly.

PR #180, opened 2020-07-07
Bundler problems

sfisher:
I hope this merges correctly and works since I had to go back to an earlier commit and work out a bunch of problems.

The things I found out.

- Bundler 2 randomly works sometimes and randomly fails others.  Some of the messages indicated it's not compatible with Rails 4 when it doesn't work (but in some cases worked).  So going back to bundler 1.17.3.
- Bundler 1.17.3 fails with "default gems" that seems to come with the ruby core classes.  It's better to switch these to older versions if it's trying to update them.  But you can manually switch a default gem if you need to like `gem install --default -v1.0.1  etc`  (etc was a gem causing me lots of issues.)
- Most of our server deploys are done in `deployment` mode which made bundler fail in different ways than normal and so testing manually in that mode and with bundle exec was the best way to predict if something would work on the server.
- We seem to have readline libraries on some servers and not others and so different behavior on some.  Using the rb-readline gem worked OK, but only with the special "require: false" flag in our Gemfile since it otherwise brought up problems.

If the master branch still keeps the bundler 2 things, I may need to make another pull request actually manually patching those changes out.  But this branch did actually work correctly on the development and both stage servers finally.


ryscher:
Thanks for working through these issues, Scott. Sorry bundler is causing so many problems. This PR is ready to merge whenever Travis is happy.

PR #184, opened 2020-07-08
This replaces the cirneco methods with our own

sfisher:
cirneco was using Maremma gem (one of their own making) to do HTTP requests and it seemed no longer able to do http basic auth with Ruby 2.6.6.

I took the methods out of their code, inlined them into our class and changed the http library to http.rb and rewrote into slightly different syntax to do the requests.


ryscher:
This looks good and tests fine for me on the staging system with DataCite sandbox. Thanks for putting this together so quickly!

PR #187, opened 2020-07-10
Allow more flexible base URLs for Merritt Repo and handle timeouts

sfisher:
This changes the config repo so that the domain is really port+domain+(possibly protocol).

It also handles timeouts from Merritt (they return a 408 to us as part of their status hash).

I'll have Terry test this out with his docker containers, but tried to mock something up.

Displays a message that our download isn't working right now and lets them know to try files or to try again later on timeouts.

Also corresponding PR at https://github.com/cdlib/dryad-config


ryscher:
Looks good, and basic submissions to Merritt work fine on my machine. Thanks for the extra testing!

PR #188, opened 2020-07-13
Testing for the DataCite submission requests

sfisher:
This PR adds tests for the three methods I created to replace the Cirneco methods we used to use from their library and broke.

### DIFF
--- a/spec/lib/stash/doi/datacite_gen_spec.rb
+++ b/spec/lib/stash/doi/datacite_gen_spec.rb
@@ -22,23 +22,26 @@ module Stash
 
         describe :post_metadata do
           it 'creates a successful sandbox post request to DataCite' do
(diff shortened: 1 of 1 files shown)


ryscher:
The tests themselves look fine, and pass. However, I don't think we should avoid including the auth string in the stubs. Can we just omit it and match on any string? If not, we should be able to call a method to generate it from the username and password using `ActionController::HttpAuthentication::Basic.encode_credentials`

sfisher:
Ok, will come back to that when I'm not in the middle of working on a different branch, maybe later today.  I think it's useful to test that the username and password get passed through the method and make it into the request.

sfisher:
Ok.  I updated this code to use that method for basic auth.

PR #190, opened 2020-07-17
Rework of most things about sending to DataCite

sfisher:
The readme.md files below explain most of it, I think.

This now gets all the reports from DataCite and then uploads the bad ones based on the files in a directory.  It doesn't rely on any of our state in the python tool, so should be able to be run independently.

I also added a bunch of tests.


ryscher:
Looks good. Hope this improves the processing...

PR #192, opened 2020-07-20
Fix for someone deleting and then adding and then deleting files

sfisher:
This is a fix for https://github.com/CDL-Dryad/dryad-product-roadmap/issues/787 .

It doesn't rely on any of the current version's state but looks back at the previous version (if it exists) and figures out if a deletion is just a deletion from the file system and from the database or if a deletion means sending a removal request to Merritt instead.


ryscher:
Looks good, and tests fine on my machine. Thanks for catching this crazy edge case!

PR #194, opened 2020-07-22
Making the notifier more robust and lots of tests

sfisher:
This is for ticket https://github.com/CDL-Dryad/dryad-product-roadmap/issues/793 .

This took way longer than I thought.  The tests for the main app and command line stuff turned out to be tricky and I also got some weird state going on and spent a lot of time debugging and figuring out how to reset class loading and other stuff.

PS. Travis kept failing on the first test because of some weird loading/dependency issue, so I eventually gave up and added a fake first test to get it out of the way after wasting literally 4+ hours.  I don't like this solution but don't know what else to do since I cannot isolate the problem because it doesn't happen either on My Mac or on a fresh install I had of Linux Mint (based on Ubuntu).  Their service is a black box with no great way to troubleshoot aside from making pushes and waiting 5-10 minutes for things to fail.  (They say they may allow some kind of access if you email them and beg for access, but I'm not inclined to do it over one small test.)

I've also never seen this problem when running the script in the real environment.  This testing turned out to be tricky.


ryscher:
Looks good. Thanks for working through all the testing issues.

PR #196, opened 2020-07-24
Documentation about how to update notifier code

sfisher:
These are instructions on how to swap out the code for a new version of the code when we update the notifier.


ryscher:
Thanks, this is a useful update.

PR #197, opened 2020-07-24
Changes to allow the counter token to be stored with our config

sfisher:
It is pulled out in the monthly cron script in the config repo.  See also that PR.


ryscher:
Ok, looks good.

PR #198, opened 2020-07-29
Changes to cache and pull usage counts from DataCite stats

sfisher:
This is more or less similar to what was in place before (a while back), but with the following changes:

- Switching out to HTTP.rb client since I'm using it many other places at this point
- Looking at stats, it seems like the statistics from another tree area in the JSON is more accurate and reliable.
- Handles getting all pages of results (if there are more than 500 per page).
- Caching is for 7 days before updating numbers again.
- If numbers of views/downloads mysteriously fall in number (maybe DataCite hasn't ingested it all yet) then it doesn't downgrade the numbers.
- "Generic query" builds URL from either a full URL or from base URL and separate params hash.
- Made raised errors all use one error to be easier to deal with.

I tested paging manually by setting page size to 5 (Instead of 500) to be sure it aggregates pages.

On development or stage it should work fine against some of the test identifiers we imported while doing migrations.

I took a look at the actual citations again. We set those up to check them weekly (but only re-retrieve if they haven't been updated in a month).  Those still just come out of our database but are updated from the remote API periodically.

### DIFF
--- a/spec/features/stash_engine/landing_spec.rb
+++ b/spec/features/stash_engine/landing_spec.rb
@@ -13,6 +13,7 @@ RSpec.feature 'Landing', type: :feature, js: true do
   include Mocks::RSolr
   include Mocks::Ror
   include Mocks::Stripe
+  include Mocks::Counter
 
   before(:each) do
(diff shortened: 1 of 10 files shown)


ryscher:
Looks good, updates stats reasonably. Thanks for hooking this up again!

PR #201, opened 2020-08-06
FOS subjects in our subjects and sent to DataCite

sfisher:
These are the changes for https://github.com/CDL-Dryad/dryad-product-roadmap/issues/776 .

This basically creates a migration to be sure the FOS vocabulary is added to our database with the FOS terms and the subject scheme 'fos'.

It changes the places where we don't want to expose that with the ActiveRecord scope "non_fos."

Current tests pass and added tests to check that FOS subjects are excluded by the scope and that when sending these special subjects they are prefixed as Martin says in that ticket/chat.


ryscher:
Looks good. Thanks!

PR #203, opened 2020-08-07
Adds the Fields of Science subjects UI

sfisher:
According to requirements at https://github.com/CDL-Dryad/dryad-product-roadmap/issues/811 .

Also added UI tests to be sure they appear as entered.


ryscher:
Looks good, and works fine on my VM. Thanks for adding this feature -- we have needed it for years!

PR #205, opened 2020-08-11
Tweaks to ordering of fields and remove required display asterisk

sfisher:
A very small cosmetic PR to change order and take away the required asterisk for fields of science.


ryscher:
Looks good, works fine. Thanks!

PR #206, opened 2020-08-13
Fixing up bad DOIs for related identifiers

sfisher:
This does updates for common formats by regular expression with SQL updates and then goes through the remaining items to see if it can find a DOI-looking string inside the thing they typed in.

The regular expression I'm using is one that CrossRef published on their blog.

This just write the results to fixed_id field and then when we feel comfortable with the results we can update from that field to the real related_identifier field for those items.

Tests for all the individual patterns and teh rake task just aggregates and calls all of them.


ryscher:
Looks good, seems to run well on dev. Thanks for cleaning these up!

PR #210, opened 2020-08-24
Mangled the tenants and didn't even seem needed for the test

sfisher:
I removed this and it fixed test failures in my other tests and also didn't seem needed to begin with since the tests for the dataset_parser_spec still work fine.

I re-setup rspec bisect and had to change it to shell mode and let it run tests for a while to find which class was destroying things.


ryscher:
Looks good. Thanks for finding this!

PR #211, opened 2020-08-25
Rails 5.1

ryscher:
Upgrade to Rails 5.1, for https://github.com/CDL-Dryad/dryad-product-roadmap/issues/769

Updates include:
- In many tests, replaced `Object.create` methods with factory creation. This is due to the fact that Rails 5.1 performs more strict validation of objects before saving them to the DB. In many cases, the old tests were only creating partial objects, which would not save to the DB correctly.
- Annoying new change: `uniq` no longer works for ActiveRecord. So now you must use `distinct` for ActiveRecord relations, and `uniq` for arrays/hashes. This makes sense from the DB implementation perspective, but in practice it's annoying to track when something is still a relation and when it has turned into an array.
- Started to remove the `ResourceBuilder`, in favor of factories. This means that some tests are less specific, since they don't have all of the detail of the XML files that are associated with the `ResourceBuilder`, but these tests are now also less fragile, and can focus on testing exactly what we need.
- Removed a lot of the style that used `attr_reader` to access instance variables within a single spec class. This style didn't work with 5.1, and tended to produce less readable code.


ryscher:
Ugh, this is failing due to one of the old DB migrations. Will need a small update...

PR #212, opened 2020-08-26
Double encode hash sign if present in a filename

sfisher:
This is related to https://github.com/CDLUC3/mrt-doc/issues/426 .

It seems like the Merritt folks will consider fixing in the future as they begin to release further APIs and can take a more consistent approach. This is a #wontfix from them for now.  Our workaround should allow our users that have number signs in their filenames to still download.

You can test on dev (or wherever by uploading a file with a number sign in it and see that it downloads).  Example: https://dryad-dev.cdlib.org/stash/dataset/doi:10.5072/FK25H7MF7Q


ryscher:
Looks good, works fine.

PR #213, opened 2020-08-26
Update documentation about deploys and maintenance

ryscher:
Much of this was moved from the Confluence page so it can all live in one place. Also integrated some of my notes about the process of doing a full deploy, and the changes that Ashley made on the servers.


documentation/server_maintenance/deploying_new_code.md
```diff
@@ -0,0 +1,179 @@
...
+### deregister
+load-balancer.bash uc3-dryad-prd-alb deregister uc3-dryaduix2-prd-2a
+load-balancer.bash uc3-dryad-prd-alb deregister uc3-dryaduix2-prd-2c
+
+### reregister
+load-balancer.bash uc3-dryad-prd-alb register uc3-dryaduix2-prd-2a
+load-balancer.bash uc3-dryad-prd-alb register uc3-dryaduix2-prd-2c
+```
+
+Checking for Chunked Upload in the UI
```
sfisher, line 156: I have tested uploads while taking servers out of rotation with the ALB and it seems to seamlessly keep uploading chunks to the other server since they both use shared state (EFS is shared for files and DB state is also shared).  I wasn't clear this was the case, but I don't think anything needs to be done about chunked uploads so long as only one server is removed from the ALB at a time.

sfisher:
IDK why these tests are suddenly failing. Weird.  It looks like
`DELETE http://127.0.0.1:9515/session/280be91dc7f6329c44ce5d194594cad8` is being caught as a network request that is disabled.

PR #215, opened 2020-08-27
Fixing Travis complaint about un-mocked delete statements

ryscher:
Hopefully, removing these extra sign_out statements will remove Travis's issue.


sfisher:
Weird that is still shows this and also that it seems to pass in the Pull Request, but not the branch.  Maybe after it is merged it works right or maybe Travis is just flakey.

Go ahead and merge this in if you want.  This is a weird problem.

ryscher:
Whew! It finally works.

sfisher:
Thanks you, Ryan.  So odd that this change just cropped up but this seems like a good solution.  It's so odd that it just triggered from one action we didn't know where it was coming from.

PR #216, opened 2020-08-31
Here is a sql query to run a report to get info about authors at an institution

sfisher:
You just need to substitute ROR ids in two different places.


tracykteal:
Thanks for adding this to the repo. It will be very useful to have handy.

ryscher:
Thanks, Scott. This will be very useful for quick reports.

PR #217, opened 2020-09-01
Rename software to supporting information

sfisher:
This is just renaming "software" to "supporting information" and it seems like Rubocop decided to change something.  I guess we're not going to merge any of this into main for a while.


ryscher:
Looks good. Hoping we can get all of this merged back in soon!

PR #218, opened 2020-09-02
Adds a query that gets list of UC authors

sfisher:
This uses DISTINCT so variations on names and institutions means it has duplicates and our data is messy.


ryscher:
Looks good. Yes, the near-duplicates are a problem. If only we could force ORCIDs for all authors. But even then, I've seen many people who don't understand and create multiple ORCIDs.

PR #219, opened 2020-09-02
Rails 5.2

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/770

This is a relatively simple update. 

The primary functional change is the addition of bootsnap, which makes the server start faster, and seems to give a little speed boost to each page load.

The other major change in 5.2 is the addition of Active Storage, which is not enabled, but could allow us to simplify some asset handling in the future.


sfisher (approved):
Thanks you for going through all this, Ryan.  It seems good to me as I read over the changes and also tried running the server and testing basic functions on my local machine.

I think we should merge it and then test and correct any problems we find from this point.  I put in a ticket for one keywords entry problem in 5.1 (not sure if it's still a problem in 5.2), but we can work on any minor issues we find separately and we will have a week to find and correct anything (and I don't believe there are any huge problems that can't be fixed before our next release).

PR #220, opened 2020-09-03
Update API docs on API accounts without ORCIDs

ryscher:
Institutions often use role-based accounts for accessing APIs. Updating our API documentation to indicate how this will work within our existing system.


sfisher (approved):
Looks good.  Thanks, Ryan.

PR #221, opened 2020-09-04
Add DOIs to submission_queue page

ryscher:
Addition to the table for convenience when troubleshooting.


sfisher (approved):
It will be convenient to have this.  At some point in the future, I'd love to have a details page with even more info about an item.  This will make life a little easier.

PR #222, opened 2020-09-10
Disable extra rails source checking

ryscher:
Rails 5.1 introduced stricter checking of HTTP Request objects. The stricter checking is not compatible with our current Apache config, so this disables it.


sfisher (approved):
Thanks for working with this, Ryan.

PR #224, opened 2020-09-10
Add note about system user

ryscher:
Some of our automated tasks assign objects to user 0. In the past, we didn't actually have a user associated with this ID, but as of Rails 5.1, this user is required to be present for the associated objects to validate.

I'm adding a note to the install instructions about creating this user, though I'm not sure this is the best way to handle the requirement. Perhaps we should put this in a rake task that is run when a new database is set up?


sfisher (approved):
Thanks for the documentation change.

PR #225, opened 2020-09-14
Upgrade Bundler, Rubocop, and other gems

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/771

Upgrade Bundler to 2.1.4. This should fix many of our issues with Bundler compatibility. 

Also upgrade Rubocop to the latest version, and make most of the minor code updates it suggests. Migrate all of the rubocop configs into the top-level `.rubocop.yml`


sfisher (approved):
Thanks you, Ryan.  This looks good to me. The only one I wasn't sure of was the removing of an initialize method and super, but perhaps it automatically uses the parent's initialize method?

Anyway.  Thanks for dealing with all these little changes.

ryscher:
Ruby will default to using the `initialize` from the superclass unless you override it. So Rubocop flagged that initialize as being redundant, because it did nothing other than explicitly call the super.

PR #226, opened 2020-09-15
Reinstate tests that had intermittent failures

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/806

Reinstate the tests that were disabled were failing intermittently when we moved to Rails 5.


sfisher (approved):
Thanks for making the changes and getting things working.

PR #227, opened 2020-09-16
fix deprecated render nothing and use body: nil instead.

sfisher:
This is apparently obsolete now to use "render nothing" and no longer works.

It seemed to work on dev, at least most of the time.  I saw some errors earlier on, but they resolved around the time of the changes to loosen the security for the problems we were having.

Even when it was causing us to get error emails I think it still worked most or all of the time.

This should fix the problem, though. https://stackoverflow.com/questions/34688726/the-nothing-option-is-deprecated-and-will-be-removed-in-rails-5-1  🤷 

I just branched off main and deployed that branch so we don't get email message all night and day.

I didn't see the stage ones since I have them going into their own folder I didn't look in.


ryscher:

PR #228, opened 2020-09-16
This was a manual SQL query that was a "one off" for some reporting

sfisher:
but we keep getting requests for it or slight variations, so it's easier to put it in a rake tasks and automatically fill in the RORs from the tenant config.  It saves me copying and pasting and reformatting the ror_ids in the YML file into the SQL IN clause for each report we want to run.

It should make it easier to run these reports quickly in the future and documents the SQL query in code.

If we think tests are needed, I could probably add something but nearly all the logic is in SQL and very little in code aside from extracting the RORs into SQL format and outputting as CSV.  I suppose I could separate those into separate methods that are more easily testable if we get more requests for similar kinds of reports.

### DIFF
--- a/lib/tasks/reports.rake
+++ b/lib/tasks/reports.rake
@@ -1,0 +1,15 @@
+require_relative 'reports/ror_author_datasets'
+namespace :reports do
+
+  # use like: bundle exec rake reports:ror_author_submitted tenant=ucop RAILS_ENV=production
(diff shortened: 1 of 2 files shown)


ryscher:
Thanks, this will be useful. One minor concern... when I run the rake task, it's not immediately obvious that the result will go to a file. It would be nice to output the filename.

PR #230, opened 2020-09-22
Intermittent submission job error

ryscher:
Attempt to address another intermittent testing failure.


ryscher:
Ugh, this still fails with seed 51506.

sfisher:
Hi Ryan,

I ran bisect with that seed and it claimed this is the minimum reproduction of the problem.

```
bundle exec rspec ./spec/features/stash_engine/dataset_queuing_spec.rb[1:1:2] ./spec/models/stash-merritt/submission_job_spec.rb[1:1:3:2] --seed 51506 -fd
```

The test that runs before the failure is `should pause transfers`.  It may be that we just want to disable some of the queue pause, etc tests.  They were kind of annoying and hard to test and there must be something interacting badly between those two tests or state doesn't clear completely after that first test.

sfisher:
I think I might have a fix for this.

sfisher:
I put a patch in PR #231 since I got tired of my tests in there failing again and again.  I had to change some weird class patching to a rspec mock that is automatically removed after each test.

ryscher:
Closing, since it looks like the fix in #231 is more effective.

PR #231, opened 2020-09-22
Hide display for disliked items

sfisher:
As part of cleanup Daniella wanted to hide many icky related works, but not delete the data completely.  This allows items that are disliked to have `hidden: true` set and they will no longer display on the landing page for the dataset.

Right now this hidden display will be set manually based on our related identifier cleanup.

Added some "requests" tests and just requesting that javascript for that part of the page.

Not sending the bad items (validation or no display) to zenodo since they won't take badly formatted things and also don't want a backdoor to displaying them through their UI when it's replicated.


ryscher:
Looks good. Thanks for tracking down the problem with the tests!

PR #232, opened 2020-09-24
Rails5 secrets

ryscher:
This moves all config files from the `dryad-config` repository into this
repository. Passwords and other secret information are now stored in
`config/credentials.yml.enc`, encrypted with a key that will be stored
on each server.

To use the new credentials, the `master.key` file must be placed in
the config directory. (Copy this file from the Dryad dev server.)

Commands for working with the encrypted credentials:
```
rails credentials:show
rails credentials:edit
```

Within the encrypted file, I have collapsed much of the hierarchy in
an effort to keep things simple. We can re-introduce the hierarchy
if needed, but it would increase complexity of the commands for
referencing the values.

The configuration files are structured so that the `test` environment
does not need any of the secret information, and this environment will
not attempt to decrypt the credentials file. Configuration
files for the `test` environment are loaded from
`dryad-config-example` (this has not changed from previous versions of
the code). 

Since the dryad-config repository is no longer needed, it is best to remove
any symlinks that you have in your config directory before updating to this pull request.
The Capistrano config has been changed so it no longer symlinks
content from the dryad-conf repo, but it does symlink any *.key files
that are located in the Capistrano `shared` directory.


sfisher (approved):
I read over all this and it looks good to me.

I tested on my local machine and it worked fine.  I had a weird login issue with ORCID, but then it worked.  IDK why the login issue.  Maybe it was something on their end.

Thanks, Ryan.

PR #234, opened 2020-09-29
Add dataset details to API

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/847

Enhance results in the API to include:
- license (at the dataset level only, not version)
- lastModificationDate
- visibility

Also add a test for `relatedWorks`, which were already in the API, and rename `storage_size` to `storageSize` for consistency.


sfisher (approved):
Thanks for adding these things, Ryan.  Looks good to me.

PR #235, opened 2020-09-29
Configuration fixes to fix bugs

sfisher:
- Fixes problem with secret_key_base not working correctly from environment sections of config file.
- Fixes problems with EZID not correctly minting identifiers because it uses the configuration named "account" but it was set to use "username".
  - I thought about changing the code to use username instead of account, but I didn't want to introduce a lot of code changes so just changed the config to use the correct key.
- This also has a bunch of `binstub` updates made by bundler when I regenerated them, which afaict, are basically wrappers around using binaries/commands contained in gems that are part of your app. Online resources said binstubs would be commited to repo.


ryscher:
Thanks for figuring out these issues! Sorry about the EZID configs... looks like that was a copy/paste error on my part.

PR #236, opened 2020-09-29
Fix autofill by doi

sfisher:
This fixes the autofill, works with the standard DOI url format used in the UI for storage and uses the 'cites' relation which is what has been defined for papers.  Tests fixed up.

Why binstubs have any changes, IDK, but created by bundler and shouldn't have effects on the application itself.


ryscher:
Looks good, and works for me. Thanks for fixing. 

Note for @dlowenberg and @LoriEP -- this fix will go out tonight, so the auto-population of the article relationship will still work.

PR #229, opened 2020-09-21
Simplification and display for related works

sfisher:
Things in this PR

- Removing PDF generation, special layouts, links, WickedPDF
- Rename citation to related_works for validation
- Change the form to simplifed format that
  - Accepts DOIs (url form) and URLs automatically
  - Transforms strings that appear to contain DOIs into DOI https format
  - Detects and transforms accession numbers like 'GenBank:AF3875738` into URLs
  - Displays errors near entry for formatting or not able to validate
- Displays related works on review page
- Displays related works in sidebar on landing page
  - Categorized by work type
- work type and verified added to database
- General testing for entry of a related work in UI
- Individual tests for functions, especially those added to related_identifiers

PR #238, opened 2020-09-29
More instances of converting `issupplementto` to `cites`

ryscher:
Supplement to https://github.com/CDL-Dryad/dryad-app/pull/229 for https://github.com/CDL-Dryad/dryad-product-roadmap/issues/837

These changes affect various places where Dryad uses article DOIs, including:
- in the submission system: 
  - autofilling of DOI field at the top of the page when it has been filled previously, and we're re-rendering the page
  - validating whether the user has entered their DOI or manuscript number when a journal name is present
- some reports and rake tasks that we occasionally use 
- article DOI that is stored in SOLR, so users can search our content by article DOI
- article citations in schema.org metadata
- article citations that appear in the pub updater
- related DOIs that are sent to LinkOut


sfisher (approved):
Thanks Ryan.

PR #240, opened 2020-09-30
This will merge two users the same way we always do it in the UI

sfisher:
and should make it less boring and error prone to merge two users.

I think in most cases this should work fine, but of course we should check users in the database first to be sure there is nothing weird.  You're kind of forced to check users, anyway, in process of getting their ids.


ryscher:
Thanks for making this a little easier!

PR #242, opened 2020-09-30
Add live validation for users entering it in autofill form

sfisher:
This adds checks of the DOI when it changes the identifier and saves it to DB from this form.


ryscher:
Looks good, works on my system. Sorry you had to go back over this.

PR #243, opened 2020-10-02
Ui cleanup tweaks, see #878

sfisher:
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/878

Things in this PR.

- Unlocked and updated web-console and colorize version locks in Gemfile because at least web-console was erroring with Rails 5.2 and our installed gems when trying to display error messages in development mode.
- Added localhost and port for local-dev copy of configuration environment.
- DataCite publisher always set to Dryad as well as display of Publisher always showing Dryad (and updated tests, and places it's set).
- Removing version and publisher from calls to create citation because publisher is always Dryad and version is never shown in citation now.
- Footer copyright year always shows current year.
- Subscribe to mailing list removed since we didn't have any link for that.

After we release this code then I put in a separate ticket for updating all publishers to Dryad and re-submitting all metadata to DataCite/EZID at that point.

BTW, this is the record from a newly submitted record in EZID for DataCite metadata showing Dryad (even though submitted as UCOP tenant) https://ezid.cdlib.org/id/doi:10.5072/FK2TD9WW48 .

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -62,8 +62,8 @@ gem 'http'
 # Development and testing
 
 group :development, :local_dev do
-  gem 'colorize', '~> 0.8'
-  gem 'web-console', '~> 2.0'
+  gem 'colorize'
+  gem 'web-console'
   # gem 'httplog', not needed always, but good for troubleshooting HTTP requests to outside http services from the app
 end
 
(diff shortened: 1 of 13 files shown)


ryscher:
All good. Thanks for the cleanup, Scott!

PR #245, opened 2020-10-05
Unrevert the revert commit from months ago for counter stats from live DataCite

sfisher:
This was a PR that I worked on in August but then Daniella said she didn't want it going live yet, so had to do a revert patch to take it out.  Now I'm doing a revert on the revert.

To test that it pulls from DataCite and not out of our database you can go to a record like /stash/dataset/doi:10.5061/dryad.3bc76 (against development).  Then find the identifier_Id (1460) in the database.  Then open the stash_engine_counter_stats table and delete the cache record for this identifier_id.  Then refresh the page and it should pull the stats back again from DataCite and add a cache record to the database table again and display the stats on the page.

### DIFF
--- a/spec/data/mdc-usage.json
+++ b/spec/data/mdc-usage.json
@@ -1,125 +1,476 @@
 {
-  "data": [],
+  "data": [
+    {
+      "id": "9459a75c-a45f-49f0-8e96-94cce0d7c5fd",
+      "type": "events",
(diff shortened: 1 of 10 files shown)


ryscher:
Looks good, works in my tests. Will be great to get these live numbers up again. Hooray!

PR #246, opened 2020-10-05
Better error message that says which user ids it thinks are duplicates

sfisher:
This just tacks on the users it thinks are duplicates so it's easy to look up the users or the ORCID or the emails from the database.

I should've made this a better error message the first time around.


ryscher:
Thanks for fixing this. I'm sure it will save us some time.

PR #247, opened 2020-10-09
Zenodo downloads

sfisher:
This creates downloads for zenodo files (even if not public) on the review page and also in the sharing link.

It also contains some additional methods in models and the RATs presigned URL code for zenodo and tests.


ryscher:
Thanks for doing the updates to get this working! The code changes look fine, but I encountered two issues in testing:
1. The Zenodo DOI is entered as a Related Work with type `issupplementto`. The landing page shows this as "Undefined". Is that intentional?
2. When editing a submission (as a curator), I no longer get the header box that says "You are editing XXXX's dataset." This means I don't have the option to cancel the edit, and it is easier to leave a dataset "in progress".

ryscher:
Clarification for #1 above: I was accidentally testing on this branch when merged into `main`. Currently, the `zenodo-code` branch doesn't have the Related Works on the landing page, though I expect this issue will appear when it is merged into main.

sfisher:
Thanks Ryan, I put in a ticket to fix the two bugs you noted (in the zenodo board)  https://github.com/CDL-Dryad/dryad-product-roadmap/issues/907

I'm not quite sure why tests pass in one of the places but not the other.

PR #248, opened 2020-10-12
Updating the counter token

sfisher:
Kristian sent us more new tokens again (seems like he did this just a month ago) that we need to update.

I also make the account configurable rather than hard coding the places I found it.

It looks like things I need to process and send reports to datacite are already OK in our code and in our crons (monthly task).


ryscher:
Looks good. Thanks for the updates!

PR #249, opened 2020-10-12
Search API

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/739

- Added an API endpoint for `/search`
- Added documentation for search options. Currently, this is both in swagger, and a separate "search API" doc, which is unfortunately redundant, but I find the query options difficult to read in the swagger doc, so the separate doc seems more convenient for external users.
- Corrected math in the paging system... previously, when the number of results was an even multiple of the page size, the API claimed that there was an extra page


sfisher:
Thank you , Ryan.  This was a ton of work and especially the testing was painful, I'm sure (but probably valuable).

I'm also amazed you figured out the Swagger stuff so quickly since it's really painful to edit and the documentation isn't so great (and if you use the official javascript-based editor it says everything is an error).  I think it is really made to be generated automatically rather than by hand.  😞   But i guess we're dealing with it for now, but I wish there was a better way.

I don't see any problems and it's great that you're getting the result IDs from SOLR and then just displaying records from our normal database display.

PR #251, opened 2020-10-14
Put crons in the main repository and use them instead of config repo

sfisher:
I looked at all for sensitive information and think we're good.

I needed to modify the deploy tasks since the crons from from the shared directory.  So just copy the latest from the main app over the ones there.  If we like we can get Ashley to change the puppet to use it from current instead and then can remove this task.

This should end our reliance on the config repo and allow us to update that code like the rest.

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -4,6 +4,9 @@ require 'capistrano/setup'
 # Include default deployment tasks
 require 'capistrano/deploy'
 
+require 'capistrano/scm/git'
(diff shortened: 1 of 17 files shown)


ryscher:
Looks good. I didn't do much testing, since this is mostly copies of existing scripts. Thanks for moving these files over. It will be great to finally archive the dryad-config repository.

PR #252, opened 2020-10-15
Updating harvey mudd's shibboleth entity_id

sfisher:
I think they changed it and are now outsourcing their shib to someone else.


ryscher:
Looks good. Thanks for testing it out.

PR #253, opened 2020-10-17
Reinstate sitemap w/ list of sitemaps and 1000 per page

sfisher:
This should work for the list of sitemaps and the sitemap files.


sfisher:
I'm trying to bisect the tests failing to see what is causing the problems.  I could reproduce it locally with seed 48590.  I don't think this is because of something I just added, though.

ryscher:
Looks great! This will be great to make our content more visible.

PR #254, opened 2020-10-20
Modifies the relationships for the new related identifier world and Zenodo

sfisher:
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/907

This adds a relationship of  "is part of" for supplemental information since that is the relationship D. said to use for this type.

Sets the relationship in our database so it appears under the correct category in the review page and shows it is verified.

Modifications to what we send to zenodo to take into account changes.

Test fixes for 1) longer dois (full DOI URL) and 2) relationship changes.


ryscher:
These changes look good, and test ok for me. Thanks for fixing up the spec tests.

PR #255, opened 2020-10-21
Allow zenodo (supplemental) uploads for certain journals (issns)

sfisher:
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/826

To test manually, I think the only one that currently is in the list is "International Journal of Agronomy."

The selected journals should see the extra tab "Upload Supporting Information" as well as have the page-to-page navigation allow going to it with next/prev buttons on bottom of the screen.  Without that journal, navigation should only be 3 pages.

Also has config for journal ISSNs and test updates for the new special structure.

Note: I'm also including the previous pull request in this one (so merge it first if you want to see just the changes for this one).  That PR includes some testing fixed I don't want to have to re-create again just for this PR.

### DIFF
--- a/config/app_config.yml
+++ b/config/app_config.yml
@@ -135,6 +135,8 @@ defaults: &DEFAULTS
     - 'Zimbabwe'
   funder_exemptions:
     - 'Chan Zuckerberg Initiative'
+  supplemental_info_issns: [1687-7969, 1687-7977, 1687-9309, 1687-9317, 1687-7667, 1687-7675, 1468-8115, 1468-8123,
+    1687-8159, 1687-8167, 1687-9708, 1687-9716]
   link_out:
     # LinkOut FTP information for Europe PubMed Central
     labslink:
--- a/spec/support/helpers/dataset_helper.rb
+++ b/spec/support/helpers/dataset_helper.rb
@@ -14,6 +14,12 @@ module DatasetHelper
   end
 
   def navigate_to_software_upload
+    # Sets this up as a page that can see the software/supp info upload page.
+    se_identifier = StashEngine::Identifier.all.first
+    StashEngine::InternalDatum.create(identifier_id: se_identifier.id, data_type: 'publicationISSN', value: '1687-7667')
+    se_identifier.reload
+    navigate_to_upload # so the menus refresh to show newly-allowed tab for special zenodo uploads
+
     click_link 'Upload Supporting Information'
     click_link 'Upload directly'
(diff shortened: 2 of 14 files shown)


config/app_config.yml
```diff
@@ -135,6 +135,8 @@ defaults: &DEFAULTS
     - 'Zimbabwe'
   funder_exemptions:
     - 'Chan Zuckerberg Initiative'
+  supplemental_info_issns: [1687-7969, 1687-7977, 1687-9309, 1687-9317, 1687-7667, 1687-7675, 1468-8115, 1468-8123,
+    1687-8159, 1687-8167, 1687-9708, 1687-9716]
```
ryscher, line 139: I'm a little concerned about having to manage this in the config file. Why not add it as a column in `stash_engine_journals`? It probably depends on how we're planning to roll out this feature. Will we be continually updating these settings? Or do we expect it to go live for all journals after a short pilot phase?

spec/support/helpers/dataset_helper.rb
```diff
@@ -14,6 +14,12 @@ def navigate_to_metadata
   end
 
   def navigate_to_software_upload
+    # Sets this up as a page that can see the software/supp info upload page.
+    se_identifier = StashEngine::Identifier.all.first
+    StashEngine::InternalDatum.create(identifier_id: se_identifier.id, data_type: 'publicationISSN', value: '1687-7667')
```
ryscher, line 19: So any time a test wants to access the software upload page, it must be associated with this particular journal? That could be confusing in the future. Maybe allow the ISSN to be passed in, and set a default value for the common case that the caller doesn't care? Or maybe move the process of setting the journal into a different method to make it clearer?

stash/stash_engine/app/models/stash_engine/identifier.rb
```diff
@@ -480,6 +480,15 @@ def last_zenodo_resource(copy_type:)
         .order(id: :desc).limit(1).first
     end
 
+    def allow_supplemental_info?
+      @allow_supplemental_info ||=
+        begin
+          my_internal = StashEngine::InternalDatum.where(identifier_id: id, data_type: 'publicationISSN').first
+          my_issn = my_internal&.value
```
ryscher, line 487: there is a `publication_issn` method that can take the place of `my_internal` and `my_issn`

ryscher (commented):
I love the approach of just changing the page links, and nothing else needs to be disabled. Overall, looks good, just a few minor questions/suggestions.

PR #256, opened 2020-10-22
Removes the select license and always sets it to CC-BY-4.0

sfisher:
This builds on the previous PRs and just adds additional changes, so probably best to go through those in order first.

Moved the former choose-license to a different partial so it can be resurrected when it comes back again.

Migration to add all of the non-obsolete licenses, even if they don't meet the standard of open source purity (because CC-BY wasn't in the list before).

Change the submission form to always submit CC-BY-4.0

Test to be sure it is set as CC-BY-4.0


ryscher:
This looks good, tests good. I'll leave it unmerged until the prior branch is resolved.

PR #257, opened 2020-10-23
Enables an ORCID-free test login for development

sfisher:
see https://github.com/CDL-Dryad/dryad-product-roadmap/issues/920 .


ryscher:
Thanks for enabling this. It will make it much faster to set up Dryad for testing in a lot of different situations.

PR #258, opened 2020-10-26
Zenodo code 905

sfisher:
Makes the links look like buttons which they aren't but similar to other styles we have.


ryscher:
Looks good!

PR #260, opened 2020-10-30
Curation state changes

sfisher:
This should fix problems with curation states being applied out of order because.

It is applied to the latest version of the dataset if it is possible and the state follows from a previous state.  It also will not apply an embargo or publication state if there is a version that is still being modified or hasn't gone through submission completely.

It refreshes the entire row for a dataset after a curation activity, also.

It is actually kind of hard to get bad states, but you can get one of two different new error messages if you 1) open the "change curation state" modal dialog in one window and 2) open another window and apply changes to the dataset.

There is one error because you're trying to publish or embargo (release metadata) while the dataset is still submitting or in-progress.

The other error you can get is if you change the curation status in another window and then try to apply an illegal status in the dialog you left open in the other window.

I revised quite a bit of the process and ajax and also added some tests and tried make it clearer which states follow from which other states.

I didn't add all possible simultaneous state changes in the UI, but there is a test which goes through the process of changing curation state through the UI and verifies that it saves.

### DIFF
--- a/spec/features/stash_engine/curation_activity_spec.rb
+++ b/spec/features/stash_engine/curation_activity_spec.rb
@@ -1,4 +1,6 @@
 require 'rails_helper'
+# require 'pry-remote'
+
 RSpec.feature 'CurationActivity', type: :feature do
 
   include Mocks::Stripe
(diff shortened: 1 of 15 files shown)


ryscher:
Looks good, and works well for all the tests I can think of. I can't wait to see whether the curators are able to break it!

PR #261, opened 2020-11-02
Makes edit note show up in curation notes

sfisher:
This builds on the previous pull request, so if you merge that one into main first then it should only show the changes for this PR when doing the diff.

This adds another 'in_progress' status on submission to the curation notes with the curator's note they left about editing the dataset.  The state flips to 'submitted' when it has gone through Merritt.

This also modifies the person that is given attribution for the tasks.  The curator that edits will get attribution for the events that happen around their editing and re-submission (it could be some else before--and was the last person to curate the *previous* version who go attribution for some of the new events).


stash/stash_engine/app/models/stash_engine/resource.rb
```diff
@@ -694,7 +701,7 @@ def curation_to_curation(prior_version, attribution)
       return unless %w[action_required curation].include?(prior_version.current_curation_status)
 
       curation_activities << StashEngine::CurationActivity.create(user_id: attribution, status: 'curation',
-                                                                  note: edit_histories.last&.user_comment || 'ready for curation')
+                                                                  note: 'system set back to curation')
```
ryscher, line 704: If this is a system activity, shouldn't we attribute it to the System user (id 0)? Then we could remove the logic above that calculates the attribution, which is still somewhat confusing.

sfisher: Ok.  I changed the state change to reset to curation to the system user.

ryscher:
Thanks, Scott! This will be very helpful for seeing why datasets are in particular states.

PR #262, opened 2020-11-03
Adds institutional name for submitter to the CSV output

sfisher:
Added a test and added the tenant institution name to the CSV output.

This is a pretty small change and can be tested by looking at the CSV output from the ten


ryscher:
Looks good. Thanks for adding this!

PR #259, opened 2020-10-26
Embedded submission (aka iframe)

ryscher:
For 
- https://github.com/CDL-Dryad/dryad-product-roadmap/issues/850
- https://github.com/CDL-Dryad/dryad-product-roadmap/issues/909
- https://github.com/CDL-Dryad/dryad-product-roadmap/issues/910

Everything you need should be in the included documentation. 

After discussion with stakeholders, they generally prefer opening Dryad in full separate tab/window, rather than an iframe. There is one piece of security config that would need to be added if we ever decide to support an iframe; I've left it in the PR, but commented out.

PR #264, opened 2020-11-04
Embedded submission

ryscher:
This replaces https://github.com/CDL-Dryad/dryad-app/pull/259, accounting for updated requirements.

For
- CDL-Dryad/dryad-product-roadmap#850
- CDL-Dryad/dryad-product-roadmap#909
- CDL-Dryad/dryad-product-roadmap#910
- CDL-Dryad/dryad-product-roadmap#926

Everything you need should be in the included documentation.

After discussion with stakeholders, they generally prefer opening Dryad in full separate tab/window, rather than an iframe. There is one piece of security config that would need to be added if we ever decide to support an iframe; I've left it in the PR, but commented out.


documentation/apis/embedded_submission.md
```diff
@@ -0,0 +1,244 @@
...
+  "editLink": "/stash/edit/doi%3A10.7959%2Fdryad.83bk3jc0"
+  }
+```
+
+Allow users to edit metadata about the dataset, and upload data files
+---------------------------------------------------------------------
+
+The manuscript system can now redirect the author to Dryad by opening
+the `editLink`. This URL may be opened in a new browser tab, or
+directly in the tab where the author has been editing their manuscript
```
sfisher, line 113: I thought it would only be for a new window at this point.

documentation/apis/embedded_submission.md
```diff
@@ -0,0 +1,244 @@
...
+
+The author will then proceed through the normal Dryad submission
+process. They will be able to edit metadata that has been
+pre-populated by the journal, and upload any files.
+
+When the data submission is complete, the author will be redirected
+back to the correct page in the manuscript system. If you have chosen
+to open the Dryad submission in a new window, and you want the new
+window to simply close, specify the `returnURL` as a page that closes
+itself immediately upon opening.
```
sfisher, line 133: This is nice flexibility, since I imagine many systems will want to open Dryad in a new window and then close the window when they're done.  It also gives them some kind of notification that it's done by that URL being hit.

stash/stash_api/app/models/stash_api/dataset.rb
```diff
@@ -104,7 +105,17 @@ def add_license!(hsh)
...
+      ensure_edit_code
+      return unless version.resource.permission_to_edit?(user: @user)
+
+      hsh[:editLink] = "/stash/edit/#{CGI.escape(@se_identifier.to_s)}/#{@se_identifier.edit_code}"
+    end
+
+    def ensure_edit_code
+      return if @se_identifier.edit_code
+
+      @se_identifier.edit_code = SecureRandom.urlsafe_base64(10)
```
sfisher, line 117: This urlsafe_base64 looks good.  IDK if we use it elsewhere, but maybe we should.

stash/stash_engine/app/controllers/stash_engine/resources_controller.rb
```diff
@@ -85,7 +86,16 @@ def update
     def destroy
       resource.destroy
       respond_to do |format|
-        format.html { redirect_to return_to_path_or(dashboard_path), notice: 'The in-progress version was successfully deleted.' }
+        format.html do
+          notice = 'The in-progress version was successfully deleted.'
+          if session[:returnURL]
```
sfisher, line 91: When you return, do you need to clear the `session[:returnURL]` just before a redirect in case they come back to Dryad and are still logged in and redirected somewhere?  I know it's an unlikely scenario.

sfisher (approved):
This looks good to me, Ryan.  You put in a lot of work to get this together.  Thanks.

documentation/apis/embedded_submission.md
ryscher: It will be a full browser page. But we don't really have a way to control whether it's the same page as the manuscript system or a separate tab/window. That depends on how the caller sets the `target` of the link in their page.

stash/stash_api/app/models/stash_api/dataset.rb
ryscher: I copied this from the code that generates the peer-review links...

stash/stash_engine/app/controllers/stash_engine/resources_controller.rb
ryscher: Good idea. I'll fix that.

PR #266, opened 2020-11-05
Switch ftp to sftp

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/820

Since NCBI doesn't have a test server, we cannot fully test this until it is deployed to production, but I have tested it with transferring some dummy files, and those appear fine.

Note that this does *not* switch the transfer mechanism for LabsLink. Although that system uses the same files, they have not converted their servers to SFTP yet.


sfisher (approved):
Looks good to me.  Thanks Ryan.

Weird that the session closes automatically.

PR #269, opened 2020-11-10
Update new curator bios and photos

LoriEP:


ryscher:
Looks good. Thanks for the updates!

PR #271, opened 2020-11-10
Allow peer_review to be edited when resource in peer_review status

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/869

### DIFF
--- a/spec/models/stash_engine/identifier_spec.rb
+++ b/spec/models/stash_engine/identifier_spec.rb
@@ -448,6 +448,12 @@ module StashEngine
         expect(@identifier.allow_review?).to be(false)
       end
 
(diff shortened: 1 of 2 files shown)


sfisher (approved):
I think this looks good.  Does this change the list of allowed states when the curators get the popup to change the curation status?  There are some hashes/arrays in a constant in the curation that express which states are allowed from which other states.

ryscher:
The change in this PR only affects whether the "Peer Review" section is shown within the submission system. The `CURATOR_ALLOWED_STATES` only affect the admin UI and do not need to change. We assume that curators have a reason for putting items into a review state.

PR #276, opened 2020-11-18
This is the accessibility rollup of a bunch of small things

sfisher:
See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/603 .

I corrected most of the File Upload things, but Daniella said it wasn't necessary because we'll be redoing it soon, anyway.  The links instead of buttons for "upload by file" and "upload by url" is a little more convoluted and is in the "supplemental materials" branch which still hasn't been moved in.  I didn't want to try to separate it all out of that branch and since it's a little involved.

The rest of the fixes should be in there.  The download accessibility one is in another ticket.


ryscher:
This all looks fine. Thanks for dealing with these tedious changes.

PR #277, opened 2020-11-20
Get working under Python 3.7 and log cleanup fix

sfisher:
These are two separate changes.

1. Sets the environment variables and other things necessary to get it working under Python 3.7.9 with virtualenv which Ashley installed.
2. The remote counter logs (older than 60 days) from the other server were not getting deleted, just the one on the server.

For the deletion, it is hard to test this in our testing framework, but I duplicated the code changes onto the server and tested it there.  If you wanted to try it there you could add a file like `counter_2020-09-01.log` to the directory on `dryad@uc3-dryaduix2-prd-2c:~/apps/ui/current/log` and then run `bundle exec rails counter:remove_old_logs` on `2a` and then check that the older counter log file on 2c that you created is now gone. 

You'll notice that the "remove_old_logs_remote" is the same logic and structure as the "remove_old_logs" with only three real differences.  1) It iterates through each remote host defined, 2) it gets the list of files with an ssh command instead of from a local command and 3) it deletes with an ssh command instead of a local File.delete(fn) command.

### DIFF
--- a/cron/counter.sh
+++ b/cron/counter.sh
@@ -25,8 +25,9 @@ bundle exec rails counter:combine_files
 # set up python and run counter-processor (maybe twice)
 # ---------------------------------------
 echo "Running counter-processor"
-export PATH=$HOME/opt/bin:$PATH
-export PYTHONPATH=$HOME/opt/bin/python-3.6.9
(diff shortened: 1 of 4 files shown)


ryscher:
Thanks, this is great. I didn't realize the servers had permission to ssh to each other.

PR #278, opened 2020-11-20
Adding log check for counter upload logs

sfisher:
This will wait until Thursday every week and then check if the log looks like it has successfully processed and submitted since last week.

Should work OK because even if it fails repeatedly I think it only notifies once unless the service is flapping with failure/success/failure.


ryscher:
Looks good. It will be nice to have this more automated.

PR #279, opened 2020-11-23
Fixing manifest upload to correctly decode encoded URLs.

sfisher:
CGI unescape all filenames we extract from URLs so they no longer have percent encoding in them.  This also allows the filename sanitizing and renaming to work correctly.


ryscher:
Thanks for fixing this! It should prevent a few problems.

PR #281, opened 2020-11-25
Editorial Manager bridge API

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/718

This is the basis for supporting Editorial Manager's API workflow. I expect it will be refined as we iterate with the Editorial Manager team.

Takes requests in the Editorial Manager format, transforms them into the format used by the Dryad API, runs the regular process, and reformats the results into the output format desired by Editorial Manager.

A single endpoint (`em_submission_metadata`) is used. It handles both creating a new dataset (POST) or modifying an existing dataset (PUT), though I believe Editorial Manager will be using POST for both types of request. The endpoint is also agnostic about whether the request is the minimal "deposit" request or the more fully-specified "submission metadata" request.

To test: use the normal format of a submission request with the new endpoint, like `curl --data "@my_metadata.json" -i -X POST https://<domain-name>/api/v2/em_submission_metadata -H "Authorization: Bearer <token>" -H "Content-Type: application/json"`. But ensure your JSON document has the format of the samples included in this PR.

Note that I've changed the Doorkeeper config so that all new requests for tokens are assigned the scope `all` by default. I have asked EM to use a more descriptive scope, but have not received a final decision on that yet. We may not want to continue using a default scope... at some point, it would probably be better to let EM use `all`, and assign more specific scopes to other users. But we were not using the scopes before, so I think this is ok for now.

### DIFF
--- a/documentation/apis/emDeposit.json
+++ b/documentation/apis/emDeposit.json
@@ -1,0 +1,45 @@
+{
+    "_comment": "This is a sample file to illustrate the format of the data sent by the Editorial Manager API. Although Dryad supports this API, it is not recommended for general use, as our processing may change to suit the needs of Editorial Manager.",
+   "publication_code": "DEMO170",
+   "journal_full_title": "Journal of Demonstration for Version v17.0",
+   "document_id": "12345",
+   "authors": [ {
+      "first_name": "Amy",
+      "last_name": "Author",
+      "email": "amy.author@test.demo",
+      "people_id": "126",
+      "orcid": "1234-1234-1234-1234",
+      "corresponding": true,
+      "custom_user_ids": [
+         {
+            "custom_id": "12065346",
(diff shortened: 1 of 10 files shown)


ryscher:
Added feature... calls via this API can only update metadata while the dataset is `in_progress`. Once it has been submitted (presumably by a human user), the metadata must be edited via the UI. Note that although the metadata is locked, the curation status of a `peer_review` dataset may be changed.

sfisher (approved):
Thank you, @ryscher .  This looks good to me.  Thanks for dealing with this.

Thanks for the great tests and even faking the metadata with Faker.  Looks like lots of work. I also appreciated the CURL example which made it easy for me to do a basic test with the test data.

Seems to work it looks good to me.

PR #282, opened 2020-11-30
Modify the query for Merritt Migration info to include version

sfisher:
David was mentioning the versions, so I think this may be helpful to him.  Though, I think if we have had to do any manual intervention (in a few rare cases) then the number of the versions may differ.


ryscher:
Looks good. Hope it provides the needed information.

PR #283, opened 2020-12-02
Related Works update for the API

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/925

- Changed the meaning of the `relationship` field in the
  API so it uses the `work_type` enumeration instead of the DataCite 
  relationships. Technically, this change breaks functionality, since it's using a
  different vocab. I don't think anyone was using the old vocab, but
  if it seems like an issue, I could add a mapping from the old vocab
  to the new one.
- Changed the http header restrictions on most API methods --
  instead of "headers must specify JSON", changed to "if headers are
  specified, they must allow JSON". Since we only support JSON for
  most calls, we can assume that a request with unspecified format
  is using JSON. This simplifies using the API. Calls that don't
  require authentication can now be resolved in a browser (or other
  tool) without extra headers.

### DIFF
--- a/documentation/apis/dataset_metadata.md
+++ b/documentation/apis/dataset_metadata.md
@@ -45,8 +45,8 @@ Descriptive fields
 Other fields with descriptive metadata include:
 - `relatedWorks` - Relationships to other objects may be specified by
   giving a `relationship` and identifier information. The allowed
-  values for `relationship` are the same as for DataCite's
-  `relationType` field.
+  values for `relationship` are: `undefined`, `article`, `dataset`, `preprint`,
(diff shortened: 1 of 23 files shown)


ryscher:
I used `work_type` instead of `work_type_friendly`, because it's more machine-readable. But I realize that this introduces the value `supplemental_information`, which does not follow the CamelCase style of the rest of the API. Thoughts?

sfisher (approved):
Thank you for doing this, Ryan.  The strict requirements for json headers have been bugging me for a while and this will make things easier and better for everyone.

PR #284, opened 2020-12-02
Email validation issue (author ordering)

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/949

I'm not 100% certain this fixes the problem, since I can not reproduce the issue, but this seems like the most likely cause. Authors in the UI, and generally throughout the system, are ordered by the default database ordering (presumably `id`). Prior to this change, the validation step was explicitly using a different ordering (`created_at`). There is the potential that the two orders could get out of sync, which would cause the validator to operate on a different author than expected.

### DIFF
--- a/spec/models/stash_datacite/completions_spec.rb
+++ b/spec/models/stash_datacite/completions_spec.rb
@@ -518,6 +518,9 @@ module StashDatacite
             author.author_email = nil
             author.save!
           end
+          @resource.save!
+          byebug
+          @completions = Completions.new(@resource)
           warnings = completions.all_warnings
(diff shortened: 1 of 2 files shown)


ryscher:
I was able to confirm that this is indeed the issue. When metadata is imported (from a DOI or manuscript number), the `created_at` values are all identical, because millisecond values are not stored. This can cause a sort by `created_at` to be somewhat random, why the default sort by `id` is stable.

sfisher (approved):
This is a good small change that makes sense.  IDK why we ordered the way we did, but this seems good.

PR #285, opened 2020-12-03
Embargoed sharing link

ryscher:
When providing metadata through the API, do not include the peer-review `sharingLink` if the item is in status `embargoed`. In general, the API assumes that anyone who has rights to view the metadata will also have rights to download the item, but this is not the case for `embargoed` items.

Note that `embargoed` items will still display the links for downloading files, but these links validate permissions on their own, and disallow downloads if the user does not have correct permissions.


sfisher (approved):
Thank you, Ryan.

This is a super helpful fix.

PR #286, opened 2020-12-04
Updating the forms from using form_for and form_tag to use the Rails 6 form_with

sfisher:
For the most part this went smoothly and I tried to test each form afterwards to be sure it still worked.  The rules for changing all this are.

- Form_for forms should have a "model:" param specified instead.  Where we were using symbols for the model no longer work, so seem to need to have _ModelName.new_ instead of the symbol.
- Form_tag forms would just have the _url:_ param and no model, but a bunch in the admin area didn't have special routes and were using the "action: controller:" syntax to specify the action and controller to use.  these had to be converted to _url_for_ with the model and controller and _only_path_ as true (so as not to generate full URL with domain).
- All forms are now "remote: true" by default which is what we want for AJAX stuff, but is annoying for non ajax.  the opposite of "remote: true" is actually "local: true" rather than "remote: false" which seems kind of ridiculous to have two different keywords for opposite states of the same thing, but whatever.
- I found a problem saving curation notes while testing, but only if a dataset had never been submitted, so added some fallback code to write to the in-progress version if a note was needed and no submitted version was available.

I manually tested form submissions after making the changes to be sure they worked and the tests all seem to work still.

### DIFF
--- a/stash/stash_datacite/app/views/stash_datacite/authors/_form.html.erb
+++ b/stash/stash_datacite/app/views/stash_datacite/authors/_form.html.erb
@@ -1,7 +1,7 @@
 <%# the local variable 'author' should be passed in to this partial when there are multiple per page.
   # the 'path' to submit to should also be passed in. %>
 <% form_id = unique_form_id(author) %>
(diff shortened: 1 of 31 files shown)


ryscher:
Looks great! This seems straightforward, but I know it took a lot of effort to get all the details right. Thanks for slogging through it.

PR #287, opened 2020-12-07
Adding Rancho Santa Ana as author institution for Claremont admins

sfisher:
This adds Rancho Santa Ana Botanical Gardens to the view for the administrators for including datasets to look at


ryscher:
Minor update, great.

PR #288, opened 2020-12-07
constrain matching to only valid strings to old Dryad redirects

sfisher:
fixing routes that redirected.  Constraining what they'll accept only valid strings

See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/973 .

It seems like the old way accepted anything with "/resource" or something like that.  Now it constrains accepting the URLs so they are only processed if they match valid values for the full path, including all IDs.  The URI module errors were from it accepting too much which were invalid URLs and caused internal URI module to throw errors somewhere in Rails.

I think this is fine and made them more restricted to the exact expected values.

Ryan can you validate that this is all correct for the Dryad classic URLs that this is doing redirects for and everything still works?

While we could theoretically do things like remove spaces from URLs where they are not supposed to be, I'm not inclined to do that since no one else in the world babysits users that type bad URLs.  When people put in bad URLs anywhere else on the internet they get a 404 if they can't bother to get the URL right.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -105,19 +105,17 @@ Rails.application.routes.draw do
   get '/themes/Mirage/docs/:doc', to: redirect('/docs/%{doc}.%{format}')
 
   # Routing to redirect old Dryad landing pages to the correct location
+  # Regex based on https://www.crossref.org/blog/dois-and-matching-regular-expressions/ but a little more restrictive specific to old dryad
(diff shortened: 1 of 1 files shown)


ryscher:
This looks good. I tried a lot of examples of old-style Dryad URLs, and they all transformed correctly. Thanks for cleaning this up. I hope it saves us from some of the silly email messages.

PR #289, opened 2020-12-07
Fix URI errors when people enter invalid URLs in the upload box

sfisher:
The URIs were erroring when being passed through the URI translator (that tries to convert URIs to special URIs for things like Google Drive).

This changes the translator so that it just passes the bad URI through without doing anything if it gets an invalid URI error.


ryscher:
Looks good, tests fine. Thanks for fixing this!

PR #292, opened 2020-12-08
Display ORCID invite links to curators

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/944

Once a dataset has been published and the initial ORCID invitations have been sent to co-authors, curators who edit a dataset will see the ORCID invitation links that have not yet been used.

PR #293, opened 2020-12-09
Only show ORCID invite link when email is present

ryscher:
Slight fix for #292.

When an author's email address is NULL, the invite link would cause a match to all NULL emails in the same dataset, which would attach the ORCID to more authors than desired. This requires the email address to be present, so the invite link will connect to the correct author.


sfisher (approved):
Thank you for catching this, Ryan.  Looks good.

PR #294, opened 2020-12-11
Making return-to after editing work gloriously with many windows open

sfisher:
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/933

This traps the edit links for curators, adds something to the form that gets submitted and then the controller stores a return to url in the session for each open dataset they are editing at once.

Fixed problem with the old-style return-to (one in session for user).

Added tests to be sure both the editing and cancel workflows return to the page they came from afterwards.

### DIFF
--- a/spec/models/stash_engine/curation_activity_spec.rb
+++ b/spec/models/stash_engine/curation_activity_spec.rb
@@ -55,7 +55,7 @@ module StashEngine
 
         # get rid of callbacks for adding one and testing
         @curation_activity1 = create(:curation_activity, resource: @resource)
-        CurationActivity.set_callback(:create, :after, :submit_to_datacite)
+        # CurationActivity.set_callback(:create, :after, :submit_to_datacite)
       end
 
       it "doesn't submit when a status besides Embargoed or Published is set" do
--- a/stash/stash_datacite/app/controllers/stash_datacite/resources_controller.rb
+++ b/stash/stash_datacite/app/controllers/stash_datacite/resources_controller.rb
@@ -70,10 +70,10 @@ module StashDatacite
 
       resource.reload
 
-      if session[:returnURL]
-        return_url = session[:returnURL]
-        session[:returnURL] = nil
-        redirect_to(return_url)
+      if session["return_url_#{resource.identifier_id}"]
+        return_url = session["return_url_#{resource.identifier_id}"]
+        session["return_url_#{resource.identifier_id}"] = nil
+        redirect_to return_url, notice: "Submitted updates for #{resource.identifier.to_s}, title: #{resource.title}"
       else
         redirect_to(stash_url_helpers.dashboard_path, notice: resource_submitted_message(resource))
       end
--- a/stash/stash_datacite/app/controllers/stash_datacite/resources_controller.rb
+++ b/stash/stash_datacite/app/controllers/stash_datacite/resources_controller.rb
@@ -57,6 +57,7 @@ module StashDatacite
       end
     end
 
+    # rubocop:disable Metrics/AbcSize
     def submission
       resource_id = params[:resource_id]
       resource = StashEngine::Resource.find(resource_id)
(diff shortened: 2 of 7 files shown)


spec/models/stash_engine/curation_activity_spec.rb
```diff
@@ -55,7 +55,7 @@ module StashEngine
 
         # get rid of callbacks for adding one and testing
         @curation_activity1 = create(:curation_activity, resource: @resource)
-        CurationActivity.set_callback(:create, :after, :submit_to_datacite)
+        # CurationActivity.set_callback(:create, :after, :submit_to_datacite)
```
ryscher, line 58: Was this causing a problem?

stash/stash_datacite/app/controllers/stash_datacite/resources_controller.rb
```diff
@@ -70,14 +71,19 @@ def submission
...
-        return_url = session[:returnURL]
+      # There is a return URL for a simple case and backwards compatibility (only for for whole user and for journals).
+      # There is also one for curators and need to return back to different pages/filter setting for each dataset they
+      # edit in one of dozens of different windows at the same time, so needs to be specific to each dataset.
+      if session["return_url_#{resource.identifier_id}"] || session[:returnURL]
+        return_url = session["return_url_#{resource.identifier_id}"] || session[:returnURL]
+        session["return_url_#{resource.identifier_id}"] = nil
         session[:returnURL] = nil
-        redirect_to(return_url)
+        redirect_to return_url, notice: "Submitted updates for #{resource.identifier}, title: #{resource.title}"
```
ryscher, line 81: I notice you removed the parens here, but not in the call immediately below. Is there a preferred style? Or reason to use one or the other? (I don't care either way; this is just one of those Ruby things I haven't internalized yet, so I'm interested in learning more.)

ryscher (approved):
This looks great, and does indeed work gloriously. 

I had a couple of minor questions, though it's fine to merge any time.

spec/models/stash_engine/curation_activity_spec.rb
sfisher: I ended up having a bunch of trouble (like half a day) getting tests to pass consistently (seemed to pass and then not pass depending on test ordering of tests).

The conclusion I've come to is that it's much better to either use mocks or other ways to work around problems or side effects or things you don't want to happen in a test rather than modifying any object behavior directly.  Mocks are automatically rolled back after each test so they don't interfere with other tests while direct modifications to objects often aren't rolled back after a test and then cause order-dependent and intermittent test failures.  I called a bunch of pre-existing mock methods somewhere else to avoid all the side-effects of the callback.

I believe this was set to re-insert a callback method and I couldn't find where it was removed to begin with.  IDK if it was making two copies of the callback in some cases or what.  I think it's probably better to just mock out the method the callback uses to do nothing where you don't want it rather than removing it completely to get tests to be sane.

In the few cases where I've had to do something that isn't an automatically rolled back as part of a mock (like writing a file or something), I've tried to add `before(:each)` and then a corresponding `after(:each)` to explicitly clean up any weird state isn't left lying around that might break other tests depending on ordering.

stash/stash_datacite/app/controllers/stash_datacite/resources_controller.rb
sfisher: Yeah.  This could be done either with or without parenthesis.  Parentheses are almost always optional in Ruby method calls unless there is ambiguity about where one call begins or ends.  For nested method calls, they're probably required or at least a very good idea (at least on inner calls) to avoid ambiguity or if you're using them to order something.

Maybe we should be consistent and I'd be ok with that if we want to.  From a lot of what I see on the internet, it seems like most of the redirect/rendering type calls have parenthesis left off and I'm probably not that consistent about it.

Almost everything in Ruby is a method call even items that look like keywords in other languages are often some kind of special method call that doesn't require a dot.  I think maybe there is a preset list of these kinds and users can't easily define new method calls without dots.  But because methods can be overridden in Ruby then they can be overridden (and Rails does some overriding or extending of base objects to make them more flexible).

For instance something like == might be a method call on something like the integer class rather than a language construct.  Most classes will define things like ==, =, +, etc to work with their own data type.  You could write something like `1.==(2)` or `1.== 2` or `1 == 2` and they will do the same thing.  So part of the reason all these built-in/operator classes work and look normal is because parenthesis aren't required for method calls.

PR #295, opened 2020-12-12
Making rake tasks to update to david loy's ARK update format

sfisher:
The example update format is in the fixture in here <doi>|<old-ark>|<new-ark>

We have it embedded in the merritt download URL it gives us so we need to find and update those URLs in the resource.

It is annoying and tricky because it's embedded in a URL and also URL encoded, but believe this works and included tests.


ryscher:
Looks good. Thanks for dealing with this messiness.

PR #298, opened 2020-12-14
Update to have hiding report for Zenodo

sfisher:
This will give a list of items that have a last embargo state after a publish state.  This means they were published to zenodo but then need to be hidden because they went back to embargo.  Also takes into account if they were sent (because only things since ~july or something were).


ryscher:
Thanks! This will be useful as we sort out policies with Zenodo and the curation staff.

PR #299, opened 2020-12-16
Add master key from SSM so Rails uses it

sfisher:
This retrieves the master key from SSM and puts it in master.key upon deployment.  So any changes to the master keys should change on the file system for rails to use after the next deploy.

This was conceptually and otherwise pretty simple aside from the problems with the environment problems and troubleshooting.  The environment variables on the servers were not useable when deploying from a different server besides the local one.  So I needed to add them to my capistrano deploy environments as variables.

I also had trouble getting the SSM gem from the Merritt working correctly and it also seemed kind of overkill when I can do the same functionality that we were using from it in about 4 lines of code and it was massively overcomplicated and heavy for our simple use case.  We are using the same paths on SSM it uses, so could transition to use it sometime in the future if we needed to.

Also added a couple capistrano tasks for troubleshooting.


ryscher:
Looks good. I tested this on dev, and it works fine. 

Do you want to disable the Capistrano task `my_linked_files`? This symlinks any keys that are found in the shared directory, which means that currently, updating the key for any release causes it to be changed for all releases.

PR #300, opened 2020-12-17
Documentation updates -- moving content from the wiki (and elsewhere)

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/642

Moving some technical documentation from the old wiki into this folder, in preparation for decommissioning the wiki server.

PR #301, opened 2020-12-18
Adding fee waiver info for country

sfisher:
I think this is right, can you confirm Ryan?

=fun


ryscher:
Yes, this is correct, but I added something more comprehensive in the documentation branch...

ryscher:
This info is now included in #300

PR #302, opened 2020-12-21
Fixing the missing map

sfisher:
This was an asset pipeline issue that only appeared on stage/production.

Seems caused by missing library in Geoblacklight. Handlebars needed to be separately loaded in gemfile and then required separately from Geoblacklight.  This corrected the probem, even though it seemed unrelated to the map loading and worked when the other files loaded separately.

You can test on stage since I deployed this branch there to test and see the map loaded.

Along the way I also did a little cleanup of another error or two and update the Geoblacklight config to match their newer defaults from Github.


ryscher:
Thanks for fixing this!

PR #303, opened 2020-12-22
Prevent stats from being uncached and just return what is in the database

sfisher:
Changes:

- Uncommented the code that re-populates stats into the database after weekly counter run.
- Commented out calls to `update_if_necessary` which checks cached value and updates if it is old, so it will never update in the UI itself.
- A bunch of small rubocop (auto) fixes which seemed to be left over from a couple other PRs and I didn't notice when tests stopped running.

I ran tests against this locally (since no travis) and it passes.


ryscher:
Looks good. Runs on my machine.

PR #304, opened 2020-12-29
Default the return_url to the dashboard

ryscher:
If the return_url is not set by the javascript, default it to the user's dashboard.

This addresses the issue reported in https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1016


sfisher (approved):
Smart, yet simple fix.  Thanks @ryscher.

PR #306, opened 2021-01-04
Reinstate database backups

ryscher:
Database backups have not been running for a while, because the rake file was not processing the database config as an ERB.


sfisher (approved):
Makes sense that it wasn't parsing and running the dynamic code from this file.

PR #308, opened 2021-01-06
prevent updated_at from changing in Merritt update script

sfisher:
Minor changes with comments, test and preventing updated_at time from changing with changes in the URL for merritt

BTW, I ran tests against this locally and they passed.


ryscher:
This looks good. Thanks!

PR #309, opened 2021-01-07
changes to be sure collection name in sword update url is also changed

sfisher:
Sorry about this additional update for this.  The requirement for this wasn't clear to me since I just saw the DOI in the URL and didn't spot that they had embedded more of the Merritt internal data into the URL (collection name) also.

I updated the code for this as well as tests and ran tests locally which show it's working.

I also went and manually changed the URLs for the ~10 items we changed yesterday and tried a new version submission for one of them which worked in the new collection.


ryscher:
Looks good. Sorry it's so messy. I hope these fixes make it easier to migrate the other collections now.

PR #311, opened 2021-01-07
Monitor DB backups

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1031

Add a dependency checker to ensure database backups are occurring regularly. Also improve the rake task for setting up the dependency checkers, so new checkers can be added more easily.

Before testing, re-initialize the dependency checkers in your database with `rails status_dashboard:seed`


sfisher (approved):
This looks good to me.  Thanks, Ryan.

PR #314, opened 2021-01-12
Note that the identifier can be appended to the URL

ryscher:
Add a comment to the docs for the Editorial Manager API, indicating that the dataset identifier should be included when making a request to update metadata.


sfisher (approved):
👍

PR #315, opened 2021-01-13
Tenant for Indian institute of Technology Dhanbad

sfisher:
This adds the tenant and also creates a new authentication strategy called "author_match" along with documentation and tests.


ryscher:
Looks good. Thanks for finding a way to work around the shibboleth requirements. That will be useful in the future.

PR #316, opened 2021-01-13
Rake task and documentation for embargoing at zenodo

sfisher:
Embargoes at zenodo, I needed to do that for another previously published item.

The smaller things this changes.

1. Allows arbitrary metadata to be passed into the zenodo update method, so it can take existing metadata already there without having to regenerate it (which is our usual submission/update case).
2. Puts the series of steps needed at zenodo into a rake task so we don't have to do it manually or email Alex every time.
3. Added to the documentation about how to run the rake task after the other queries that embargo something. (In normal circumstances the extra copy isn't sent to zenodo until published).

I didn't add a test for the rake task since it mostly is just validation and calls a few steps from the stash/zenodo_replicate/deposit class (basically *get*, *reopen* for editing, *update metadata* with embargo set, *re-publish*).

### DIFF
--- a/documentation/server_maintenance/troubleshooting.md
+++ b/documentation/server_maintenance/troubleshooting.md
@@ -147,6 +147,13 @@ update stash_engine_resources set publication_date='2020-07-25 01:01:01' where i
 update stash_engine_identifiers set pub_state='embargoed' where id=;
 ```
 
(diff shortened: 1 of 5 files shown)


ryscher:
This is great! Thanks for fixing this so we don't have to worry about the Zenodo copies getting out of sync when we have to make changes.

PR #317, opened 2021-01-15
Convert from TravisCI to GitHub Actions

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/999

Remove configurations for Travis and replace them with somewhat equivalent configurations for GitHub Actions. 

I have broken the test suite into four parallel pieces. This allows us to get results faster, at the cost of some duplicated overhead per test run. The duplicated overhead is often 45 seconds per piece, when the gems are retrieved from the cache, but it can be as much as 6 minutes if the cache misses. We will need to keep an eye on this and see whether the duplicated work is pushing us too close to our credit limit. At current time, I don't believe it will be a problem. 

Currently, each piece of the test suite runs as a separate "workflow". It may be possible to combine the pieces into a single workflow, so little work is duplicated, but my initial attempts at this were unsuccessful.

I left the test database as `dashv2_test`, with username `travis`, since this is already working on our servers. If we decide to change it, we will need sysadmin assistance. Note that the Ubuntu image that is installed comes with a database that has a default `root` user, so our scripts create the `travis` user before the RSpec tests are run.

No effort should be needed to install/use this new setup; it works automatically with GitHub. It is currently set to run the testing with each push and each pull request. (We may want to remove the `push` setting in the future.)

Many thanks to @briri -- his configuration files for using GitHub Actions with DMP Tool were an excellent starting point, and made this work much easier.


sfisher (approved):
Thanks, Ryan.  This is looking good to me.

PR #318, opened 2021-01-15
In API, do not "auto match" affiliations to the closest ROR entry

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1046

Disables automatic matching for affiliation names submitted via the API, and adds some documentation about how affiliations are implemented (to make these issues easier to solve in the future).


sfisher (approved):
Thanks, Ryan.

Also thanks for adding docs.

PR #320, opened 2021-01-15
Improve wording of withdrawn message

ryscher:
Improve wording of this message to reduce curator confusion.


ryscher:
Removed a test that had been broken by #318 -- that test is no longer needed, because it was testing the "auto-matching", an removing that feature was the entire point of #318.

sfisher (approved):
Thanks, Ryan.

PR #321, opened 2021-01-15
Adding Kyoto University as partner

sfisher:
This is all working, though I'm getting an API error about RORs.  I think this will correct itself once I merge in some bugfixes from main.

They are ususual since they have one testing login server  and one real one.  (why there are two different URLs in there).


ryscher:
Looks good.

PR #322, opened 2021-01-18
Add Tenants -- Macquarie and U Minnesota

ryscher:
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/983
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1029


sfisher (approved):
Thanks, Ryan.  I looked over the configuration and it seems good to me.

On a random note, I wonder if we should convert the our_community page to a non-markdown page and just make it an html.erb file.  I think we did that a long time ago for people uncomfortable with HTML for editing, but I see that Tracy mostly edits in HTML, anyway.  If we made it dynamic, we could just output the list of partners from our configuration rather than manually editing it and trying to keep it up that way.

PR #324, opened 2021-01-18
Set github actions to ignore pushes

ryscher:
Only run the actions for pull requests.


sfisher (approved):
I think this is a good change.

PR #325, opened 2021-01-18
Fix rubocop issue with prior merges

ryscher:
Minor rubocop update for #316, since that PR was created before the GitHub actions were running correctly.


sfisher (approved):
Thanks Ryan.

The type of quote thing is kind of annoying to me and having to think about contents of the string (will it use interpolation?) in order to choose the quote type (single or double).  I know Brian Riley and their team decided to just use double quotes everywhere unless there was a special need for single quotes.

I suppose it's mostly not a big deal and the autofix switch for rubocop will change it based on the string contents, but I don't find being picky about quote type to really improve any kind of readability and I suspect the execution time difference between the two string types is negligible.

PR #329, opened 2021-01-22
Transform github blob urls into the raw format

sfisher:
So that people don't get web pages saved when they want a raw file.

See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1061 .

### DIFF
--- a/spec/models/stash/url_translator_spec.rb
+++ b/spec/models/stash/url_translator_spec.rb
@@ -50,6 +50,11 @@ module Stash
       expect(translator.original_url).to eql('cat dog')
     end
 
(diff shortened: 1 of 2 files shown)


sfisher:
It seems like the tests keep failing with 

```
Exiting: pid file already exists /home/runner/work/dryad-app/dryad-app/stash/stash-notifier/state/test.pid
```

Could this file not be deleted between runs or something else?  IDK, strange.

sfisher:
Got it working by rerunning tests a couple of times.  🤷‍♂️

ryscher:
Looks good, works in my tests. Thanks for the update.

PR #332, opened 2021-01-26
Add role for tenant_curator (Institutional  curator)

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1051

Adds a new role, `tenant_curator`, which has the privileges of a tenant `admin`, plus the ability to edit datasets belonging to the tenant.

To test, set your user account to the `tenant_curator` role. On the curation page, you should only see datasets from your tenant. These datasets should be fully editable.


sfisher (approved):
Thanks for finding all the places we had security and adding the new security methods.

This looks good.

PR #333, opened 2021-01-26
Add "progress" format to remaining tests

ryscher:
I forgot to add these into every test. It's just nice to see the line of periods in the output to know that the tests are actually running.


sfisher (approved):
Thanks.

PR #335, opened 2021-01-27
Reduce warnings when running GitHub actions

ryscher:
Making some of the gem versions more precise to avoid all the messy warnings that rspec produces in the GithHub actions.


ryscher:
Ok @sfisher, this is ready.

sfisher (approved):
Thanks for all these updates.

PR #337, opened 2021-01-28
Fixes capistrano deploy

sfisher:
Capistrano wouldn't deploy while the listed gems and the Gemfile.lock were different and the Gemfile.lock changes were not checked into the repo.


ryscher:
Oops. Thanks for fixing this up.

PR #339, opened 2021-01-30
updating FAQ and submission_process pages

tracykteal:


tracykteal:
I tried to check all my open and closing of tags and quotes, but I'm not sure that I caught them all.

sfisher (approved):
These look good.  I made a few corrections to closing tags.  We can deploy and look to be sure it looks right and edit again if there are any problems.

PR #340, opened 2021-02-01
Allow ROR affiliations to be converted to non-ROR

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/736

Allows users to update affiliations regardless of whether the previous affiliation had a ROR.

To test: 
- enter the dataset submission system
- enter an affiliation that is associated with a ROR (appears in the autocomplete list)
- change to a different page of the submission system, and return to the Describe page
- change the affiliation to garbage characters (which won't match a ROR)
- change to a different page of the submission system, and return to the Describe page
- verify that the garbage characters still appear, and they have the asterisk appended


sfisher (approved):
Thanks for making these change, Ryan.  These look good.

PR #341, opened 2021-02-01
Fixing the edited "static" pages

sfisher:
- Reformatted so that lines are reasonable length instead of run-on paragraphs all in one line.  Indents of html makes sense so it is readable.
- Fixed a couple errors, including a link that didn't have a close quote.
- Linked to the FAQ question from one place in the file uploads.


ryscher:
Looks good, all pages render properly. Thanks for cleaning up the formatting.

PR #342, opened 2021-02-02
Fixing all UC Campuses to use Dryad merritt collection

sfisher:
for after migration.

This won't be merged into a branch for production until we update and deploy this with David Loy, but I want to ready to go.

Ryan can you verify that this all looks correct as a second pair of eyes?  Everything should be in the dryad collection when this goes live later.

We can likely remove the collection configuration from the tenants and put it in the app_config sometime in the future, but for now I'm trying to keep any code changes to a minimum until we see things working in one collection in Merritt.


ryscher:
Yes, this all looks good.

PR #344, opened 2021-02-02
Adding better instructions on how to resubmit errored zenodo

sfisher:
These instructions are especially helpful for resubmitting because of temporary problems at zenodo or because things were attempted to be submitted out of order.


ryscher:
Thanks for this update.

PR #345, opened 2021-02-03
Remove outdated solr example

ryscher:
Remove an outdated example from the installation instructions. The referenced sample file has not existed for many years.


ryscher:
Closing, since I rolled this into #347

PR #347, opened 2021-02-04
Update docs for fee waiver, installation

ryscher:
Updates to installation documentation as a result of getting Cassiano started. Also an update to improve findability of the fee waiver information.


sfisher (approved):
Thanks Ryan.

I was searching for the wrong terms!

PR #348, opened 2021-02-04
Prevent nil error from publication updater

ryscher:


sfisher (approved):
Thanks for adding the nil check.

PR #349, opened 2021-02-04
Increase default page size in API

ryscher:
And better document which API calls may result in multiple pages.


sfisher (approved):
looks good.

PR #350, opened 2021-02-04
Adding the directory name method for s3

sfisher:
I think you'll need to use this method, @ryscher, so I'm doing a PR for it and tests for it now.  I also forgot to add the s3 to the example configuration which the tests use.


ryscher:
This looks great. Thanks for adding it!

PR #354, opened 2021-02-10
a few website updates

tracykteal:
Updated advisors page with new bio for Elizabeth, some FAQ info and membership list.


sfisher (approved):
Thanks Tracy.  These changes look great.

PR #358, opened 2021-02-12
Use S3 for Merritt/SWORD submission

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1078

- Removes the ZIP files that were previously sent to Merritt
- Puts all Merritt files *except* the main `manifest.checkm` into S3, and sends them to Merritt
- Adds a utility class for working with S3

Known problem: This does not correctly handle some special characters, including the 4-byte Unicode characters.

To test:
- create a dataset, where some files are stored in S3 (Just upload the files through the UI, since it's all merged in)
- when you finish the submission, you should see the `/manifest` directory being created in S3, under the resource's main directory
- Merritt should correctly process the files
- when the notifier runs, it should delete all files in `/manifest` and `/data`, but the other directories will remain (since they need to be cleaned up by the Zenodo processor).


ryscher:
I have updated this branch to include the UI portions, so complete submissions through Merritt are now possible. One substantial change to the UI code: I combined the S3 class with my similar class. This is now a static class, partially to avoid memory bloat, but more importantly, it was easier to write the rspec tests that way.

sfisher (approved):
Thank you for taking all this on, Ryan.  I know this was a ton of work and not the most fun code to work with.

Thanks for also merging our two different S3 wrapper libraries and making everything work.  That was really nice of you.

I read over all the code and as far as I can tell from reading and testing, everything looks good and works (aside from the things you mentioned).

Also thanks for dealing with the tests, which seemed like they were kind of a pain to fix and modify.

PR #359, opened 2021-02-13
S3 Upload UI for files

sfisher:
These has Javascript code cleaned up by moving it into three partials which should make it cleaner to maintain.  I'm sure there are better JS ways to encapsulate some of this, but I didn't want to modify too much and break things that were working.

I tried to test what I could with requests (pre-signing for EvaporateJS and adding to the db) and the use of the s3 webmocks without getting into the JS things which I think will be hard to test.  I've tested uploads pretty extensively manually.

We can revisit and do more refactoring if you like but this is the best I can do before our meetings on Tuesday and the time constraints.


ryscher:
This looks good. Thanks for getting all of the JavaScript stuff working. Note that I had to make a few minor updates before merging:
- change target branch to `s3-main`
- resolve trivial merge conflicts
- update the directory names in `resources_spec.rb`

PR #360, opened 2021-02-16
Switch IIT to use shibboleth

ryscher:


sfisher (approved):
Thanks Ryan,

🤞  I hope this works for them.

PR #363, opened 2021-02-18
Removes related work for zenodo software if all files removed

sfisher:
This removes the related work for software if all software was removed and adds it again if it's added.

We keep the records staged at zenodo and add and remove files there as they add and remove them in our UI.  Not published at zenodo until we publish at Dryad.

PS.  It's a bit annoying to test with delayed job, but i deployed and tested on dev and it works.

Before with software:

Screen Shot 2021-02-17 at 4 22 32 PM

After with software files removed from zenodo.

Screen Shot 2021-02-17 at 4 24 26 PM

### DIFF
--- a/spec/models/stash_datacite/related_identifier_spec.rb
+++ b/spec/models/stash_datacite/related_identifier_spec.rb
@@ -102,6 +102,48 @@ module StashDatacite
       end
     end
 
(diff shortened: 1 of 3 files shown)


ryscher:
This looks great. Thanks for the extra tests.

PR #364, opened 2021-02-19
Delete S3 files when deleting from user dashboard

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1119

Clean up temporary S3 files when a resource is destroyed, either from the `Delete` button on the user dashboard or through other means.

This *does* include files that were slated to go to Zenodo. It is assumed that if the resource is no longer needed, we do not want to send its files to Zenodo either.


sfisher (approved):
Thanks Ryan.

Very clean and and easy to use.  Mocking tests for changes like this is always fun.  :-/

Looks good.

PR #366, opened 2021-02-22
Migrate files from EFS to S3

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1079

Copy temporary files from the local `uploads` directory into S3. This command may be run multiple times; files will not be transferred if they are already in S3 or if the associated resource no longer needs them (it has already been submitted to Merritt and/or it has not been touched in the last 6 months).

The local files are not deleted. I assume that will be a one-time operation we perform manually after verifying that all files are properly in S3.

To use: `rails local_to_s3:copy_files`

### DIFF
--- a/lib/tasks/local_to_s3.rake
+++ b/lib/tasks/local_to_s3.rake
@@ -1,0 +1,48 @@
+# Local-to-S3 conversion
+# Moves files from local storage to Amazon S3
+# See ticket https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1079
+
+# rubocop:disable Metrics/BlockLength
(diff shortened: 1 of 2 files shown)


sfisher (approved):
This looks good, @ryscher .  I know we need to think about the database files being deleted, also, for the `created` files and it sounds like you're looking into it.  Fine if you do it here or in another PR.

ryscher:
I made some minor updates to this to 
- remove the "created" files for old unsubmitted resources that we're not migrating to S3. 
- improve descriptions in the output so we can easily see if something went wrong

PR #367, opened 2021-02-23
Only override superuser settings in prod

ryscher:
This is a special case for testing API integration with publishers. We have some superuser settings that override default behaviors, such as sending confirmation emails. Dryad's normal behavior resets these settings to their defaults whenever a dataset is edited in the web interface, because we want confirmation emails to be sent when users take actions through our web interface.

When publishers are testing, however, they sometimes do not want any emails to be sent to users. They may want to use realistic email addresses, but avoid confusing real people with emails from the test system. In these non-production circumstances, we want to retain any settings that were made through the API (by the publisher).

So, this update changes the behavior to only reset those special flags when the environment is `production`.


sfisher (approved):
Thanks, Ryan.

Seems like this will be a good change for them to test.

PR #368, opened 2021-02-25
S3 -- file migration only

ryscher:
Merging files from the s3 branches into main. This is *only* the updates that are required to run the migration rake task, so no other functionality is affected.


sfisher (approved):
Thanks for separating all this out, Ryan.

I was a little concerned about the encoded credentials file, so I decrypted the two and compared and the only difference was that secret, so that looks good.  It's makes it hard to know what is going on in that file with changes.

Thanks for doing this.

PR #369, opened 2021-02-25
Only show asterisks to superusers

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1125

When a user is not logged in, or the user is not a superuser, suppress the asterisks for Related Works.


sfisher (approved):
Looks good.

PR #370, opened 2021-02-26
Convert API uploads to use S3

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1111

Allows file uploads through the API to go directly through S3.


sfisher (approved):
Thanks for making these changes, @ryscher .

PR #372, opened 2021-03-02
Improve robustness of EFS->S3 migration

ryscher:
- Handle the case where the `uploads` folder has directories that are not resource IDs.
- Check whether file sizes match before deciding to skip the upload to S3


sfisher (approved):
Thanks, Ryan.

BTW, for what I was wondering about in our meeting today (about files changed between 1st and second migration), I think we might just be able to do a database query for things changed after `first-migration-time` and are created again and have the same name as an earlier upload for the same identifier through an earlier resource.

PR #373, opened 2021-03-02
Don't make files "visible" if they are all deleted

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1113

When a user mistakenly deletes all of the (data) files in a resource, the curators sometimes want to publish it anyway, but they don't want that resource to be listed as a visible version.

Note for the future: The file visibility code is incredibly similar in the `identifier` model and the `curation_activity` model. It would be good to consolidate these whenever we're doing more intensive work in this area.


sfisher (approved):
Thanks for these changes.

PR #374, opened 2021-03-02
Prevent nil errors in pub updater

ryscher:
Cleaning up more potential places that send email alerts.


sfisher (approved):
Always nice to get rid of some errors.

PR #375, opened 2021-03-03
Changes to the zenodo copying/replication models and tests to make it all work with S3

sfisher:
Overall a lot of refactoring here:
- deleted all classes that deal with the file system
- can no longer rely on md5 or similar to tell us what has changed (Amazon doesn't give them with multipart upload) and also we were downloading and comparing first before
- calculate md5 while streaming, so it's a separate object that gets called while streaming (may calculate up to 5 different digest types because that is how we roll in our database)
- Separate class that deals with differences between versions and decides what to upload (file_change_list).  Straightforward for things that only go to zenodo since things get synced with every version, but uglier for dataset replication which just gets synchronized on publications at random times.
- Separate digest class to handle all that fun
- Some small changes to file models when they have different ways of getting source URLs for sending things to Zenodo (From merritt-presigned vs from s3-presigned/live-url)

Let me know if you have questions.  I hope it makes sense and I added a lot of tests.


ryscher:
This is great. I was able to submit software to Zenodo and everything worked as expected. Thanks for putting in all the effort to get so many tests running.

PR #376, opened 2021-03-03
typo fix to discoverability

LoriEP:


ryscher:
Thanks for catching that, Lori.

PR #377, opened 2021-03-03
Require S3 in the resource

ryscher:
Require the S3 library in the resource, to avoid problems like https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1145

I'm not 100% certain this is needed, since it seems to work fine when the other portions of `s3-main` are in place. Guess I need to learn more about how the Rails loading process interacts with the engines.


sfisher (approved):
Thanks, Ryan.

PR #378, opened 2021-03-03
Prevent crazy creation of dataset versions

sfisher:
This update really does a few simple things:

- Stops caching of the editing and review pages so people can't just use their back button to see obsolete information
- Checks that the resource state is "in_progress" for the resource they want to edit and if not kicks them out for things not in that state
- Checks the state before submission to be sure starting from in-progress and not already submitting or submitted.  There were some checks there, but this adds another.
- Disables the "update" button for creating a new version of the dataset and replaces it with an "updating..." text so they can't click it again until it open.  I was able to create 5 resources for the same dataset by clicking this as rapidly and impatiently as I could.


ryscher:
Thanks. I hope this fixes the issues. I wasn't able to break it.

PR #379, opened 2021-03-05
Update UI library processes, create upload widgets, initial work on upload modal

JoelCDL:


sfisher (approved):
Thanks, Joel.

Are main new styles are at stash/stash_engine/ui-library/scss/_upload-widgets.scss ?

I needed to do a big merge of the react-uploads-main back into your feature branch to incorporate all the other changes we've made since you checked out originally.  For some reason, it's still not doing a diff very well, so hard to see just the things that you modified.

The tests look good and I was able to run gulp build (after some problems with xcode on my machine).

Thanks!

Looks good overall and I tested

PR #381, opened 2021-03-05
Remove constant updating of resource.updated at

sfisher:
Pretty much anytime someone a single piece of metadata, the resource.updated_at was getting updated.  We display it in the "my datasets" page, but other than that, I don't think it's very critical.

It could literally be updating every few seconds, anytime users quickly typed a new title, name, description, etc, it updated constantly.  I believe updating it upon entering the editing screen is good enough.  It will also get updated on submission because the actual data in that table changes with Merritt URLs being stored.

Took out the include and callbacks in like 25 different models that also updated resource.updated_at anytime any of them updated.

If we ever really need this to be really accurate we can do a slow query across the 25 different models and take the latest updated at time from all of them.


ryscher:
This makes sense. Hopefully it cuts down on those pesky deadlock errors.

PR #384, opened 2021-03-09
With no files, don't publish to Zenodo

sfisher:
If a dataset gets published (and it's being synchronized to zenodo because it once had software files), then it updates metadata but skips publishing because no current files are present and zenodo doesn't allow publication without files.

I tried to be sure that it does the right thing by adding a version with files and then afterwards adding another version with the files deleted and then publishing.  I also tried adding files in again afterward and publishing it again with a later version.  It all seems to keep file changes up to date as far as I can tell.

If an earlier version was published with files and then files are removed and published again by a curator, then the earlier version is still available at Zenodo because that is how versioning works.

If we need to retroactively delete old versions then that may be a user requirement we need to know and document reasons for.  IMO, the point of public versioning is transparency to force people to "show their work."  I think randomly yanking data or software that is already published and being used by others is a bad practice unless there is an exceptional reason.


ryscher:
Looks good.

PR #385, opened 2021-03-10
Better zenodo error in the database

sfisher:
This does a couple of things that may help in troubleshooting.

Clears out error when a request completes successfully.

Adds a timestamp for each error.

Keeps all errors and appends most recent to the first of the string.

Editing this pr to see if I can get it to refresh tests.


ryscher:
Looks great. This should help us keep track of the issues.

PR #386, opened 2021-03-10
React upload page

cacods:
Includes react-rails; refactors Start New/Edit data set page to be closed to the new page design (remove Upload Data and Upload Software tabs in favor of Upload files tab); the new tab has the dynamics done with React; it's going to be necessary to skip some (or many) tests by now; they will be refactored and/or new ones will be created as of the new backend management (this is to be done).


ryscher:
Thanks for all the work to get this started! I was able to get it running on my machine, though there were a few steps to get webpacker working correctly. We should document those if we're planning to keep webpacker as part of this component.

PR #387, opened 2021-03-10
Create a new list for zenodo emails

sfisher:
Daniella doesn't want to be getting zenodo error info but does want submission to Merritt emails, so we need a new list for this configuration.


ryscher:
Looks good. Thanks for updating this.

PR #388, opened 2021-03-11
aws_db Rails environment

ryscher:
Setting up a new environment `aws_db`, which is the same as `local`, but uses the new AWS database. It currently uses the same credentials as the development DB, so the only change to the credentials file is to add an empty `aws_db` section.


sfisher (approved):
This looks ok to me.  Thank you for doing this Ryan.

PR #390, opened 2021-03-12
Retry on constant errors

sfisher:
This adds retries to the class I use to make almost all requests to Zenodo.

Adds testing of retries.

Allows disabling retries or setting retry number if the caller needs to.  We need to do that for file streaming since retrying only either the input or output stream will have weird results, so the whole streaming operation needs retrying in that case.

I tried moving our workers down to 2 instead of 3, but don't think it helped with anything at zenodo. Also added some extra time between requests.

Or maybe their service is just flakey lately.

I've been seeing rashes of 504 server errors from them every day, especially between 3-6PM PST.  Maybe hackers are going wild.


ryscher:
Looks great. Hope this cuts down on some of the errors.

PR #391, opened 2021-03-12
Allows replication of earlier version, if later version has the problem

sfisher:
The code would find later versions of the same dataset that had errors and not allow replication of an earlier version (if the errors back up and there is more than one).

This fixes that so only considers the older versions and blocks submission when older versions have errors or are not complete.


ryscher:
Great, hope this helps cut down the errors.

PR #392, opened 2021-03-15
Curation Stats Reporting

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1141

Adds capabilities to track curation statistics. These statistics are cached in a new table `stash_engine_curation_statistics`, for easy view/download. A cron job is added to update the statistics daily.

Introduces two rake tasks: `curation_stats:update_recent` and `curation_stats:recalculate_all`. The first time it is deployed on a server, you must run the `recalculate_all` to populate the initial table.


sfisher (approved):
Thanks for doing all this Ryan.  This is a ton of work and code!

I looked over it all, including the 400 lines of code for tests.

It makes me wonder if our states are too complicated, but this all seems to work and thanks for testing which makes it more clear for a lot of different scenarios.

PR #395, opened 2021-03-16
Improve EFS->S3 migration

ryscher:
Make file migration more robust when filesystem and database are out of sync


sfisher (approved):
good.

PR #396, opened 2021-03-16
Prevent Merritt problem with empty string mimetype

ryscher:
If the mimetype gets set to "", treat it as if it were null.


sfisher (approved):
Good fix.

PR #397, opened 2021-03-16
Set a default mime type everywhere

ryscher:
Better handling for file uploads with no known MIME type.


sfisher:
This looks great.  Thanks, Ryan.

PR #399, opened 2021-03-16
Remove spaces from headings in CSV file

ryscher:
For easier processing with external scripts.


sfisher (approved):
Thanks for changing this.

PR #400, opened 2021-03-17
Fixes for minifying ES6 and deployment to stage/production

sfisher:
After trying a lot of different things (because of terrible error messages), this is what fixed the deploy problems.

The uglifier gem needed to be updated to the latest version and needed to use the `harmony: true` flag on initialization.

I hope ES6 works for most browsers.  I think one of the things used is promises https://caniuse.com/promises (show by date relative).    Seems like most browser released since 2014 support them.  Probably major security problems for people running 7 year old browsers, also.

I tried installing a transpiler for older versions and it just made the precompilation time out and get killed instead.  :-/

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -44,7 +44,7 @@ gem 'sass-rails', '~> 5.0'
 gem 'therubyracer', platforms: :ruby
 gem 'turbolinks'
 
-gem 'uglifier', '~> 3.0.4'
+gem 'uglifier', '~> 4.2.0'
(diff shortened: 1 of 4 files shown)


ryscher:
Wow, thanks for tracking this down, Scott! I wish we knew why this was behaving differently on dev and stage, but it's always good to move the gems to current versions.

Note that we may have to deal with this issue again soon, since the uglifier documentation claims it doesn't work with ES6, and suggests switching to ruby-terser.

PR #401, opened 2021-03-17
Improve robustness of stats calculator (and GitHub Actions fix)

ryscher:
Add handling for crazy data that "should never happen". Guess I'll need to start using a copy of the prod database in my testing, because it always breaks even my simplest assumptions, like "all resources are attached to an identifier". Sigh.

A secondary part of this PR is to fix issues with GitHub Actions. We were using the `ubuntu-latest` tag, which caused problems when GitHub upgraded the "latest" version of Ubuntu. It is now set to Ubuntu 18.04. As a side effect, the key had to be changed for the set of cached gems, so we only load an environment that has gems with the correct Ubuntu support files.

### DIFF
--- a/.github/workflows/rspec_features.yml
+++ b/.github/workflows/rspec_features.yml
@@ -5,7 +5,7 @@ on: [pull_request]
 jobs:
   rspec-features:
     name: RSpec Feature Tests
-    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


sfisher (approved):
Thank you for tracking down these crazy caching issues.  I wonder if Brian Riley ran into them also.

PR #402, opened 2021-03-17
Just adding some information about contributing flow

sfisher:
I hope it's helpful to have this info about contributing.


ryscher:
Thanks, it's great to have this specified in more detail.

PR #403, opened 2021-03-18
Changes to former UC collections in Merritt

sfisher:
We did the migration and this branch is currently deployed on production.  It is the same as the last tag but with the collection changes merged in (which I had you look at before, a long time ago).

Just merging it into the main branch so our main configuration will be updated now.


ryscher:
Yes, all looks good. Thanks for taking care of this.

PR #404, opened 2021-03-19
Reorg membership menus

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1176

- pulls in Joel's UI formatting updates that were previously only in the React branches
- adds "Join Us" page, with new formatting in the menu to be subgroup of "Our Membership"
- renames "Our Community" to "Our Membership" (but the old `our_community` URL still resolves)
- adds some initial text on the membership pages, but it's still incomplete (Tracy will work on)


sfisher (approved):
Thanks, Ryan.

This looks good to me.  I switched to the branch and looked at things.  The indent seems fine.

There was a quote on the "join us" change that look kind of squished and double-spaced, but I assume that some of the pages are still being worked on.

Seems OK to me.

PR #405, opened 2021-03-19
Creating files table, models, associations, factorybot factories, rspec files, basic test

sfisher:
I mostly just created new table and classes.  I copied the methods into these files but haven't done testing for all the methods I moved yet which is part of another ticket.

Just trying to do smaller PRs.


ryscher:
Looks good. Thanks for doing this in smaller pieces.

PR #407, opened 2021-03-22
Allow changing from a "known" journal to an "unknown" journal

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1159

In the current system, if a dataset is associated with a journal that Dryad knows about, the ISSN will be present in the submission system. Attempts to change the journal to an "unknown" journal will change the title in the metadata, but not most journal settings, because the ISSN of the old journal will remain.

This PR clears the ISSN whenever the user types in the Journal Name field, so any old ISSN will not remain, and the journal information will be reset correctly.


sfisher (approved):
Thanks for messing with this Javascript, Ryan.

PR #409, opened 2021-03-25
Single table inheritance -- model tests all working

sfisher:
Sorry, this turned into a huge pull request because this is something that was hard to change without all tests failing.  (doesn't have UI/controllers/views really checked yet, but will do as part of later tickets, so a failure there is probably expected.)

- Lots of refactoring of model names and associations
- Fixing of model tests and factories
- Removing old model classes for FileUploads
- Haven't yet fixed and tested the controllers (aside from some renaming of the file classes in them)
- Haven't fixed Zenodo type files aside from some cursory renaming.

More tickets yet to do on this, but this ended up encompassing about half the tickets I have in our current.  :-/


ryscher:
Thanks, Scott. Looks like it was a pain to get some of the tests working again. But this will be worth the effort as we update file behaviors in the future.

PR #410, opened 2021-03-26
React upload url errors

cacods:
- Displays manifest files already created (both with URL errors and the ones without) when GET Upload Files page.
- Removes manifest files both from GUI and the database.


ryscher:
Thanks Cassiano, this looks good. It's nice to see the functionality starting to come together.

PR #411, opened 2021-03-26
Test mimemagic change

ryscher:
This branch is for testing how to get Dryad working again, now that mimemagic has been yanked out from under us. See: https://www.theregister.com/2021/03/25/ruby_rails_code/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Dryad
+# Dryad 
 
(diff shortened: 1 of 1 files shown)


sfisher (approved):
Thank you, Ryan.  This sounds like a mess to deal with.

I installed that library and updated my gems and everything seems functioning normally on my local machine.  Also tested deploying and running on the development machine and seems good there, also.

PR #414, opened 2021-03-29
adding content to our_membership page

tracykteal:
The Our Membership page mainly has a list of members. I checked formatting locally just for the HTML but am not sure if it will render ok on the website. Please let me know if I should make changes. 

This is primarily a formatting check. I'll separately confirm all membership.


sfisher (approved):
This looks good to me.  Thanks, Tracy.  I'm going to merge it in and we can preview and correct any problems that we see.

PR #416, opened 2021-03-29
Fixing deprecation error that just appeared on Wednesday night

sfisher:
Something changed to from a deprecation warning to an error in our tests on Wednesday night with bundler on github actions.  Now all tests fail because of the change.

I'm doing a PR so we can get test working again and accept pull requests again.


sfisher:
I'm going to merge this pull request so I we can use other tests on other pull requests.

PR #419, opened 2021-03-30
React upload url errors 2

cacods:
Doesn't save bad urls to database anymore; just display the errors on screen.


sfisher (approved):
Thanks for doing this, Cassiano.  It functions very well and looks good.

I had a few questions about how React works with the CSS and if the rules needed to be rewritten, but maybe you can give me a tutorial on how it works sometime later.

I also noticed a couple of small UI things but we can clean them up separately sometime later.  (the model dialog doesn't disable the menus and it would be nice to have the text next to the checkbox for urls as a label for it.)

Overall it looks very good and great job hooking it into the validation.

cacods:
> I had a few questions about how React works with the CSS and if the rules needed to be rewritten, but maybe you can give me a tutorial on how it works sometime later.
Yeah Scott. I'm using React CSS modules for the CSS rules, but I'm doing that, as the focus were on developing the React functionalities more than in CSS itself. So I was just copying the rules from Joel or creating my own at the beginning, when we hadn't the CSS yet. I think is the time to try what you suggested before, i.e., using the classes from Joel's work directly in the code. I'm going to see how that works. Let's discuss more about the CSS part inside React on our next meeting tomorrow if you can.

> the model dialog doesn't disable the menus
This is yet to be fixed, yes.

Thanks to you and Ryan for helping me to clarify the URL validation process. And if you have suggestions about the controller changes that I made please let me know.

PR #422, opened 2021-04-01
This fixes the UI to use the new Single Table Inheritance models

sfisher:
Mostly changes the following things in the UI/controllers and tests for them.

- file_uploads --> data_files
- software_uploads --> software_files
- Similar for class names (SoftwareUploads, etc)
- Fixes tests
- Updates URL generators to follow same format
- Corrects zenodo replication classes for new models, but doesn't yet add supplemental information.

This to manually test:

- [ ] We can still access old files in Merritt that worked before (less than 2021-03-01 or whatever we migrated)
- [ ] New files (urls and files for both software and data) get correctly entered in the new table
- [ ] New files get correctly uploaded to Merritt
- [ ] New things are still being sent to zenodo correctly and replication being logged in the Zenodo copies table


ryscher:
This looks good, and my manual testing all worked. Thanks for changing all of the variable names. I'm sure it was tedious, but it will help to avoid future problems.

PR #423, opened 2021-04-02
React upload UI library update

cacods:
This incorporates UI Library directly in React code.


app/javascript/containers/UploadFiles.js
```diff
@@ -12,13 +12,16 @@ class UploadFiles extends React.Component {
     state = {
         upload_type: [
             {
-                id: 'data', name: 'Data', description: 'eg., Spreadsheets, example1, example2',
-                buttonFiles: 'Choose Files', buttonURLs: 'Enter URLs'},
+                id: 'data', logo: '../../../images/logo_dryad.svg', alt: 'Dryad', name: 'Data',
```
sfisher, line 15: This seems ok for me for now as long as it is working.

We'll have to get our deployment working for pre-compiled assets for stage and production and be sure things work.  Rails often uses some things called "image_tag" to serve images with the asset pipeline, but it probably wouldn't work within a React file.

We'll see what happens later and may need to do something else if it doesn't show correctly.

sfisher (approved):
Looks good to me.  Thanks, Cassiano.

app/javascript/containers/UploadFiles.js
cacods: Yes, I see. I put the relative paths in there as it's working locally but you pointed very well. Let's see what the best way to refer to the images from within React. Thanks Scott.

PR #425, opened 2021-04-05
Inherits from GenericFilesController rather than includes methods from concern

sfisher:
This might be helpful to Cassiano.  I never use a GenericFile directly as a controller since the sections are specific to each kind of file and they just included the common methods.

This inheritance has the same effect for the controllers, but it may be that Cassiano might not need the specific controllers for each file type and just use the GenericFiles controller but with params for models and using specific models instead.  (or something like that).

This still retains compatibility with the way things are in our current interface.

Also removed a couple of unused methods in there that were about uploading files directly to the filesystem which we no longer use.


ryscher:
Thanks for this reorg. It does seem like it will be more useful this way.

PR #426, opened 2021-04-06
Software-like replication to Zenodo can operate on multiple types

sfisher:
Essentially this isn't a huge change in function, but with changes that queries or activerecord associations dynamically choose the type that is being worked with rather than it always being software (for software-style replication to zenodo).

It also works with the new values in the zenodo_copies table for supplemental.

Had to work through the tests and re-get things working.

Also changed a query that was hard to follow and read to something more understandable.

Also includes code from the previous PR for the controllers, though it doesn't interact much with this code.

I'll add another pull request for monkeying with the automatic associations which we will need to be different for supplemental.


ryscher:
Looks good, works fine in my testing.

PR #428, opened 2021-04-07
Add delayed_job for supplemental info

sfisher:
This is treated the same way as software for delayed_job and so certain events can trigger supplemental (submitting, publishing) just like software.

Added tests for it like the tests software has.

This makes it ready to work once supplemental files show up and are uploaded to work on.

We don't have a really easy way to add and test manually yet until we dont have the UI to add supplemental files.


ryscher:
Thanks for adding these features.

PR #429, opened 2021-04-08
React upload sti

cacods:
React working with new Single Table Inheritance (STI) for data files and software files. The UI tests for uploading software type manifest files were released again.


sfisher:
Looks good to me, Cassiano.

PR #430, opened 2021-04-08
Verify any files that are supposed to be present on S3 are available before submission

sfisher:
We had 3 cases over the past few weeks or a month where files uploaded and written to our DB were missing from S3 after submission.  I suspect this may be an upload or other issue from Evaporate.js or Amazon, even though Evaporate.js triggered the callback that marks a file complete and writes it into our database.

It could also be that the assembly of one file from many parts at S3 failed.  Two of the files were large ones ( > 10GB ) and one was some random small file.  Non-present files make Merritt submissions fail and then we have to tell users "Sorry, upload again or use the FTP server."  Lots of variables here with the user's browser and connection, so IDK what goes wrong and can't reproduce it.  I don't believe it's because files are being deleted that shouldn't be since I looked over all that code carefully.

Anyway. This adds a validation that all files that should be on S3 actually exist on the review page before allowing submission.  It tells them to re-upload a file (if the problem persists past a few minutes).

You can manually test the error by uploading some files to S3 and then going into the S3 UI, finding a file or two that you just uploaded and either rename or delete in S3.  Then go to the review page and it will show you the error.

PS.  I originally tried doing the validation immediately after each upload completion, but that triggered a race condition where the file didn't exist on S3 yet since it takes some amount of time for the parts to assemble into a full file, even though Evaporate notifies it as complete.  The race condition only notifies on larger files since small files assemble almost instantly.  I'm hoping delaying this error to the review page will be sufficient time.  I'll try with a 10GB+ file on the dev server.

### DIFF
--- a/spec/models/stash_datacite/completions_spec.rb
+++ b/spec/models/stash_datacite/completions_spec.rb
@@ -533,6 +533,7 @@ module StashDatacite
 
         before(:each) do
           expect(completions.required_completed).to eq(REQUIRED_COUNT) # just to be sure
+          @resource.generic_files.each { |f| f.update(url: 'http://example.com') }
         end
 
         it 'warns on missing title' do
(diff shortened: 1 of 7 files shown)


ryscher:
This is a great solution. Hopefully it will prevent most of the upload problems.

PR #431, opened 2021-04-08
Adding webpacker info for local_dev

sfisher:
I use this one for development sometimes.  IDK if it's actually better than dev, but I think there are a couple of differences.


ryscher:
This looks fine. Thanks for making the update.

cacods:
> I use this one for development sometimes. IDK if it's actually better than dev, but I think there are a couple of differences.

It makes sense. Thanks to sharing with us.

PR #432, opened 2021-04-09
Utilities for connecting with the journal email account

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/703

Adds a utility class to facilitate processing emails from Dryad's partner journals.

Before working with the journal email account, Rails must obtain a token that allows access to that account. The token can be obtained by running `rails journal_email:validate_gmail_connection` and following the listed directions. (Note that this requires access to the GMail account for journal-submit-app@datadryad.org)

There is only one journal email account, which receives both test messages and real messages from journals. It is used by the production email processing system (currently still on the v1 server), so it is best to avoid deleting any messages from this account. Testing can be performed by adding and removing the email label that triggers processing. It is `dev-journal-submit` for all non-production servers.

To test: 
1. Create/verify the token with `rails journal_email:validate_gmail_connection`
2. Ensure at least one message in the account has the label `dev-journal-submit`
3. In a Rails console session:
```
require 'stash/google/journal_gmail'
m=Stash::Google::JournalGMail.messages_to_process.first
mc=Stash::Google::JournalGMail.message_contents(message: m)
```


sfisher (approved):
Wow, this looks like a hassle to get working.

If you want to put the token.yaml into the "shared" directory then stuff there stays the same for every deploy on our servers.  IDK if there is really an advantage over what you did, though, which is to put it above the "current" root directory (which is the releases directory on the servers).

I honestly don't remember how all the shared directories get linked in to the app and also if it is in our app code directory on a development machine then we need to be sure to .gitignore it so we don't accidentally commit it, where if it is above the root then shouldn't be a problem.

Seems fine to me.

PR #433, opened 2021-04-09
Adding experimental Counter CoP tracking on production

sfisher:
This just allow Counter CoP stats tracking from the service DataCite is developing.


ryscher:
Looks good.

PR #436, opened 2021-04-15
Deletions for something that doesn't exist don't error

sfisher:
For troubleshooting and manual info, see https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1186 .

Also re-jiggered some requires because depending what else was already required it would sometimes fail.


ryscher:
Thanks for this extra error catching.

PR #437, opened 2021-04-16
Quick one-off task for checking DataONE collection to see what items download

sfisher:
This just does a query to get published DataONE items and checks one file in each dataset to see if it downloads and outputs to a CSV.

Useful for getting list for Merritt.  Because this is a quick one-off for Merritt I haven't added tests.  It mostly just reuses our already existing method for creating presigned URLs and then save the results to output as CSV.


ryscher:
Thanks, Scott. This will be useful to check on Merritt in the future. Sorry you had to deal with the GitHub Actions problem.

PR #438, opened 2021-04-16
React uploads local files

cacods:
I think the most of uploading files from local file system to S3 is done for the new version of a resource.

I've created more feature and response tests and tried to organize them a little.

Please ignore the mess with removing files by now. This is my next ticket.

Fill free to make suggestions or propose improvements.


sfisher:
Sorry, this started breaking at Github today, @cacods .  It seems like something got deprecated.

To get the tests to run right you can change `action/setup-ruby@v1` to `ruby/setup-ruby@v1.68.0` .  These are the files in `dryad-app/.github/workflows` that have that: rspec_features.yml, rspec_internal.yml, rspec_responses.yml and rubocop.yml.

app/javascript/containers/UploadFiles.js
```diff
@@ -57,23 +64,116 @@ class UploadFiles extends React.Component {
...
     }
 
+    uploadFilesHandler = () => {
+        const config = {
+            aws_key: this.props.app_config_s3.table.key,
+            bucket: this.props.app_config_s3.table.bucket,
+            awsRegion: this.props.app_config_s3.table.region,
+            // Assign any first signerUrl, but it changes for each upload file type
+            // when call evaporate object add method bellow
+            signerUrl: `/stash/data_file/presign_upload/${this.props.resource_id}`,
```
sfisher, line 90: I wonder if this URL can be passed in somehow to the javascript when it's created.  Maybe it's a pass it in.  It might be something like "<%= data_file_presign_url_path(resource_id: @resource.id) %> or something like that (depends how the rails routes are set up).

app/javascript/containers/UploadFiles.js
```diff
@@ -57,23 +64,116 @@ class UploadFiles extends React.Component {
...
+                },
+                error: function (msg) {
+                    console.log(msg);
+                },
+                complete: (_xhr, awsKey) => {
+                    const csrf_token = document.querySelector('[name=csrf-token]');
+                    if (csrf_token)  // there isn't csrf token when running Capybara tests
+                        axios.defaults.headers.common['X-CSRF-TOKEN'] = csrf_token.content;
+                    axios.post(
+                        `/stash/${file.uploadType}_file/upload_complete/${this.props.resource_id}`,
```
sfisher, line 123: I also wonder if the generic_file (controller) could handle all of these rather than having to know the different controller names.

In rails it seems like if you save the `type` field correctly for the GenericFile model then it may be automatically converted to the correct type of file and you wouldn't have to use 3 different urls?

app/javascript/containers/UploadFiles.js
```diff
@@ -249,8 +351,7 @@ class UploadFiles extends React.Component {
         if (csrf_token)  // there isn't csrf token when running Capybara tests
             axios.defaults.headers.common['X-CSRF-TOKEN'] = csrf_token.content;
 
-        const typeFilePartialRoute = file.type_ + '_files';
-        axios.patch(`/stash/${typeFilePartialRoute}/${file.id}/destroy_error`)
+        axios.patch(`/stash/${file.uploadType}_files/${file.id}/destroy_error`)
```
sfisher, line 354: Similar to the other places, IDK if the generic_files_controller could handle all of the destroys instead of having to know the type of the file.

stash/stash_engine/config/routes.rb
```diff
@@ -25,7 +25,7 @@
   post 'curation_note/:id', to: 'curation_activity#curation_note', as: 'curation_note'
   post 'curation_activity_change/:id', to: 'admin_datasets#curation_activity_change', as: 'curation_activity_change'
   resources :tenants, only: %i[index show]
-  resources :data_files, :software_files do
+  resources :data_files, :software_files, :supp_files do
```
sfisher, line 28: (related to the other things I mentioned).  I wonder if it's possible to get rid of the controllers (*not the models*) for the data, software and supplemental and do it all with just the `generic_files` by setting the type for the file when it is saved.

IDK, maybe that is a bad idea.  You'll know more than I do because you'll know if any of the three types of files need anything different in the controllers.  If they don't need anything different beside having the `type` set then maybe we can only use the one controller and get rid of the other three.

sfisher (commented):
Thank you for all the tons of work, Cassiano.  It's great that it is working so well.  I think my main questions were just about URLs and if we can simply use a GenericFilesController and get rid of the other three.  (but still keep the different models)

Also, thanks for working through all that tests.  It looks like it was a lot of work.  I hope it wasn't too much of a pain.

sfisher:
This looks pretty good to me and we have other tickets to check into a few things, so I'll merge it in.

app/javascript/containers/UploadFiles.js
cacods: Very sorry for theses absurd later responses Scott. I planned to answer them but forgot about.

I didn't understand very well. You mean passing this URL's as props to the UploadFiles React component? As we do with the other props, like:

`<%= react_component("UploadFiles", {
  resource_id: @resource.id,
  file_uploads: @resource.generic_files.validated_table.map(&:attributes),
  app_config_s3: APP_CONFIG[:s3],
  s3_dir_name: @resource.s3_dir_name(type: 'base')
}) %>`

in `upload.html.erb`?

About doing something like `<%= data_file_presign_url_path(resource_id: @resource.id) %>` directly inside React components IDK if it's possible for that React doesn't know about embedded ruby I guess.

app/javascript/containers/UploadFiles.js
cacods: Yes, I think it's possible. And maybe this is straightforward to do. Currently we have the three routes for the three different upload types:

```
                  data_file_complete POST         /data_file/upload_complete/:resource_id(.:format)       stash_engine/data_files#upload_complete
               software_file_complete POST         /software_file/upload_complete/:resource_id(.:format)   stash_engine/software_files#upload_complete
                   supp_file_complete POST         /supp_file/upload_complete/:resource_id(.:format)       stash_engine/supp_files#upload_complete
```
Maybe I could create only one route for `generic_file` and with the type defined as a string like 'data', 'software', 'supp' at the moment of the ajax request.

app/javascript/containers/UploadFiles.js
cacods: Just like my previous answer. In this case I think it's easier since it's a patch request.

stash/stash_engine/config/routes.rb
cacods: I agree with you. I don't need anything more than we have now for the three controllers other than the type definition. As I told in our last meeting there's that way of defining different resources for the three types and direct the requests for the same controller (https://devblast.com/b/single-table-inheritance-with-rails-4-part-2):

```
# config/routes.rb
resources :generic_files
resources :data_files, controller: 'generic_files', type: 'data' 
resources :software, controller: 'generic_files', type: 'software' 
resources :supp, controller: 'generic_files', type: 'supp'
```
or something like that.
But IDK if it's the best way to do that. Maybe it's just a matter of having only `:generic_files` and defining always the type when doing the requests. And for that, when the requests are `get` requests, maybe including a route parameter for the type.

cacods:
> Thank you for all the tons of work, Cassiano. It's great that it is working so well. I think my main questions were just about URLs and if we can simply use a GenericFilesController and get rid of the other three. (but still keep the different models)
> 
> Also, thanks for working through all that tests. It looks like it was a lot of work. I hope it wasn't too much of a pain.

I'm used to try to cover all the tests, but sometimes this can make me slow my job. So, if you fill I'm getting slower, please let me know and I will skip some unessential tests.

PR #439, opened 2021-04-19
File checker for Merritt

sfisher:
One off task. Checks all files from Merritt in public datasets.  I suppose if this becomes more useful we could expand this to be more regular occurrence in the future.

This ran against the production database over the weekend and it created a CSV file with the results which I'll look at shortly.  It seemed like most errors were either encoding errors and inability to get presigned download from Merritt or else the Content-Length reported size was different than what we show in our database.

I looked at some of the content-length mismatches and it seemed like everything I looked at downloaded and appeared to be correct (for file types I could look at easily like excel).


ryscher:
Thanks for putting all of these details in a report. They will be useful in discussions with Merritt, and it will be good to have this for future use.

PR #440, opened 2021-04-20
Don't upload zero length to zenodo

sfisher:
If a file has nothing in the body and is zero length then Zenodo will reject the entire submission, so in that case leave the file out.

Also added a test and a fix for an encoding mangling issue with our http client in some circumstances.


ryscher:
Looks good. Hope it corrects the issues.

PR #441, opened 2021-04-21
React uploads remove files

cacods:
Removes files that have been _uploaded_ (but not submitted) from AWS S3 bucket.


sfisher (approved):
Thanks, Cassiano!

Works great and the code and tests look good.

PR #442, opened 2021-04-22
Instructions on how to install Node on servers

sfisher:
I thought we might need more changes to our configuration for deployment but it looks like it works automatically with Capistrano so long as the node dependencies on the server are met.

At least deploy worked for me on development. We'll see if other changes are required for stage (and production) later.


ryscher:
Thanks, this looks good. One minor note -- on my server I needed `sudo` to install yarn, though I'm guessing permissions may be different on other servers.

PR #443, opened 2021-04-23
Journal email -- message parsing

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1190 and https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1189

This PR adds a `Manuscript` model to Dryad, and parses messages from GMail to populate the Manuscript objects. It does not yet do anything with the resultant Manuscript objects.

To test:
- Run the database migration
- Update the Journals table to include journal codes: Copy the file journal_codes.csv from the /tmp directory on any Dryad server into the directory where you run rake tasks, and run `rails dryad_migration:import_journal_codes` (I decided to deprecate the old journal import process, since the journal metadata in the v1 server is no longer correct, and we do not want to overwrite the entirety of the journal table)
- Ensure that a GMail token is present, by running `rails journal_email:validate_gmail_connection` (as described in https://github.com/CDL-Dryad/dryad-app/pull/432)
- Run the parsing process with `rails journal_email:process`. This process will only read messages in journal-submit-app@datadryad.org that are flagged with the `dev-journal-submit` label.
- Inspect the resultant `Manuscript` objects to ensure they contain the correct metadata.

### DIFF
--- a/config/app_config.yml
+++ b/config/app_config.yml
@@ -265,8 +265,8 @@ production:
     gmail_client_id: <%= Rails.application.credentials[Rails.env.to_sym][:gmail_client_id] %>
     gmail_client_secret: <%= Rails.application.credentials[Rails.env.to_sym][:gmail_client_secret] %>
     journal_account_name: journal-submit-app@datadryad.org
-    journal_processing_label: journal-submit
-    journal_error_label: journal-submit-error
+    journal_processing_label: journal-submit-v2
+    journal_error_label: journal-submit-error-v2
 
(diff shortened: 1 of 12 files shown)


sfisher:
Hi @ryscher , I'm trying to run through your instructions for testing and can't find the file `journal_codes.csv` on the servers.

sfisher (approved):
Thanks for all this work, Ryan.

Parsing these things doesn't look like lots of fun.  It seems like it works.

PR #446, opened 2021-04-23
Modifications to DataONE tenant info

sfisher:
We met with the Merritt team yesterday and it turns out the DataONE merritt collections are already using the Dryad S3 bucket on their side for storage.  However it is still a separate collection in Merritt.  They were understandably reluctant to move those datasets into the Merritt Dryad collection since that involves a bunch of work for no real difference in storage underneath.

We talked over the reasons for the migration of the UC collections as twofold 1) moving storage to be under Dryad for billing and 2) to simplify management  of all the separate Merritt accounts for different tenants into one account.

Eric was able to give the standard Dryad user access to DataONE content so we can still use the same user credentials, though the collection (endpoint) is still different than all the other tenants which is annoying but probably workable.

### DIFF
--- a/config/tenants/dataone.yml
+++ b/config/tenants/dataone.yml
@@ -49,7 +49,7 @@ production:
   manager_email: ["Daniella.Lowenberg@ucop.edu"]
   repository:
     domain: https://merritt.cdlib.org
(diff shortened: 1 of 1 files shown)


ryscher:
Looks good.

PR #447, opened 2021-04-26
React uploads remove files

cacods:
- Considers files from previous versions when removing them, displaying in Upload Files tab, and validating submitted urls;
- Does not add file to files table when the file is already there for the same upload type.


sfisher (approved):
I tested this out and it works great and sets previous files to deleted for deletion from the repo. Looks good and seems to work well.

Maybe you can walk us through the React code a bit sometime since I'd be curious to learn more in depth about it.

PR #448, opened 2021-04-26
added consortia member information to join_us page

tracykteal:


ryscher:
This all looks good, and the page renders properly. Thanks!

PR #450, opened 2021-04-27
Adds rework of relationships for adding metadata to zenodo and dataset

sfisher:
This adds the "added_by" field to related identifiers and it sets it to "zenodo" if something was automatically added by zenodo.

This then enables us to not-send those as part of metadata to zenodo since the relationship will be different coming from the other side.  It also makes it easy to just destroy them and re-set them up for things like only showing the latest version at zenodo (since each version has another DOi there).


ryscher:
Ok, looks good. Thanks for making sense of this complicated workflow.

PR #451, opened 2021-04-27
Journal email -- submission system

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1188

This builds on https://github.com/CDL-Dryad/dryad-app/pull/443, so you can either review that first, or just this one for everything together.

This PR connects the Manuscript model to the Dryad submission system, and removes reliance on the v1 journal module. 

To test:
1. Ensure the GMail connection is authorized and journal codes are imported, as described in https://github.com/CDL-Dryad/dryad-app/pull/443
2. Parse at least one email message, as described in https://github.com/CDL-Dryad/dryad-app/pull/443
3. Create a new Dryad dataset in the submission system.
4. Enter the journal name and manuscript number associated with the email you parsed, and click "Import Manuscript Metadata"
5. All of the details from the email should get imported into the dataset.

### DIFF
--- a/config/app_config.yml
+++ b/config/app_config.yml
@@ -265,8 +265,8 @@ production:
     gmail_client_id: <%= Rails.application.credentials[Rails.env.to_sym][:gmail_client_id] %>
     gmail_client_secret: <%= Rails.application.credentials[Rails.env.to_sym][:gmail_client_secret] %>
     journal_account_name: journal-submit-app@datadryad.org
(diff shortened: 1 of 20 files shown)


sfisher (approved):
Thanks for this, Ryan.

I tested and found it was importing and working great.

It's great that old stuff over two years old gets cleaned up daily.

PR #452, opened 2021-04-27
Remove v1 journal service from the status dashboard

ryscher:
I have already removed the v1 journal service from the dashboard's database entries, so this should have no visible effect for users. 
- removes unused code
- adds a bit of documentation about the status dashboard
- corrects capitalization for the DB Backup class


ryscher:
@sfisher there is no need to rush reviewing this. It can wait for next sprint; I just wanted to finish it before I forgot.

sfisher (approved):
Thanks for doing this, @ryscher .

PR #453, opened 2021-04-28
Document more details about the v1 server

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1235

Documenting the current state of the server as we prepare to shut it down.


sfisher (approved):
Thanks for putting this in just in case.

PR #454, opened 2021-04-28
added members to _our_membership

tracykteal:
Added new member University of Wisconsin, added some of The Claremont Colleges that weren't on the list and expanded UCLA to be consistent with other UCs. Moved Texas Tech so lists in each column were more even.


tracykteal:
Amended to revert back to UCLA and hold off on Wisconsin

sfisher (approved):
Thanks for these updates, Tracy.

PR #455, opened 2021-04-29
One off script for duplicating dataset version into a new dataset

sfisher:
This is a little crazy and a one-off for some datasets that got deposited and the EZID DOIs were invalid because of some account thing.  So now they need to get new DOis and the DOIs need to to be for UC Santa Cruz.

Originally, we thought we could just do some local_id fun in Merritt to fix the DOIs if we made new ones, but apparently it's hard or impossible.

This script is kind of hack-ish.  What it does is 1) create a new resource for a dataset with Amoeba dup; 2) reset some info on it so it's no longer associated with the old identifier; 3) generates a new identifier under the preferred tenant; 4) re-associates the resource with the new identifier instead and resets versions back to version 1; 5) generates presigned URLs from Merritt for every Merritt file present in the old dataset; 6) removes any deleted files.

After all this is done the item is an "in progress" dataset that someone can go look over and submit from the user interface before the merritt presigned URLs expire since they're used to ingest it again into Merritt for the new dataset.

I didn't really add tests since this may not ever be used again and people will look over the new datasets it creates before manually submitting them.

### DIFF
--- a/lib/tasks/dev_ops.rake
+++ b/lib/tasks/dev_ops.rake
@@ -172,6 +172,63 @@ namespace :dev_ops do
     puts 'Done'
   end
 
+  desc 'Takes a DOI, user_id (number), tenant_id and copies the latest submitted version into a new dataset for manual submission'
+  task version_into_new_dataset: :environment do
+    # apparently I have to do this, at least in some cases because arguments to rake are ugly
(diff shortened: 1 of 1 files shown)


ryscher:
Thanks, Scott. This will probably be useful in the future.

PR #456, opened 2021-04-29
Correct issues with dryad_manuscript_spec

ryscher:
Add the `spec/lib` directory into the GitHub Action tests, and correct issues with the manuscript tests.

Most of the manuscript testing has moved into `spec/models/stash_engine/manuscript_spec.rb` and `spec/services/stash_engine/email_parser_spec.rb`. Those tests already exercise the functionality of `DryadManuscript`. I have adapted a couple of the old tests that exercise `DryadManuscript` directly, but I removed the ones that duplicate the other parsing tests.


sfisher (approved):
Thank you for fixing these up, @ryscher .

PR #457, opened 2021-04-30
React uploads items to finish

cacods:
- Recovers sanitizing file names feature from previous Upload Files tabs.
- Updates texts in Upload Files tab.
- Validates URL manifest files without the checkbox to confirm.


sfisher (approved):
Overall looks pretty good, but I wasn't sure if importing that existing library was needed.

PR #459, opened 2021-04-30
Scripting to delete datasets

sfisher:
I had to delete 19 at once so didn't want to have to do all the steps by hand for all 19.

Example to run:

```
RAILS_ENV=development rails dev_ops:destroy_dataset "10.5072/FK25T3QP13"
```

It will 1) Remove from SOLR, 2) Tombstone in EZID or remind you to do something in DataCite UI--we can build this out later if we want to automate, 3) List all the Zenodo dependencies you should ask them to remove, 4) destroy from our database.

I'll add a note in confluence.

### DIFF
--- a/lib/tasks/dev_ops.rake
+++ b/lib/tasks/dev_ops.rake
@@ -172,6 +172,63 @@ namespace :dev_ops do
     puts 'Done'
   end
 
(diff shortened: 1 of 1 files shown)


ryscher:
Thanks, this will definitely be useful!

PR #460, opened 2021-05-04
Add supplemental information to license, review and private downloads for peer review

sfisher:
The changes do three things.

1. Add cc-zero as the type for supplemental information.
2. Add sections to the review page for supplemental information for display if they exist.
3. Add section to the private for peer review "sharing page" for supplemental if it exists in zenodo.

These mostly aren't huge changes, but duplicating the existing software sections with a different parameter and some of the same rendering.

I'm merging into react-uploads-main since there really isn't a UI for supplemental information except in that branch.


dlowenberg:
Realized after Scott began working on this that for accessibility reasons we should let users know what the SI file license is going to be -- and also we should do CCBY 4.0 instead of CC0 for the SI files in case they are images.

Requested changes:
- Default send any SI files as CCBY 4.0 instead of CC0
- In the Agree to terms section: 1) change the Zenodo one to say "License and Terms of Service for Zenodo" instead of software 2) In the sentence before "For guidance, see choose a license" add sentence "Supplementary information are published with a CCBY 4.0 license".

ryscher:
This looks good, and works in my testing so far. I'll wait for the merge until Daniella's changes are addressed.

PR #461, opened 2021-05-04
Supplemental license PR 2 -- CC-BY-4.0 and text changes

sfisher:
Builds on the previous pull request.

Adds the license as CC-BY instead of CC-zero.  Also displays slightly different text depending on what files are present.

Any zenodo files -- displays a few general intro things about zenodo.

software files -- shows the choose a license and info about sfw

supplemental -- Shows info that it's being licensed under CC-BY


stash/stash_datacite/app/views/stash_datacite/licenses/_software_license.html.erb
```diff
@@ -1,8 +1,5 @@
 <% # moving this here since I suspect this will need to be used again later %>
-<div>
-  <h3>License and Terms of Service for Software</h3>
-  <p>Software and other files uploaded will be managed and preserved at Zenodo. Your related DOI will be linked to
-    your dataset. Please select a license from the below dropdown list for your Zenodo deposit. For guidance,
+    <p>Please select a license from the below dropdown list for your Zenodo deposit. For guidance,
```
ryscher, line 2: Shouldn't this say "for your Zenodo software" ? Just below, it says that supplemental files are always CC-BY, so this  license should not apply to them.

ryscher:
This looks good, and all very straightforward, except for the comment above.

stash/stash_datacite/app/views/stash_datacite/licenses/_software_license.html.erb
sfisher: Yes, this is a good catch.  I'll changing the wording now.

PR #462, opened 2021-05-04
Add provenance info for datasets created via API

ryscher:
Adds some extra CurationActivities for datasets that are created through the API. This will help in tracking and debugging as journals start using the API more.


sfisher (approved):
Thank you, Ryan.

This will be very useful to know as we see a lot more use from the API coming soon.

PR #463, opened 2021-05-04
Adds lockout for unfinished uploads

sfisher:
This locks the navigation buttons and displays a warning if an item is pending or in the process of uploading (showing a progress bar) still.

I also cleaned up a bunch of the libraries from loading with Browserify and Sprockets (the old way) since we really don't need to load them twice and that way is no longer used.


ryscher:
Thanks, this works great! I'm glad you were able to find a way to detect the completion of the upload without having to do something complicated.

PR #464, opened 2021-05-05
Bundle updates for security issues

sfisher:
There were a number of security notifications, so doing `bundle update` on our app and engines.

I'm hoping this fixes many of the issues we are being notified about.

Also removing rspec out of some of the gemspecs since in the gemspec and the gemfile and it gives warnings for being in both places with bundler.


ryscher:
Looks good. I did a basic runthrough of the main functionality, and nothing seems to be broken.

PR #465, opened 2021-05-05
Fixes spurious size errors for some especially evil CSV files

sfisher:
We were getting file size mismatches for some CSV files on github when trying to stream to zenodo.

- Gives better error for size mismatches by stating the size of both in the error
- Changed to use `bytesize` in the additions of the chunk size because length gives different values in some very specific circumstances that only seem to come up in some very rare circumstances.

I think the .length treats the chunk as a string so may have more bytes per character for some kinds of text files.  Either that or it's doing some other kind of weird calculation.  Using bytesize instead seems to make it accurate.

The data stream was fine but just the size calculations were off for two bytes for `https://raw.githubusercontent.com/JayRGopal/GlaucomaAI/main/GlaucomaCSV.csv`


ryscher:
Wow, interesting problem. Hope this fixes it.

PR #467, opened 2021-05-06
Adding information about zenodo supplemental doi to the email

sfisher:
Adding to the email and adding to the test to determine it shows them when they are present.


ryscher:
This is good. Thanks for adding a thorough test.

PR #468, opened 2021-05-07
Fix callback logic and associated tests

ryscher:
Fixes the issue noted in https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1259, which was appearing intermittently, depending on the order for running some of the tests.

The root problem was that some CurationActivity callbacks were being called at times they should not, because there were several methods listed in one callback line, and the `if` statement was only applying to the last one. I moved these methods into a single block so the `if` applies to all of them. Then, I discovered that the tests for the callbacks were all useless, because they were testing for the absence of a method call that didn't exist. They are actually testing the real functionality now.

### DIFF
--- a/spec/models/stash_api/dataset_parser_spec.rb
+++ b/spec/models/stash_api/dataset_parser_spec.rb
@@ -317,7 +317,6 @@ module StashApi
         new_resource.current_editor_id = @user.id
         new_resource.save!
         @resource = new_resource
-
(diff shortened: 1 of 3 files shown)


sfisher (approved):
Thanks Ryan!

Wow, that's kind of crazy, so the if only applied to the last item?

Does that mean it called process payment constantly?

I suppose the first two might not be as big a deal, at least for published things if they got updated again.

In any case, this is probably more clear.

PR #469, opened 2021-05-07
Get rid of the remaining React CSS modules -- use only UI classes.

cacods:
I used some few react inline styles following the inline styles from UI. Let's see next if it works on Stage.


sfisher:
Thank you cassiano.  Looks good to me.  HAve a good weekend.

PR #470, opened 2021-05-07
Adds another call to update related identifiers earlier in the process

sfisher:
We should be able to call this update method as often as we want.  It only re-updates the zenodo-added relationships so doesn't affect other things.

The case for this is this sequence of events:

1. Dataset V1 updated at 2021-05-06 17:57:03
2. Dataset V2 created at 2021-05-06 17:57:36 (still no Zenodo DOIs to duplicate into new version since not finished submitting yet)
3. Submission finishes at Zenodo for Dataset V1 at 2021-05-06 18:00:01 and relationships are added to V1.
4. Dataset V2 Submitted to Zenodo again, but it exits early because no file or other changes needed aside from publishing.

This adds updates to our relations before an early exit so that these items will always get the correct relationships added, even if other things about the dataset don't need updates.


ryscher:
Looks good.

PR #476, opened 2021-05-12
Loosen the rubocop restricitons

ryscher:
Loosening the restrictions that were most frequently disabled. This has the side effect of removing a large number of `disable` statements.

I tried to maintain a reasonable level of strictness, trying to keep the code well-structured while reducing the level of annoyance. I raised the limits that we constantly have to disable, but for every limit I raised, there are still some instances where our code is over the limit and we have some corresponding `disable` statements for it.


sfisher (approved):
Thanks for going through this, @ryscher .

This is a good change!

PR #477, opened 2021-05-14
Fixing citation processing and adding tests.

sfisher:
It fixes the json parsing that was preventing completion and adds dashboard check and test of check.


ryscher:
This looks fine, though I have to admit I have no idea what "&!" does....

PR #478, opened 2021-05-17
Warn about submission unless they give us data files

sfisher:
Data files are why Dryad exists and we shouldn't be a conduit for people to use Zenodo with no data.


sfisher:
Hold off on this one.  I'll convert it to a popup warning instead of a completion requirement for submission.

ryscher:
This looks good and works well. The warning box looks great.

PR #479, opened 2021-05-17
Publisher/society dashboard permissions

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1231

Allows administrators for publishers and societies to view datasets associated with their journals. They can not edit the datasets, though they can change the dataset to be associated with a different journal.

The easiest way to test is to set up relationships in the console, something like this:
```
o = StashEngine::JournalOrganization.create(name: 'Some org')
j = StashEngine::Journal.last
j.sponsor = o
j.save
u = StashEngine::User.find(<ID>)
StashEngine::JournalRole.create(journal_organization: o, user: u, role: 'org_admin')

# verify the user has admin privs for the correct journals
u.journals_as_admin
```

Once the relationships are set up, you can login as the appropriate user and see the associated datasets in the Admin dashboard.


sfisher (approved):
This looks good.  I tested manually and read over the code.

It's kind of crazy how many types of admins we have.  Looks like running into the `type` field was fun.  :-)

PR #480, opened 2021-05-18
Removes passenger from our deploy and code

sfisher:
And this adds puma to our gemfiles, etc, though I don't know if it will be really needed there if puma is managed outside of capistrano.

I'm creating a new branch `deploy-update-main` to accumulate our changes for moving to new deployment models and merging this change into that rather than into our regular main branch which we probably still need for normal development before the deploy changes all roll out.

I tested out our deploys without passenger and everything deploys OK, but it just doesn't start or stop the server as part of Capistrano like it used to.  I tested with the puma-systemd stuff, but doesn't work because of permissions and I think it will not be managed that way (through capistrano) anyway?


ryscher:
Looks good, works fine for me. Thanks for removing all the cruft.

PR #481, opened 2021-05-19
Reinstate local counter

sfisher:
This re-instates the counter stats so they just come from our database.

It also has a little bit of rework in the stat generation so it tallies all of them from the reports into a hash in memory and then writes the 40,000 or so updates to the database in one pass.  This is instead of tallying from the database like we did before for each month which took a lot longer.

Add a test which checks the tally results for 6 items and 2 months of reports.


ryscher:
Sigh. Sorry you had to do this again!

PR #482, opened 2021-05-19
Default value for manuscript number in Editorial Manager bridge

ryscher:
If Editorial Manager does not provide a manuscript number, fall back to use deposit ID, or as a last resort fill it with a generic string. This allows users to progress through the submission system without worrying about the manuscript number. Editorial Manager will store and use the dataset DOI, so we do not need the manuscript number for later matching.


sfisher (approved):
Seems sensible.  Thanks Ryan.

PR #484, opened 2021-05-20
Curator assignment and filtering

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1232

On the "Admin Dashboard" page:
- Allows curators to "assign" themselves or other curators to a dataset.
- Adds a Curation Activity to record the fact that someone changed the assignment, with an optional comment made by the curator who changes it.
- When a dataset is versioned by the author, automatically re-assigns the same curator as the previous version
- Adds the capability to filter by Curator


sfisher (approved):
Look pretty good to me, though I wasn't quite sure where to find this in the UI and I think maybe I didn't look in the right place.

PR #485, opened 2021-05-21
Add note about ensuring tenant

ryscher:
When creating an account for API-only use, sometimes the `tenant_id` does not get set. This adds a step to the documentation so we ensure it is set properly.


sfisher (approved):
Thanks for adding this, Ryan.

PR #486, opened 2021-05-21
Disallows submitting if over size or file limits

sfisher:
Changes to file limits:
- No more per-version lower size limits for direct uploads.  If they can get them into S3 (and if validates they're there before submission) then they can upload the sizes they want.  Larger files have more chances of upload problems, though, so some people might like to use the FTP server and URLs still.
- Size and number of file limits moved out of individual tenants to global configuration.
- Checks each of the three file types separately and with separate size limits for Zenodo (50GB) vs Dryad (300GB).  Each type has a 1,000 file limit (3,000 if you upload 1,000 to each).

In addition to refactoring, changing tests for file limits, I removed all the extra file upload templates we were no longer using.  Also removed some pieces of controller actions, some actions and some filters and routes for things not used anymore.  I believe I tried all the actions in the UI for file uploads and everything works and I didn't remove anything still needed.

To test limits manually:
1. Edit `app_config.yml`
2. Change something like this, which sets to 1 MB, 500 KB and 5 files.
  ```
    maximums:
      merritt_size: 1_000_000
      zenodo_size: 500_000
      files: 5
  ```
3.  exceed file size or number for a type and it should display an error and disallow submission on the review page and indicate where you've run over.


ryscher:
This is great. Thanks for moving the configs out of the tenants, and thanks for tracking down all of the old code that we don't need any more.

PR #487, opened 2021-05-21
Frictionless manifest files

cacods:
- Implements frictionless validation for manifest files right after adding them to the files table.
- Column Tabular Data Check have 4 status: "Checking...", "Issues Found", "Passed", "N/A".
- "**Checking...**": validation going on in backend via ajax request.
- "**Issues Found**": validation returned report with errors -- new entry created in the database.
- "**Passed**": validation returned report without errors -- entry doesn't created in the database.
- "**N/A**": file is not csv.

"Checking" status may be transitory, and replaced in the future).


cacods:
I think the Upload to S3 functionality is up again.
Please note that when arriving/refreshing Upload Files page, non-manifest files will appear as "Passed" in Tabular Data Check column since they don't have frictionless report. I'm working on frictionless validation for non-manifest files now.

sfisher (approved):
Thanks for all this work, Cassiano.

I tested out and it seems to work.  Also tested the regular files still upload and it seems good.

PR #488, opened 2021-05-21
Updating ucop ezid prefix

sfisher:
This is a small change for UCOP identifiers and just changing the prefix.


ryscher:
Yay for persistent identifier prefixes?

PR #489, opened 2021-05-21
Simplifying installation of solr schema

sfisher:
These are very small changes, but simply copy over some of the extra files form the solr schema for geoblacklight into our own repo and simplify the install instructions by just copying the config files out of our repository rather than getting them from two places.


ryscher:
Thanks for simplifying this.

PR #492, opened 2021-05-24
Both new and existing files show as "uploaded"

sfisher:
The "new" terminology was confusing since some "new" files had been uploaded a while back in previous versions.

Instead, changing the term to "uploaded" which shows for both newly uploaded and previously (long ago) uploaded items.

This is what shows in the table of uploads on page 2 of the process.  Page 3 still shows new next to new items in this version, nothing (and an underlined link) for submitted versions and a strikethrough for removed items.


cacods:
It makes sense. So the misleading "New" term for previous versions gets solved.
It's possible to have other labels as for that derived from database (created, copied), but I think it's great to have that fixed for this release.

ryscher:
This looks fine, and should be a better user experience.

PR #493, opened 2021-05-25
Allow forcible editing of datasets

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1199

Gives superusers a button that allows them to edit a dataset, even if the dataset is still "in progress" by another user. Because this could lead to unfortunate race conditions, the button is located out of the way at the bottom of the Activity Log page, AND it is marked as "dangerous".


sfisher (approved):
I tested this out and it looks good and works well.  Thanks Ryan.

PR #494, opened 2021-05-25
Add user_name to journal review notices

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1255

Adds the submitter's name to notices that journals receive for review purposes. Also, rearranges the mailer so @user is available in all emails.


sfisher (approved):
Looks good.  Thanks, Ryan.

PR #495, opened 2021-05-26
Cache list of curators for faster query processing

ryscher:
When there is a large number of datasets, we don't want to re-query the DB for the list of curators for every single row in the table.


sfisher (approved):
Thanks for looking at this, Ryan.  Looks good to me.

PR #496, opened 2021-05-26
Improve caching of curator_ids

ryscher:
Sorry to change this around... the approach in https://github.com/CDL-Dryad/dryad-app/pull/495 had the major drawback of not updating the list of curators until the server restarted. So instead of using a constant, I moved the cached value into the `where` method.


sfisher (approved):
This makes sense.  During my meeting, I wondered if the constant wouldn't get reloaded.  This seems to make sense.

I thought maybe the curators only needed to be loaded in the UI page but then also see the @editor_id in there, so makes sense.

PR #497, opened 2021-05-26
Add submission queue checker to status dashboard

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1215

Add a simple checker for the submission queue. This will help us to detect times when we shut down the queue but forget to re-enable it.

Also, combined the two documentation files that described the status dashboard.

Like all items on the status dashboard, this requires a database entry. To add this, run `bundle exec rails status_dashboard:seed`.


sfisher (approved):
This looks good.  Thanks for adding this and the instructions about how to set up these checks.

PR #498, opened 2021-05-27
Frictionless s3 files

cacods:
Implements frictionless validation for S3 uploaded files: triggers the frictionless validation right after the file is uploaded successfully to S3.


stash/stash_engine/app/models/stash_engine/generic_file.rb
```diff
@@ -137,22 +136,35 @@ def self.sanitize_file_name(name)
...
 
-    def call_frictionless(file)
-      `frictionless validate #{file.path} --json`
+    def download_file
+      http = HTTP.use(
+        normalize_uri: { normalizer: Stash::Download::NORMALIZER }
+      ).timeout(connect: 10, read: 10).follow(max_hops: 10)
+      dl_url = url || direct_s3_presigned_url
+      result = http.get(dl_url)
+      begin
```
sfisher, line 152: I think you want the "begin" before the "http.get" line.  Most of the time it will not matter.

sfisher (approved):
I tested and it mostly works.  Though one time that I uploaded the 4 test files at once and some didn't ever validate.  I think this might be the timing problem we talked about this morning.

If you'd like to do a different pull request to work on those things can accept and merge this now.

stash/stash_engine/app/models/stash_engine/generic_file.rb
cacods: Yes, I noted that `http.get`, in the wrong place. I'm working on fixing that now @sfisher. But this is to capture the exception. Probably it still have to tolerate some delay for downloading from s3. I will continue with these things tomorrow.
Thanks for testing that.

PR #499, opened 2021-05-27
Make the API more robust to errors in Editorial Manager metadata

ryscher:
Editorial Manager sometimes sends an `author` field, and sometimes `authors`. Sometimes it's an array, and sometimes a single value. This PR allows for any combination of these cases...


sfisher (approved):
This seems ok.  Kind of odd that they were off on this sometimes but not others.  :-)

PR #500, opened 2021-05-27
Adding to the rake task so it runs as part of checker

sfisher:
This is something I forgot to add for verifying the weekly task that checks for new citations from datacite event data.


ryscher:
Thanks for adding this. It will be good to have it checking regularly.

PR #503, opened 2021-05-31
Catches exceptions downloading file and writing to file: displays val…

cacods:
…idation error label in these cases.

So the user can remove the file and upload it again for the validation process to run again.


sfisher (approved):
Thanks, Cassiano!

It looks good.  I also tested and everything works great.

PR #504, opened 2021-06-01
Clean out temporary content

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/560

Rake task to clean old datasets and temporary files. If a resource has been `in_progress` for more than a year, it is removed. If this resource is the only one attached to a given identifier, the identifier is removed as well. After cleaning the resources, the script cleans directories on S3. An S3 directory is removed if it (a) does not have a corresponding resource, or (b) the corresponding resource entered `submitted` status more than a month ago. (We don't want to delete S3 directories immediately after the resource is submitted, because we need to allow time for copying to Zenodo, possibly recovering from a failure of the transfer.)

To run:
`bundle exec rails identifiers:remove_old_versions`

If you set the environment variable `DRY_RUN=true`, you can run this script without actually deleting content, to see the output indicating which items will be deleted.

### DIFF
--- a/stash/stash_engine/lib/tasks/stash_engine_tasks.rake
+++ b/stash/stash_engine/lib/tasks/stash_engine_tasks.rake
@@ -132,6 +132,7 @@ namespace :identifiers do
     # Remove resources that have been "in progress" for more than a year without updates
     StashEngine::Resource.in_progress.each do |res|
       next unless res.updated_at < 1.year.ago
+      next unless res.current_curation_status == 'in_progress'
 
       ident = res.identifier
       s3_dir = res.s3_dir_name(type: 'base')
--- a/stash/stash_engine/lib/tasks/stash_engine_tasks.rake
+++ b/stash/stash_engine/lib/tasks/stash_engine_tasks.rake
@@ -168,8 +168,9 @@ namespace :identifiers do
 
       if StashEngine::Resource.exists?(id: res_id)
         r = StashEngine::Resource.find(res_id)
-        if r.submitted? && r.updated_at < 1.month.ago
-          # if the resource is state == submitted and has been for a month (so zenodo has processed), delete the data
+        if r.submitted? &&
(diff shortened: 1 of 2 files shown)


stash/stash_engine/lib/tasks/stash_engine_tasks.rake
```diff
@@ -127,6 +127,60 @@ namespace :identifiers do
...
+      res_id = if id_prefix.include?('-')
+                 id_prefix.split('-').last
+               else
+                 id_prefix
+               end
+      puts "checking S3 key #{s3o.key} -- id_prefix #{id_prefix} -- res_id #{res_id}"
+
+      if StashEngine::Resource.exists?(id: res_id)
+        r = StashEngine::Resource.find(res_id)
+        if r.submitted? && r.updated_at < 1.month.ago
```
sfisher, line 171: We might just want to check that things we may be storing have successfully transferred (or in addtion).

The query for that would be
```
r.zenodo_copies.where("copy_type LIKE 'data%' OR copy_type like 'supp%'").where.not(state: 'finished').count == 0
```

If it's zero, they're all finished.  If it's 1 or more then there is an unfinished replication.

I don't expect it to go over a month for the S3 things (though we have some Merritt things that haven't replicated for months because of encoding issues).

sfisher (commented):
This looks good, @ryscher .

The only thing I suggest is to check the zenodo replications before deleting.  Just to be sure we're really finished it.  I suspect a month is fine (it has been so far), but just for safety.

stash/stash_engine/lib/tasks/stash_engine_tasks.rake
sfisher: Ooops.  The above should be `("copy_type LIKE 'software%' OR copy_type like 'supp%'")`

ryscher: Thanks, this is updated now.

PR #506, opened 2021-06-02
Improve ISSN robustness

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1282

When submitters are entering a journal name in the autocomplete box:
- Only erase an existing ISSN when a non-recognized journal name is being saved. This prevents shortcuts like CMD+Tab from erasing the ISSN when it would otherwise be valid.
- When a non-recognized journal name is being saved, first check whether it exactly matches a known name (just with the extra asterisk). This allows the journal to be recorded properly if users type it exactly, but don't actually click on the autocomplete.


sfisher (approved):
These look like great changes.  Thanks for doing this, Ryan.

PR #509, opened 2021-06-03
Update documentation for deleting objects

ryscher:
Update the outdated documentation that was on Confluence, and move it here, since it isn't particularly secret.


sfisher (approved):
Thanks for updating this, Ryan.  It looks really good.

PR #510, opened 2021-06-04
Dashboard for tracking the zenodo failures

sfisher:
We need to deal with problems at Zenodo, especially if they go on for a while.  It is a pain to deal with them and figure out what is going on from the database.  Also, when I use my 350 hours of vacation, someone else will need to deal with at least the urgent things sometimes.  ;-)  Zenodo also seems to frequently have problems during their nighttime.  Errored items get attempts at resubmission on a cron, but it may not catch everything.

This should make it quite a bit easier to at least resubmit, see and diagnose problems.

- Adds menu for superusers "Zenodo Submissions"
- Shows all items that aren't finished, pretty much using the same query I've been using for months manually.
- Clicking the ID gives full details for delayed job and our internal zenodo job information.
- Clicking identifier id gives just the errors for that dataset.  Sometimes it's a lot easier to just see one dataset so you can submit things in order if there is more than one failure backed up.
  - (Each update to software & supplemental goes in zenodo and creates versions there.)
  - (Only published changes update for extra copies in zenodo.)
- Error info in mouseover or by clicking into details.
- Action to resend an item into the delayed job queue.
- Action to set state to *error* for stalled items which happens when Zenodo has problems or huge submissions .  It detects stalled items when they've dropped out of the delayed job queue, but are in non-error states in the zenodo job history.

Probably to test, you can look at our development and we have a bunch of junk data in there already that has failed.  You'll see pre-failed things you can try submitting or look at.

You can also submit a dataset.  The supplemental or software should show up right after being submitted and then disappear once successful.  The data replication should show up briefly after publishing.

I haven't done a lot as far as testing of this and most is very simple and basically views into the database.  If you'd like I can go back and add some tests, but I just wanted to get something out there we can begin using so it's not such a burden to deal with the failures.

### DIFF
--- a/stash/stash_engine/app/assets/javascripts/stash_engine/zenodo_queue.js
+++ b/stash/stash_engine/app/assets/javascripts/stash_engine/zenodo_queue.js
@@ -1,0 +1,2 @@
+// Place all the behaviors and hooks related to the matching controller here.
+// All this logic will automatically be available in application.js.
--- a/stash/stash_engine/app/assets/stylesheets/stash_engine/zenodo_queue.css
+++ b/stash/stash_engine/app/assets/stylesheets/stash_engine/zenodo_queue.css
@@ -1,0 +1,4 @@
+/*
+  Place all the styles related to the matching controller here.
+  They will automatically be included in application.css.
+*/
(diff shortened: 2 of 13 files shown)


ryscher:
Thanks, Scott! This is great! Will be very helpful to manage these items in the future.

PR #511, opened 2021-06-04
Do not display or count citations for random URLs that don't have citation metadata

sfisher:
We were displaying citations for some items in DataCite Eventdata that had non-doi citations like random links to github or other places.

- if the html citation text is unavailable for lookup from datacite then return a nil
- where it maps citations, it is rejecting blanks which will eliminate those nils
- I manually ran a couple of queries to remove the non-doi citations and to update the counts in our database
- Added a test for this returning nil when no citation available, seems like this class didn't have tests yet.

https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1285


ryscher:
Looks good!

PR #512, opened 2021-06-07
Update staff listings

ryscher:
Replace Tracy with Trisha, update curators.


sfisher (approved):
This all looks good to me, Ryan.

I shortened by bio a bit since it's boring.

I didn't know you made robots.  Sounds like fun.

PR #513, opened 2021-06-07
Frictionless components

cacods:
- Gets rid of Bootstrap CSS CDN import statement and incorporates only the Bootstrap CSS classes that frictionless-package effectively uses.
- To update the previous package run:
`$ yarn install`


sfisher (approved):
Thanks for doing all the work to get the styles working better, Cassiano.  It looks nice and the other styles aren't changed.

cacods:
> Thanks for doing all the work to get the styles working better, Cassiano. It looks nice and the other styles aren't changed.

Glad to read that Scott. Thank you!

PR #514, opened 2021-06-08
Allow Editorial Manager to update manuscript number

ryscher:
Once a dataset has been created using the Editorial Manager bridging API, we normally do not allow the metadata to be updated, because subsequent updates should be performed through the web interface. This adds an exception to the processing, so the manuscript number may be updated. This is due to the workflow of Editorial Manager, where they may initialize a dataset and then quickly call the API again to update the manuscript number.


sfisher (approved):
Looks good to me.  Thanks Ryan.

PR #516, opened 2021-06-08
U Wisconsin: Convert hyphen to en-dash

ryscher:
Match the en-dash that is listed in ROR, so the text matches against our autocomplete component.


sfisher (approved):
Thanks Ryan.  This would be hard to catch since they look pretty much the same.

PR #517, opened 2021-06-10
Small fix for edge case where some items fail midway

sfisher:
- Moves method from mixin class to individual class since it isn't used outside one class right now
- If a deposition_id (zenodo id) is already set for an item it doesn't try to find it from a previous deposit or reset it to nothing.


ryscher:
Looks good. Thanks for the simplification; I always forget about mixins.

PR #519, opened 2021-06-10
Reconfiguring priority according to the Delayed Job docs

sfisher:
Gives different queues different priorities and we give each type of submission a named queue.

This should basically make the data replication happen after supplemental and software.

Also, changed delayed job retries to 2 since at times when Zenodo has issues we might want to reduce retrying right then and wait for our cron to re-launch the errors for retry the next day.


ryscher:
Thanks. Low numbers are high priority for Unix `nice`. I wonder whether they are actually using it to manage the processes.

PR #520, opened 2021-06-10
Add absolute URLs for EM

ryscher:
Editorial Manager wants absolute URLs instead of relative URLs. (Note that this change only applies to the EM API calls, so it does not affect anyone else.)


sfisher (approved):
Looks good to me.  Thanks, Ryan!

PR #521, opened 2021-06-11
Improve robustness of DOI handling in API

ryscher:
This was motivated by EditorialManager, but it applies to all API calls.

When updating a dataset, the DOI may be specified:
- as part of the URL or in the body of the request
- with or without the 'doi:' prefix
- as a field named `id` or `deposit_id`


sfisher (approved):
Looks fine, @ryscher  .

It's too bad they're not paying attention to this detail consistently.

PR #523, opened 2021-06-14
Improve defaults for manuscriptNumber and title

ryscher:
These defaults need to take effect even when the request does not contain an 'article' section.


sfisher (approved):
Thanks, Ryan!

PR #524, opened 2021-06-15
Quick way to get migration stats (builds on previous pull request)

sfisher:
- Makes it easier to get migration stats when people want updates
- Link the DOI to the landing page so we can see what metadata is missing from old items and why Zenodo will not allow submission
- If the abstract is missing (which it seems to be on some old items), put in the fake abstract "No abstract available" since otherwise Zenodo will not take them

Example run of stats:

```
$ RAILS_ENV=production rake zenodo:migration_stats
Running main rspec tasks...
 -- environment production
Stats for old datasets being copied to Zenodo--excludes items already sent since we began replicating
6372 of 32165 old datasets have been replicated
19.81% by number of old datasets have been replicated
995.35 GB of 21.05 TB of the old datasets have been replicated
4.73% complete by size
Optimistic completion date: 2021-08-14
```

I think the completion date probably isn't realistic since we're leaving all the largest and most problem-prone ones until last so my guess is that it will be a month or two later than that.  It is arrived at based on a "progress bar" type calculation of speed since we started and size remaining.


ryscher:
Thanks, Scott. This is very useful. Will wait to merge until we resolve the question about the previous branch.

PR #525, opened 2021-06-16
Gets frictionless to work on our servers

sfisher:
There are some environmental difficulties on our servers and this is a little hacky.

But this will work until we can get a better systemctl/system.d service set up for initializing the pyenv as part of starting puma.


ryscher:
This looks fine.  I was able to get it running after installing the correct version of python, though there was another error that kept the validation from completing.

PR #527, opened 2021-06-16
UIUC libraries tenant setu

sfisher:
Wow, this one turned into way more of a pain than I ever thought.  They had about 20 shibboleth entries and hard to find the right one and the XML is hard to find the entity ids from the name when they are hundreds of lines apart.  I think I finally found the main one and not ones for all the different services they're associated with.

Fixed their logo since it had tons and tons of whitespace and made it look like a little dot on the screen.

I think it's all fine now.
Screen Shot 2021-06-16 at 4 38 12 PM


ryscher:
Looks good. Sorry the shibboleth was such a pain!

PR #530, opened 2021-06-17
Handle presigned URL errors and rope them into same exception that gets logged

sfisher:
This code was simple, but took me a lot of hours because it kept resending the Merritt exception when I had both handled by the same block.  Maybe the `zenodo_replication_url` in there.  But unfortunately, it never gave me any indication of why and just raised the original (wrong) exception.

Very frustrating, but should be working now.


ryscher:
Thanks for working through this issue. All looks good.

PR #531, opened 2021-06-18
Frictionless validation status

cacods:
- Fixes reading the file path when calling frictionless validation. In some fill cases the temp file were removed before frictionless process could read it.
- Changes the temp dir from system temp to dryad tmp. This must fix the safe dir error returned from frictionless validation as related by Ryan.
- There is a missing thing about updating the status on refreshing the page: refreshing the page returns the current status. So if the status is "Checking..." yet, it will not be updated after validating is complete. User has to refresh the page again. This can be solved sending ajax requests periodically until getting the stable statuses (`no issues`, `issues`, `error`).

### DIFF
--- a/app/javascript/containers/UploadFiles.js
+++ b/app/javascript/containers/UploadFiles.js
@@ -432,7 +432,7 @@ class UploadFiles extends React.Component {
         });
 
         const countRepeated = urls.length - newUrls.length;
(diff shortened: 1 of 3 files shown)


sfisher (approved):
Thanks for doing all this, Cassiano.  Thanks for the tests and also tested it manually and rad through the code and it looks good to me.

PR #532, opened 2021-06-18
Improve robustness of email parsing

ryscher:
Adds some extra protection against malformed metadata emails from journals. It was possible to create an email message that had enough metadata to start parsing, but if it did not have a manuscript number, the parsing would error and prevent subsequent messages from being processed.


sfisher (approved):
Thanks for catching this, @ryscher .

PR #533, opened 2021-06-21
Save tempfiles in binary mode

sfisher:
Avoids any weirdnesses with encoding which we really don't want to deal with and should keep file as is without changes.

https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1323


ryscher:
Thanks for finding this!

PR #534, opened 2021-06-21
Gets version and REVISION if they're available in rails.root

sfisher:
Displays version in the footer and also adds revision in a comment so if you really need to know the revision hash, you can look at it in the source code inside a comment.


ryscher:
Thanks! I've wanted to have this for a long time, but never took the time to add it in.

PR #535, opened 2021-06-21
Small change to not cause error if descriptions are missing

sfisher:
Uses the safe navigation ampersand to prevent previous error.


ryscher:
Looks good.

PR #536, opened 2021-06-21
Journal info page

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1291

Replace the outdated JournalLookup PDF with a page that displays "live" information from the database. This is a publicly-viewable page, but its presence is not advertised to the public. The only way to navigate to the page is via the (superuser) Admin menu.


sfisher (approved):
Thanks, Ryan.  This looks really helpful.

PR #537, opened 2021-06-22
Frictionless reports versioning

cacods:
Copies the frictionless reports for the files when making a new Resource version.

Note the propagate method from amoeba gem in GenericFile model. It's the way amoeba works on STI fields.


sfisher (approved):
Thanks for finding this, Cassiano and doing the tests.

It makes sense that we had to add something else because of Single Table Inheritance.

PR #538, opened 2021-06-22
In API, remove HTML from titles

ryscher:
When datasets are created via the API, remove any HTML tags that are embedded in the title string.


sfisher (approved):
Thanks, Ryan, this works great.

PR #539, opened 2021-06-22
Don't Display version in production

sfisher:
Except to superusers.


ryscher:
Thanks for the quick update.

PR #540, opened 2021-06-23
Default sort to title

ryscher:
The SortableTableHelper defaults sort order to `created_at`, which is useful for most models, but not for journals. This changes the journal screen's default sort and secondary sort to use `title`.


sfisher (approved):
👍

PR #541, opened 2021-06-23
Frictionless components setup

sfisher:
This updates the application to use the public npm library from the cdl-dryad organization and also updates documentation so we know where to find more about this library and how to work with it.


cacods:
I think it's all OK. Thanks for updating the docs for the frictionless-components @sfisher.

cacods (approved):
I think it's all OK. And thanks for updating the docs.

PR #544, opened 2021-06-24
EM updates for the review url

ryscher:
Two updates for Editorial Manager:
- re-order processing for requests that come in after a user has edited the dataset, and only return an error when no valid changes are requested
- for the field `deposit_url`, use the sharing link if one is available


sfisher (approved):
Seems good to me.

You're so nice to them, making special case changes to make their code easier.

PR #545, opened 2021-06-25
Extends frictionless validation to Excel (xls, xlsx) e json files

cacods:
Now validation is made also for Excel and json files.


cacods:
Wrong target branch.

PR #546, opened 2021-06-25
Extends frictionless validation to Excel (xls, xlsx) e json files

cacods:
Now validation is made also for Excel and json files.


sfisher (approved):
Thanks, @cacods 

This works great.  I'll deploy to the servers now, also.

PR #548, opened 2021-06-29
Frictionless file size limit

cacods:
I opened a new entry in `app_config` for setting up the max file size limit that triggers the frictionless validation.

I put 6MB as the max size, but it's just a matter of changing the value -- with this limit I'm getting something like 7 to 10 seconds for each file chosen from file system in my local machine (I guess I observed more or less these same times for a couple of manual tests in dev server). For the files that come from URL's, this time has to be added for each file that go to the table.


sfisher (approved):
This looks like a good change to limit by size!  Thanks for adding to the config and also to the example.

PR #549, opened 2021-06-29
Any frictionless error when calling external service saved as status "error"

cacods:
Save status error for any kind of error not related with the normal tabular validation. This is supposed to occur very rarely, but if occurs, display "error" for the user. Furthermore, puts in `stdout` a message telling about the error, and saves the report with the error. So it can be inspected.

### DIFF
--- a/stash/stash_engine/app/models/stash_engine/generic_file.rb
+++ b/stash/stash_engine/app/models/stash_engine/generic_file.rb
@@ -153,21 +153,21 @@ module StashEngine
     def validate_frictionless
       result = download_file
       if result.instance_of?(HTTP::Error)
-        @report.update(generic_file_id: id, status: 'error')
+        @report.update(status: 'error')
         return
       end
       result = write_tempfile(result)
       if result.instance_of?(Errno::ENOENT)
-        @report.update(generic_file_id: id, status: 'error')
+        @report.update(status: 'error')
         return
       end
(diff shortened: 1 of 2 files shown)


stash/stash_engine/app/models/stash_engine/generic_file.rb
```diff
@@ -153,21 +153,21 @@ def set_checking_status
...
       end
-      # TODO(#1296): rescue from errors here!
       result = call_frictionless(result)
       # Add 'report' top level key that is required for calling
       # React frictionless-components
       result_hash = { report: JSON.parse(result) }
       if validation_error(result_hash[:report])
-        @report.update(status: 'error')
+        @report.update(report: result_hash.to_json, status: 'error')
+        puts "Some error occurred calling frictionless. See database for report_id #{@report.id}"
```
sfisher, line 170: Hey could you change this `puts` to `logger.error` which should log any errors to the log files as an error.

sfisher (commented):
Besides the puts/logger thing I think this all looks pretty good.

stash/stash_engine/app/models/stash_engine/generic_file.rb
cacods: It makes sense @sfisher. I will change that and submit a new PR.

PR #550, opened 2021-06-29
More basic layout changes according to wireframes and accessibility

sfisher:
- Frictionless validation shows differently for different types of results.  Not just text.
- buttons instead of links for the actions according to accessibility
- No blinking and instead spinners and style change
- Warning icon
- Fixing some styles leftover from earlier work that React warned were wrong.
- Testing changes


cacods (approved):
Thanks for spending time on adapting the styles and texts for Dryad Scott. That was a thing I was supposed to do.

cacods:
Maybe this failed test is some of the ones that requires some `sleep` I guess, not sure about. Running some times now and not getting it failed.

PR #551, opened 2021-06-30
Fix progress bar for Firefox

cacods:
I noted that the Firefox browser started the progress bar full, when uploading, instead of empty. This fix this wrong behavior.


sfisher (approved):
Looks good.  Seems smart to set to zero instead of blank.

PR #552, opened 2021-06-30
Fix upload button

cacods:
Sometimes the 'Upload pending files` button could be enabled even without the input check for confirming no Personal Health Information... checked.

Now, always the user add Pending files, the input check is unchecked with the button disabled.


sfisher (approved):
This looks like a good small fix.  Thanks, Cassiano.

PR #553, opened 2021-06-30
Frictionless integration docs

cacods:
- Include instruction for upgrading the frictionless package.

- Change the instruction for running frictionless inside ruby/rails.

Before:
```
result = `frictionless validate <tabular_file_path>`
```

After:
```
result = `frictionless validate --path <tabular_file_path>`
```

Note that `--path` was included as a flag for the `frictionless validate` command. This is necessary for `json` files. But it's not hurt to add that always, so the call to frictionless stay unified.


cacods:
Erroneously made for merge with main.

PR #554, opened 2021-06-30
Frictionless integration doc

cacods:
Include instruction for upgrading the frictionless.

Change the instruction for running frictionless inside ruby/rails (already in the call from code).

Before:
```bash
result = `frictionless validate <tabular_file_path>`
```
After:
```bash
result = `frictionless validate --path <tabular_file_path>`
```
Note that `--path` was included as a flag for the `frictionless validate` command. This is necessary for `json` files. But it's not hurt to add that always, so the call to frictionless stay unified.

### DIFF
--- a/app/javascript/components/ValidateFiles/ValidateFiles.js
+++ b/app/javascript/components/ValidateFiles/ValidateFiles.js
@@ -5,7 +5,7 @@ const validate_files = (props) => {
     if (props.checkConfirmed) {
         checkConfirm =
             <div>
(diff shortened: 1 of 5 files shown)


cacods:
BTW, upgraded in my local machine from 4.10.0 to 4.12.4 and all seems good.

I was in touch with Evgeny from OKFN about some apparently misvalidations for Excel files, and looks like the current version fixes that misvalidations.

sfisher:
Thanks for the instruction updates.

PR #556, opened 2021-06-30
Remove some deprecated TODO's and solves one

cacods:
Just removing some TODO stamps that is not valid anymore. And solving one in main React component.


sfisher (approved):
Thanks for the cleanup and updates!

PR #555, opened 2021-06-30
Enhanced shopping cart reporting -- deferred journals

ryscher:
(for https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1281)

Add a secondary set of reports that can be run after the main shopping cart report. These reports replace the brittle R scripts that were previously located in https://github.com/datadryad/journal-payments/

To run:
1. Ensure at least one journal in the database has `payment_plan_type` of `DEFERRED`, and this journal has an associated `sponsor` (StashEngine::JournaOrganization)
2. Publish a dataset associated with this journal, or add the journal's ISSN to an already-published dataset.
3. Generate a shopping cart report for the month the dataset was published:
  `bundle exec rails identifiers:shopping_cart_report YEAR_MONTH=2021-06`
4. Run the deferred reports, using the shopping cart report as input:
   `bundle exec rails identifiers:deferred_journal_reports SC_REPORT=shopping_cart_report_2021-06.csv`

PR #557, opened 2021-06-30
Improve efficiency of shopping cart report

ryscher:
(Builds on #555, for https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1281)

Improves the process for selecting datasets, so the shopping cart report does not need to examine every dataset in the system. 

Note that there is still the potential for the report to examine a large number of items, since bulk updates will change the `updated_at` field, forcing many items to be examined to determine the datasets that are actually included in the report. If this becomes an issue, we could increase reporting efficiency by explicitly storing `approval_date` in the identifier table, instead of recalculating it. But I don't think it is useful to do that prematurely, because it would introduce additional complexity that would be annoying to maintain.


sfisher (approved):
Thanks, Ryan.  Looks like it will be faster.

PR #560, opened 2021-07-02
Lists all the bad things you did in tabular data

sfisher:
*includes code from previous pull requests, so the diff should be more clear if you accept those in the order they were put in*

The list already shows lists of items with issues, but some wireframes wanted to show extra information before the list of files.

IDK if this really adds much information that isn't already present, but I suppose it make things more visible if people don't look at their upload table ever again.

To test two different ways:
- Upload a file that's not a CSV by changing the file extension from jpg or something.  It shouldn't parse and give big red warnings before the upload table.
- Upload a csv or similar file with consistency issues and it should also state which files are messed up in case you can't see them in the table below.  🤷‍♂️ 

I added some jest tests for the javascript component that does this.

### DIFF
--- a/app/javascript/components/FileList/BadList/BadList.js
+++ b/app/javascript/components/FileList/BadList/BadList.js
@@ -1,0 +1,55 @@
+import React from "react";
+
+const badList = (props) => {
+  const errorFiles = filterForStatus("error", props.chosenFiles);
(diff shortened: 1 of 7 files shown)


ryscher:
Looks good. It looks a little strange to me that the error box is the full width of the page, while the table of files is narrower, but I guess the designers can worry about that.

PR #561, opened 2021-07-02
Disable navigation while frictionless is validating

sfisher:
*includes code from previous pull requests, so the diff should be more clear if you accept those in the order they were put in*

This pops up a dialog asking people to wait if they attempt navigation away while frictionless checks are happening.

to test:
- Upload a frictionless type such as csv or xls.
- While the spinners under the frictionless column are spinning, click the button to navigate to the review page or someplace.
- Should see message asking you to wait to navigate away.


ryscher:
Great! Thanks for adding this and cleaning up the "not finished uploading" notification as well.

PR #562, opened 2021-07-06
Tune timeouts retries

sfisher:
- Only retry 5 times in a row instead of 20 which takes forever and if Zenodo is that unstable maybe error and try again later instead of now.
- Add write timeout for 60.
- I don't believe the overall timeout is for the full request, but only for each chunk sent.  I don't think setting this to 6 hours helped.


ryscher:
Looks good. Sorry you still have to work around these issues.

PR #564, opened 2021-07-09
Small tweaks that I hope make Zenodo replication better

sfisher:
- Up read and write timeouts to 3 minutes (for chunks or smaller requests, not for the whole request)
- Add additional clause so it ignores later items in same resource for zenodo deposit (publication action after uploading if both had errors or aren't complete)
- If previous version with files for Zenodo never replicated, then include information about which one is the problem in the error message


ryscher:
Great, hope these help!

PR #565, opened 2021-07-12
Allow EM to update more fields

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1362

Normally, we don't want API users to change metadata after an author has visited the Dryad UI. However, since EM does not have much metadata until the very end of their process, we want them to be able to update some fields after the author has visited Dryad. This PR allows authors, funders, and keywords to be updated. If any of these fields are updated, a curation note is added to alert curators that the dataset has been updated.

### DIFF
--- a/stash/stash_api/app/controllers/stash_api/datasets_controller.rb
+++ b/stash/stash_api/app/controllers/stash_api/datasets_controller.rb
@@ -102,25 +102,6 @@ module StashApi
         fields_changed << 'Funders'
       end
 
-      # Add co-authors if they were not added by the submitter
-      if @resource.authors.size < 2 && params['authors'].present?
-        curr_auths = @resource.authors.map(&:author_full_name)
-        params['authors'].each do |auth|
-          next if curr_auths.include?("#{auth['last_name']}, #{auth['first_name']}")
-
-          new_author = StashEngine::Author.create(author_first_name: auth['first_name'],
-                                                  author_last_name: auth['last_name'],
-                                                  author_email: auth['email'],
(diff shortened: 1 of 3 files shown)


stash/stash_api/app/controllers/stash_api/datasets_controller.rb
```diff
@@ -137,8 +176,10 @@ def em_update_status
     def em_reformat_request
       em_params = {}.with_indifferent_access
 
+      # EM doesn't set an item owner, so default it to system, and the actual user will need to claim it
+      em_params['userId'] = 0
```
sfisher, line 180: I'm not clear of all the flow here.  Will it always be user 0 later in the process if they come back for these updates?

sfisher (approved):
Looks pretty good to me, just didn't know about always being user zero.   If that's the right flow then go ahead and merge.

stash/stash_api/app/controllers/stash_api/datasets_controller.rb
ryscher: Yes, it should stay 0 (unclaimed) for any calls that EM makes prior to an actual user logging in to Dryad. Once a user logs in and submits the item, it will be assigned to that user and subsequent EM calls will go to `em_update_selected_fields`, which does not allow changes to the ownership.

sfisher: 👍

sfisher:
Now rubocop is freaking out.  😿

PR #566, opened 2021-07-12
Remove related works for zenodo after all files are removed for resource

sfisher:
Before now, the relations would set the last available relationship that had been sent to zenodo.  It was defined that way in the ticket info.

Now we are removing the relationship if someone removes all the files at zenodo and it doesn't point to an older relationship that is still there for an older version.

To test:

- Upload  a file to either supplemental or software and submit the dataset.
- Wait for replication to complete to zenodo (usually quick on their sandbox)
- Check that the relationship exists for that external zenodo item. ✅ 
- Edit the dataset in a new version
- Remove the supplemental or software file you added earlier and then submit the dataset again
- Wait for updates (should happen fast)
- Check to see that the relationship for the zenodo dataset with all files removed has been removed from relationships.  ✅

### DIFF
--- a/spec/models/stash_datacite/related_identifier_spec.rb
+++ b/spec/models/stash_datacite/related_identifier_spec.rb
@@ -140,9 +140,9 @@ module StashDatacite
         expect(re.added_by).to eq('zenodo')
       end
 
-      it "removes an existing relation if all files have been removed from zenodo" do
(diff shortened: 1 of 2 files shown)


ryscher:
Looks good, works in my testing. Thanks for reworking this.

PR #567, opened 2021-07-13
Removes the javascript that hides the remove link

sfisher:
People were not allowed to remove all of them because some items require at least one, but for funders it is optional so it is fine to remove the last one.

You can always click the + button if you want to add one and they've all been removed.  Loading the page also creates a blank one that can be edited.


ryscher:
Works great. Thanks for fixing this.

PR #569, opened 2021-07-13
Separate curators from superusers

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1288

- Adds a new `curator` role
- Converts many `superuser` permissions, allowing them to be accessible by curators
- Tools that are primarily used by developers are now accessible only by `superuser`

### DIFF
--- a/documentation/apis/adding_api_accounts.md
+++ b/documentation/apis/adding_api_accounts.md
@@ -32,7 +32,9 @@ To create an API account:
    `select * from stash_engine_users where id= <user id>;`
 
 To set permissions for the API account:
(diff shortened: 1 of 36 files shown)


ryscher:
Note the change to upload_files_spec.rb -- This test originally passed, but then failed, even though it has nothing to do with permissions. So I had to change the test. It looks like the text of the error message was different than the test expects, and the duplicate file is now shown as "Pending" in the table. I'm not sure why this would have suddenly changed....

sfisher (approved):
Thanks Ryan!

I tested this out on the user model and looked through code and everything looks and works well.

Thanks for all the tests, also.

PR #570, opened 2021-07-14
Peer review and edit_url fixes

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1364 and https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1339

Several little fixes for curators and EM:
- When dataset status is updated by a journal metadata email, remove hold_for_peer_review (aka the PPR checkbox)
- When dataset is created by the EM API, default the PPR checkbox to "true"
- Ensure EM receives the sharing_link, regardless of the dataset's status. It won't return much until the dataset is actually submitted, but EM caches this URL early in their process, so they need it even with the initial API call
- When a user visits the EM edit_url, allow them to edit the dataset, regardless of its status, creating a new version if needed

### DIFF
--- a/spec/responses/stash_api/datasets_controller_spec.rb
+++ b/spec/responses/stash_api/datasets_controller_spec.rb
@@ -177,6 +177,7 @@ module StashApi
 
         expect(ident).to be
         expect(ident.publication_name).to eq(hsh[:journal_full_title])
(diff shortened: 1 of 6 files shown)


sfisher (approved):
Thanks for these changes, Ryan.

PR #571, opened 2021-07-14
Migrate over 50gb

sfisher:
Very small pull request:
- change queue feeder to > 50GB
- only add one of the super-monopolizer datasets when nothing else is running.  They take a worker forever, maybe up to 24 hours until they free that worker again.
- Timeout up to 24-hours for the overall worker (we have read and write timeouts on http requests that are much shorter and I hope we now catch all the common errors).
- Don't have delayed_job auto-retry stuff since we have lots of internal retrying already for requests and also we have a cron that retries daily.  Also doing some manual retrying.


ryscher:
Looks great. Glad they're all processing!

PR #572, opened 2021-07-15
Fixes two minor problems

sfisher:
- Adds alt text to logos on upload page
- Changes text from N/A to "Too Large For Validation.

See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1375


ryscher:
Thanks for fixing these!

PR #576, opened 2021-07-21
This sets the migration to run with the ui deploy

sfisher:
This keeps the application code and database schema in sync so everything works.


ryscher:
Thanks, hope this does the trick.

PR #577, opened 2021-07-22
Fixes webpacker rendering issues and missing css

sfisher:
It turns out the entire stage section was missing from `webpacker.yml` from when it got created.  😏 

I assumed it was OK since it was running and compiling assets in the deploy.  But no, instead of errors it just did something crazy.

Works now.

BTW, this is correct that the stage webpacker (ruby library) is using the production webpack (javascript).  The javascript webpack doesn't have a stage setting by default and uses production if the javascript environment for webpack doesn't exist.  It seems like this is default and the JS people like it this way. 🤷‍♂️


ryscher:
The new version of webpacker outputs some warnings about Babel settings, but it looks like everything still works fine.

PR #578, opened 2021-07-22
update documentation/membership/README.md with correct metadata URLs

ashleygould:
Members should use InCommon MDQ service to download dryad SP metadata.


sfisher (approved):
Thanks, Ashley!  Looks good.  Merging now.

PR #579, opened 2021-07-26
Fix nil errors if zenodo software/supp DOI hasn't been filled yet

sfisher:
This method may attempt an update for the software/supplemental relation before we've received and written the zenodo doi to our database.

So skip adding the relation unless the zenodo DOI is already present.


ryscher:
Looks good. Thanks for fixing!

PR #580, opened 2021-07-27
In EM API, only return the deposit_upload_url for "deposit" calls

ryscher:
EM sends two types of requests: "desposit" and "submission". The requests and responses for these requests are virtually identical, with the primary difference being that the "deposit" call typically includes less metadata. In Dryad, the two calls are treated the same. 

EM has recently had difficulty handling a response for "submission" calls when the response includes the field `deposit_upload_url`. This PR suppresses that field when the request is not a "deposit" request.


sfisher (approved):
Thanks, Ryan.  This looks good to me.

PR #581, opened 2021-07-27
Clarifications to documentation about GMail tokens

ryscher:
I re-ordered and re-worded a little bit, since it was confusing that the most common process was listed last.


sfisher (approved):
Thanks for this great documentation.

PR #582, opened 2021-07-28
Be more aggressive about updating zenodo related works

sfisher:
Even if it doesn't notice changes in the files going to zenodo, update anyway.

(Add 1st synchronization of related works much earlier in the process, with a couple of synchronizations later just to be sure.)


ryscher:
Looks great. Should be useful.

PR #585, opened 2021-07-29
Omit commas for authors without affiliations and funders without grant numbers

ryscher:


sfisher (approved):
Looks fine.  Thanks, Ryan.

PR #586, opened 2021-07-30
VPAT fixes for files upload

sfisher:
The fixes for https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1329

To test:  For the ARIA live region, right click a buttong on the bottom of the page and inspect it.  Find the aria-info id element above it and remove the css class `screen-reader-only`.  That will allow it to appear onscreen so you can see the messages.

You should see the 3 messages indicated in the ticket for those events.  Aria should read them out to screen readers because that is how aria-live regions work.

The others should be more obvious how to test.

### DIFF
--- a/app/javascript/components/UploadType/UploadType.js
+++ b/app/javascript/components/UploadType/UploadType.js
@@ -3,7 +3,6 @@ import React from 'react';
 class UploadType extends React.Component {
 
   componentDidMount() {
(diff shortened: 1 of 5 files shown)


ryscher:
Looks good, works in my testing. Thanks for researching a better way to do this!

PR #587, opened 2021-08-03
Moves the notifier into Rake

sfisher:
Sorry for so many commits.  This was a royal PITA to get running on the server because of libv8 (javascript) and bundler problems with the deploy.

It is running on the new dev with this branch.

- The script maintains state in the `shared/config/notifier_state.json` file which gets symlinked.
- Created code in the rake task which mostly is the same as the old `main.rb` file in the standalone script.
- Runs like `RAILS_ENV=development bundle exec rails notifier:execute`
- Fixed tests.
- Added missing file of `puppet-automation.rb` which made deploys fail in an unintelligible way when it was missing.

PS.  The "easy way" deploy script does something where it freaks out if there are merge conflicts rather than just doing the deploy of what you want.

Just edit it and replace the pull with these two lines:

```
git fetch origin $BRANCH
git reset --hard origin/$BRANCH
```

We can see if Ashley wants to modify the script this way and if it works for all types of deploys (branches/tags/references).


ryscher:
Looks good, works fine for me. Thanks for moving this. It will be easier to manage now.

PR #588, opened 2021-08-03
Uncouple "primary" article DOI from other related works

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1382

- Adds a new type of dcs_relation, the `primary_article`.
- Separates handling for two types of DOI fields: the `primary_article` at the top of the submission from, and other `article` relationships in the Related Works section.
- Simplifies field names in the top of the submission page. Instead of a field that incorrectly implies the DOI is stored as an internal_datum, use the name `primary_article_doi`.
- Removes the assumption throughout the code that the dataset is tightly tied to the first `article`. Instead, we use the most recent `primary_article`.
- Fixes an issue with importing metadata from CrossRef, where an incorrect ISSN could be saved.
- On landing pages, combines the display of `primary_article` and `article` relationships into a single "Article" section, with the `primary_article` values displayed first.

### DIFF
--- a/stash/stash_engine/ui-library/includes/import_metadata.html
+++ b/stash/stash_engine/ui-library/includes/import_metadata.html
@@ -32,19 +32,19 @@
   <p>Please provide the following information. You may either enter the information and leave it or choose to autofill
     your dataset based on you supply below.</p>
 
-  <form class="edit_internal_datum" id="edit_internal_datum_157"><input name="utf8" type="hidden" value="✓">
+  <form class="edit_157" id="edit_157"><input name="utf8" type="hidden" value="✓">
     <div class="c-input__inline">
       <div class="c-input">
-        <label class="c-input__label" for="internal_datum_publication_name">Journal Name</label>
-        <input value="" class="js-publications c-input__text ui-autocomplete-input" type="text" name="internal_datum[publication_name]" id="internal_datum_publication_name">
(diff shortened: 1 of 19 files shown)


stash/stash_engine/ui-library/includes/import_metadata.html
```diff
@@ -32,19 +32,19 @@ <h3 class="o-heading__page-span">Preliminary Information</h3>
   <p>Please provide the following information. You may either enter the information and leave it or choose to autofill
     your dataset based on you supply below.</p>
 
-  <form class="edit_internal_datum" id="edit_internal_datum_157"><input name="utf8" type="hidden" value="✓">
+  <form class="edit_157" id="edit_157"><input name="utf8" type="hidden" value="✓">
```
sfisher, line 35: Hah.   I wonder where that 157 came from.  Kind of random.  :-)

sfisher (commented):
This works, but I did notice one weird thing.  When I type it in for finding my article the button changes back and forth from "Import Manuscript Metadata" to "Import Article Metadata" while I type and tab through the form.

I think the first time I tried it, I clicked on the import button while it still said "Import Manuscript Metadata" and I don't think it worked.

When I tried it with the other button then it worked fine.

IDK if this is critical or not.

PR #590, opened 2021-08-03
Protect EM API against missing authors

ryscher:
This *should* never happen in a real API call, but I was able to create an error with a contrived test, so I wanted to protect against it.


sfisher (approved):
Thanks, Ryan.  I'll tag and redeploy to dev and stage again.

PR #591, opened 2021-08-04
Add "at a glance" stats to curation stats

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1384

To accomplish this, I adapted and simplified the mega query from the Admin Dashboard. Given this duplication, I'm starting to contemplate pulling all of the admin queries out into a separate module that just builds queries for various purposes. But that's a larger task that I don't want to tackle just yet...


sfisher (approved):
Thanks for adding this, Ryan.

PR #592, opened 2021-08-05
Counter on new servers

sfisher:
This builds on the previous PR since I needed that code to start on this one.

See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1416 for what this involved (including some manual changes to authorized keys and some other things on the servers).

To test, you can go to the *new* stage server and run:

```
cd ~/apps/ui/current/cron
$ RAILS_ENV=stage ./counter.sh
```

It currently runs, but Ashley hasn't gotten the second server in the rotation working yet and so it doesn't ssh the log files over and combine them.  However it tries, but we should try again once she has done that.


ryscher:
Thanks, this runs for me on new stage, and everything looks fine.

PR #593, opened 2021-08-06
Update statuses for EM API

ryscher:
Use different statuses, based on feedback from Daniella & Jess.


sfisher (approved):
Looks good.  Thanks, Ryan.

PR #594, opened 2021-08-06
Update the task that ensures reports sent to datacite monthly runs within rake

sfisher:
I essentially moved the code into a new rake task but made some changes to paths and also how it gets the secret key for uploading to DataCite.  Also updated docs a bit.

To test it out

```
# ssh to the new server.  I tested on the first stage server, but you can try wherever.
# The main thing is to set the path to the json report files.  You could copy a file over from
# a production machine /apps/dryad-prd-shared/json-reports if you want one to force submit

cd ~/bin
./deploy_dryad.sh counter-monthly-check-rake

cd ~/apps/ui/current/

# You could try development or stage here, but I think they both use the same DataCite sandbox for testing.
# Be sure the force submission file exists if you want to force an upload of that report.
RAILS_ENV=stage REPORT_DIR="/apps/dryad/apps/ui/shared/cron/counter-json" FORCE_SUBMISSION="2021-08" bundle exec rails counter:datacite_pusher
```


ryscher:
Looks good and submits for me. Thanks for dealing with all the crazy counter stuff!

PR #595, opened 2021-08-06
Updates to Zenodo docs

sfisher:
Should be more up to date and have better troubleshooting information.

PS.  Removed the section about ops server (now back on confluence).  While nothing really critical, Ashley didn't like seeing it in our repo and github.


ryscher:
Thanks for updating this.

PR #596, opened 2021-08-06
How to unpublish, instructions

sfisher:
I added specific instructions for "unpublishing" since this is coming up semi-regularly recently.

There is really no easy way to completely take back a publication and then re-publish it again later with all the external services involved and permanent identifiers that get set up.

This should mostly do the job, I think, but it seems like we may need some way to prevent this from happening so often since it affects external systems like Zenodo and DataCite that expect publishing to be a pretty permanent step that isn't easily recalled.

It probably should re-confirm or make people go through an extra step that the other states don't have so that it doesn't happen on accident so often.


ryscher:
Thanks for updating these instructions.

PR #597, opened 2021-08-09
Adding a bunch of role="button" for accessibility

sfisher:
where ticket indicated button functionality and they were links. It would involve much larger layout changes to make these into actual buttons since they'd have to be moved outside of forms and also would take significantly more room in the layout and change the visual look which was designed for us.

We think this addresses the accessibility problems without a huge rework, but we can evaluate at the aria review.


ryscher:
Looks good, should address the issues.

PR #598, opened 2021-08-09
Transforming long read out links into text 

sfisher:
and reformatting some of the longer and more annoying cases where huge paragraphs were all on one line.

see #1322 https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1322 .


ryscher:
Thanks, these links were always excessive.

PR #599, opened 2021-08-09
Allow curators to be unassigned from a dataset

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1404

On the Admin Dashboard, the pencil next to a dataset's curator may now be used to unassign the curator. When a curator is unassigned, the dataset is forced into curation status 'submitted', so other curators will see it and can take it over.


sfisher (approved):
Thanks, Ryan.

Looks good to me and I tested it and worked correctly.

PR #600, opened 2021-08-10
Fixes for ckeditor and also associating labels with fields on metadata page

sfisher:
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1324

If you look at CKEditor, then adding note below all the places it's used to note people can press ALT-0 for help screen.  Also modifies the text that is associated with the editor and screen readers to have correct name of field and to repeat they can press ALT-0 for help with controls.  Had to do this with trickery since it's inside external libraries (javascript changes it after loading).

Fixing and adding labels.  You can test most labels to see if they're correctly associated by clicking on the label and it automatically goes to the form element.

There were some wonky things in there that got fixed around the top form for choosing manuscript and some other areas not associated correctly on multiple elements (first ones worked but labels were not specific to multiple copies of the forms).


ryscher:
Looks good, all useful updates.

PR #602, opened 2021-08-11
3 smaller fixes

sfisher:
1. Fixing email address so it shows up correctly for contact in the "Keep my dataset private" area of review page.
2. Changing `<i></i>` to `<em></em>` if that makes someone happy.  Really no effect since it's just something for bootstrap icons and doesn't show or have any effect, anyway.  That's how bootstrap rolls.
3. Futzing with the software license and just making it a standard select list for now since apparently the other one isn't accessible.  Tried with the html 5 datalist which gets recommended by others, but their use case is different since no easy way to lock to only the options in the list with the datalist and people say to just use the standard select list.
  - We can revisit later and spend a bunch of time in javascript to make the datalist lock to only values in the list, but not going to to it now since I don't have time.

### DIFF
--- a/config/app_config.yml
+++ b/config/app_config.yml
@@ -14,7 +14,7 @@ defaults: &DEFAULTS
     host: ezid.cdlib.org
     port: 443
   google_maps_api_key: <%= Rails.application.credentials[Rails.env.to_sym][:google_maps_api_key] %>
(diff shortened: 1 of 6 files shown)


ryscher:
Looks good, thanks for squeezing these in.

PR #605, opened 2021-08-17
 Funder dashboard

ryscher:
For: https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1402

Allow users to be assigned a role as administrator of a funder. Users with this assignment may view the Admin Dashboard, and see datasets there that are associated with the funder.

This does not add an explicit model object for funders. Instead, the relationships are tracked using FundRef IDs. The IDs are stored in `dcs_contributors` for the resources, and `stash_engine_funder_roles` for the users.

To test:
- identify a user that has only basic "user" permissions
- create a FunderRole associated with the user, like `fr = StashEngine::FunderRole.new(user: u, funder_id: 'http://dx.doi.org/10.13039/100000001', funder_name: 'NSF', role: 'admin')`
- login as this user
- view the Admin Dashboard, and verify that all of the visible datasets are associated with the funder


ryscher:
Merging for deploy. @sfisher can review later...

PR #611, opened 2021-08-27
Lower timeout for requests to Merritt

sfisher:
for creating presigned downloads or packaging requests

This will make it less likely that our request queue for passenger gets > 100 which knocks our service out since we're not waiting 30 seconds or 60 seconds for Merritt to respond with info about a download.  Terry said 5 seconds was enough time for these requests to be fulfilled normally unless something bad is going on.


ryscher:
Looks good. Thanks for improving this.

PR #612, opened 2021-08-27
PR to merge main-new-server into main

sfisher:
This merges the stuff we have for the new server into the main branch.  These PRs should've been checked before when they went into that branch.  I'm just trying to be sure it merges and deploys cleanly on the new servers.

To test it out, I made another branch that is a copy of main and tried the merge manually.  It is called `test-merge-new-server` and I also made a tag of it `v0.6.07test005`.

I tried the manual deploy on development and it seems to work.  I'm resetting the stage server to use the more complicated deploy and trying to be sure puppet deploys it correctly from that tag.

It didn't deploy at first until I made a change to babel config that gets called with the webpack configuration.


ryscher:
All looks fine.

PR #614, opened 2021-08-31
Rate limiting with rack-attack

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1192

Adds the rack-attack rate limiter and implements some reasonable rate limits. These rates are currently per-server, so they will effectively be doubled on the load-balanced servers.

To test:
- Run the server with the environment 'test'. This will use the lower rate settings.
- Load a page via browser or curl, and reload it more frequently than the limits specified within `dryad-config-example/app_config.yml`
- Check the log file for a message indicating the request was blocked
- The rates will reset after 1 minute of no activity

Thoughts for future enhancements:
- Instead of IP for API users, lookup their ID number from the DB oauth_access_tokens, so we can aggregate across their requests
- Have broader throttle limits? (per hour, per day)
- Send email when requests are blocked? For now, I'm keeping messages in the log, until we see how often requests are actually blocked.


sfisher (approved):
Thank you, for doing this, Ryan.

These are good changes and will help us out for problems like we just had.  Very useful, even if it doesn't happen too often.

PR #615, opened 2021-08-31
Stop annoying messages from geoblacklight

sfisher:
The version of blacklight that is known to be compatible with geoblacklight 1.9,.1 is blacklight 6.11.

We had some later versions in there somewhere and also bundler it kept upgrading it.  For now put it at this version to prevent lots of error messages.  Verified that they stopped after deploy to stage.

Also tested search and it works fine and doesn't break anything I could find.


ryscher:
Thanks for tracking this down. As far as I can tell, search still works in both UI and API.

PR #616, opened 2021-08-31
Convert response tests to request tests

ryscher:
They were always tests of type "request". I don't remember why we had them in a directory called "response".


sfisher (approved):
Looks fine, thanks.

PR #617, opened 2021-08-31
New mexico state university

sfisher:
This finally gets the correct entity id for NMSU and tested on stage and it worked.


ryscher:
Thanks, looks good, tests ok on stage. When merged into main, the tests will pass again.

PR #618, opened 2021-09-01
Automatically close page for EM API

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1440

When users make a deposit through the Editorial Manager API, they are redirected to an `edit_url`. This PR adds a parameter to the `edit_url` so the page will close at the end of the submission process.

Unfortunately, the JavaScript for closing a page does not work consistently in all browsers. For browsers where it does not work, the page text simply instructs the user to close the page themselves.


sfisher (approved):
Thanks, Ryan.

This looks good to me.  I wasn't sure if you needed to encode any of the URLs in order to put them in the querystring (or maybe already are).  Feel free to merge if you think it's OK as is and doesn't need encoding.  IDK full background.

PR #619, opened 2021-09-02
Remove outdated email templates

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1412

Removing email templates that have not been used since 2019. They have been sitting in the directory and causing confusion whenever someone goes to update the language in all of the templates.


sfisher (approved):
Thanks for the cleanup.

PR #620, opened 2021-09-02
Fix api version downloads

sfisher:
- Makes the code for download dataset and download version common
- Fixes tests that were there for security checks and download
- Adds tests for redirect statuses and info

This turned into a huge pain for tests because of some hidden callbacks and validations on the curation activities that interfere with the resource class saving and operating independently. (User 0 didn't exist in the database so updates to resource didn't work because curation activities were being added when they weren't useful and validation failed because of missing user which canceled updates on resources without warning.)

I'm starting to be of the opinion that these kinds of side-effect callbacks on related things are something of an anti-pattern unless they really need to be applied every single time something happens and in a deep way across all models all the time.  It took forever to trace and figure out and the deep side effects weren't showing why things failed on unrelated models until I dug for a while.

### DIFF
--- a/spec/requests/stash_api/versions_controller_spec.rb
+++ b/spec/requests/stash_api/versions_controller_spec.rb
@@ -247,7 +247,7 @@ module StashApi
       describe 'permissions' do
         before(:each) do
           allow_any_instance_of(Stash::Download::VersionPresigned).to receive(:download)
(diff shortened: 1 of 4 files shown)


ryscher:
This works for me. Thanks for fixing all the tests. Sorry the system user gave you so much trouble.

PR #622, opened 2021-09-02
Ignore rack attack limits for presigning chunk uploads for files

sfisher:
Otherwise, dropping a large number of files at once makes some stall and not upload.

This is an easy change to rack::attack.

I tested on my dev machine with the huge group of files I used to find problem on production.

I can retag our release with this incorporated and test on dev/stage and redeploy to production.


ryscher:
Looks good. Thanks for testing this out.

PR #624, opened 2021-09-07
Allow journal links to work

ryscher:
Many journals direct their authors to submit data using a link like:
http://datadryad.org/submit?journalID=ECE3&manu=ECE-2021-04-12345.R1

We previously had a redirect in Apache to handle these URLs, but the redirect didn't carry over to the new servers. I'm adding it to Rails so we have direct control over it for the future.


sfisher (approved):
Thanks, Ryan.

I tested it out and it works great.

PR #625, opened 2021-09-07
Adding source for pyenv for counter script so that pyenv loads for cron

sfisher:
Pyenv wasn't loading correctly under cron environment and was using the wrong version of python when it ran weekly.

I tested this manually according to the info at https://stackoverflow.com/questions/2135478/how-to-simulate-the-environment-cron-executes-a-script-with   .  I tried getting the environment and going into a shell with it and then sourcing `~/.profile.d/pyenv`.  It then shows the correct version of python (3.7.9) rather than python 2.


ryscher:
Glad you were able to locate a simple fix!

PR #626, opened 2021-09-07
Document more details about finding v1 files

ryscher:
This will hopefully will never be needed again, but here is some more detail about finding the v1 files.


sfisher (approved):
Thanks for adding this, Ryan.

PR #627, opened 2021-09-10
Improve logic for determining the depositing_user in EM API

ryscher:
When a user account in EM has no ORCID, EM still sends an `orcid` field in their metadata, but sets it to an empty string. This empty string was incorrectly being used to initialize user accounts. 

The new logic avoids creating a user account if the ORCID is blank.


sfisher (approved):
THanks for finding and making these changes, Ryan.

PR #628, opened 2021-09-10
Force irb to be part of the environment

ryscher:
irb exists in most Rails environments, but we need to explicitly include it due to the setup of the new servers that Dryad uses.

With this change, you should be able to enter a Rails console using a command like:
```
cd /apps/dryad/apps/ui/current/
RAILS_ENV=development bundle exec rails console
```


sfisher (approved):
yay.  Working console/irb.

PR #629, opened 2021-09-10
Replace confirm buttons

sfisher:
This gets into replacing a standard function in the rails jquery-ujs library that does the "Are you sure?" dialogs everywhere.  It uses the standard javascript confirm dialogs that apparently the usability people hate and are confused by.  It replaces it with the standard HTML 5.2 dialog functionality and includes a polyfill for firefox/safari and some apple mobile browsers.

Sorry this took a long time to figure out.  The javascript confirm functionality is synchronous and the code waited for a response before continuing.  However standard html/javascript is asyncronous and so it continues executing and many changes didn't work.

There were some examples online for doing this but they weren't very good, left out some of the things in the function being replaced, used global variables and other badness.  They had examples with 3rd party dialogs like bootstrap which I think the accessibility people would complain about, even though it's commonly used.  Hopefully they can accept an HTML standard and seemed ok with it in our file upload url dialog.

I ended up having to call the function twice because of the async nature.  The first time pops the dialog and if they say yes, it sets a flag on the item in the page, re-activates the action again but this time doesn't pop the dialog and just takes the action based on the flag.

You can test on the delete dataset action on the "my datasets" page, but it is triggered some other places, also. Basically search *.html.erb for "confirm" on a bunch of rails elements and you'll get the list of places it's used.


ryscher:
Looks good, and tests fine for me. The only concern I had was adding some styles to look more like the rest of the site, but that can be fixed in the future. Thanks for working through all the crazy interactions!

PR #630, opened 2021-09-14
Replace all other modals with the html 5 modal and change headings, etc

sfisher:
This changes all the modals to the html 5.2 modal and makes them consistent and headings and etc as the accessibility people wanted.

I believe this is the complete list of the ones that are changed.

- upload files if file not uploaded before clicking "proceed to review" or navigating.
- Review page if no files in dataset
- http://localhost:3000/stash/admin if changing user role
- Curation Dashboard if clicking the stats icon
- Curation Dashboard -- change curation status
- Curation Dashboard -- edit curator assignment
- Curation -- Activty Log -- Add Note
- Curation -- Activity Log -- Add Data
- Download large dataset and assembly
  example: http://localhost:3000/stash/dataset/doi:10.5072/FK2X92972F
  - Reset it by deleting the row in the stash_engine_download_tokens table for the resource id (which can be seen in the form around the download button)
  - Test that canceling works (both close buttons)
  - (Implement dialog saying it's downloading)
- Publication Updater -- Accepting changes
- Landing page -- see citations such as http://localhost:3000/stash/dataset/doi:10.5061/dryad.ct40nm7
- ORCID invite acceptance on landing page (create a share and use it)
  - create author with email at mailinator and publish dataset
  - Follow link and add ORCID


ryscher:
All looks good. Thanks for tracking all of these down.

PR #631, opened 2021-09-14
Don't allow "reverting" datasets never submitted

sfisher:
I believe removing the dataset staging version under editing should be disallowed under circumstances where no version of it was ever submitted.

We may want to give curators a way to return control back to the user instead (aside from submitting their changes, which does that), but at least for now we can prevent them from destroying completely unsubmitted data with dangerous actions.


ryscher:
Thanks, Scott. This is a great fix to the problem.

PR #633, opened 2021-09-15
Script to download last submitted version of a dataset

sfisher:
@ryscher,

This will download the files from the latest version of a dataset.  No waiting for a zip to be assembled and no waiting while a file unzips.  It just downloads the files directly.

See the readme for more info what what needs to be filled in.  It may help with the curator environment.

It just runs as a standalone ruby script and you'd just copy that directory wherever, be sure ruby is installed and bundle install.

I didn't really add rspec tests or anything else since this is experimental, but if we think it's going to be used and needs to go through more testing and validation then we can do it at that point.


ryscher:
Thanks, Scott! This is great. Should make it much easier for the curators.

PR #634, opened 2021-09-15
Small accessibility cleanup

sfisher:
Added a few minor accessibility fixes.  We are using the same display in the review as the landing page for some items.  The headings are different in the two pages even though the display of the items doesn't really need to change.


ryscher:

PR #635, opened 2021-09-16
Allow page size selection in all API calls

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1101

- Fixes the `/datasets` call so the `page_size` is actually applied as the documentation specifies.
- Adds `page_size` to calls for `files` and `versions`


sfisher (approved):
Thanks for modifying these, Ryan.

Looks good and will be useful to page.

PR #636, opened 2021-09-20
Salesforce connection -- linking to case numbers

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1355

- Adds a connection to the Salesforce API.  
- When a curation note mentions a Salesforce case number, the Activity Log page renders a link that allows viewing the case
- Adds a section to the Activity Log page that displays Salesforce cases associated with the dataset's DOI

Note that in the Notes fields, links will only render if there is a valid case in Salesforce that corresponds to the given case number. Some examples:
- SF 00006419
- SF 00006277


sfisher (approved):
Thanks you, Ryan.

I tested and it works great.  Great use of helpers to add/format the link.

PR #637, opened 2021-09-25
Upgrades for Geoblacklight 3.4

sfisher:
There are a lot of gem updates in here and there were a lot of conflicts.  I tried to clean it up and only require things directly used in an engine in the gemspec and remove most of the development dependencies since we're developing and testing within the main app now.

Mostly this re-integrates geoblacklight again with all the changes they made.  Different configs, some changed code, some different styles and changed javascript and libraries.

I tried to simplify as much as I could and depend on the their original files as much as I could.  I ended up removing a number of views and other things that we had overridden before.

I also had to completely separate geoblacklight into a different layout instead of using our main layout with some if statements.  I tried to put most of the things in our layout into partials and each layout calls the pieces it needs to reproduce the Dryad layout around the outside of the page.

This was mostly a long nightmare because of all the dependencies, conflicts and hard to find bugs and problems. They really want you to base your whole app on the geoblacklight framework and so integrating it with our existing one that does more than search was a bit problematic.

A lot of the styles are overridden in the css by making them more specific than bootstrap and replicating many things again from our main stylesheets.  Unfortunately, bootstrap is required for geoblacklight, but it really tries to clobber everyone else and doesn't play nice.

Test and be sure it seems ok.  I've tried submissions and publishing and I think our workflow is intact with all the upgraded gems and tests pass.


stash/stash_discovery/config/routes.rb
```diff
@@ -1,61 +1,26 @@
...
-require 'geoblacklight'
-
 Rails.application.routes.draw do
-
   get '/latest', to: 'latest#index', as: 'latest_index'
   # blacklight_for :catalog
 
   # Endpoint for LinkOut
-  get :discover, to: 'catalog#discover'
+  # get :discover, to: 'catalog#discover'
```
ryscher, line 6: This route is still needed for LinkOut. I tested uncommenting it, and it works properly.

ryscher:
Thanks for working through these issues, Scott. All looks good and tests fine, except for the LinkOut route that got commented out.

PR #638, opened 2021-09-27
Fixing orcid co-author emails to have the correct name.

sfisher:
This had "assign_variables" for a resource which assigns based on the user and resource.  Instead it needed to have the `@user_name` set based on the invitation, not the user of the resource.


ryscher:
Thanks Scott, this works well.

PR #639, opened 2021-09-27
Salesforce -- create cases

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1354

- Separates production and non-production Salesforce activity into different sites.
- Improves formatting of cases shown on the Activity Log page
- Adds a button to the Activity Log that will create a new case with associated metadata

Note that this PR does not include the AJAXy update of the Activity Log page, so once a case is created, the Activity Log must be refreshed before the case will appear.


sfisher (approved):
Wow.  This is great.  It's awesome to see the SalesForce case get created so easily.

PR #641, opened 2021-09-30
Also link salesforce cases where leading zeroes were omitted

ryscher:
When I originally checked the curation activities to see the formats being used for Salesforce tickets, I thought all of the references used the full case number. However, in some recent datasets, I see the leading zeroes omitted. This will allow them to display as links.


sfisher (approved):
Thanks Ryan.  Looks good.

PR #642, opened 2021-09-30
Add notes about curator pc

ryscher:


sfisher (approved):
Thanks for documenting this, Ryan.

PR #643, opened 2021-09-30
Add index to user.orcid and also general database and code cleanup

sfisher:
- Add index to user.orcid
- Drop old file tables stash_engine_file_uploads and stash_engine_software_uploads
- Drop download history table we used for tracking when we mediated all downloads and streamed through our application directly.
- Removing the code for streaming downloads through our application directly.  It's all presigned urls now.
- add resource_id index on stash_engine_resource_states
- .gitignore all `Gemfile.lock`s on engines/gems embedded within our application.  The gem dependencies are already locked within our main application Gemfile.lock and we're doing testing within the main application, also.
  - They are also constrained to a range by the .gemspecs and may also be constrained by the Gemfile within a gem.
  - The `Gemfile.lock`s mostly just make bundler freak out when we need to update gems for security or upgrades like Geoblacklight and if stuff isn't exactly right then it will not find a way to resolve problems when it's locked so tightly and another is locked tightly to some other version and you end up in dependency hell.


ryscher:

PR #645, opened 2021-10-01
Update reporting docs for new servers

ryscher:


sfisher (approved):
Thanks for the instruction update.

PR #646, opened 2021-10-01
Prevent multiple copy records for the same resource

sfisher:
We had a couple of weird, rare failures where we had more than one copy record for replicating data to zenodo for the same resource.

This prevents another entry from being created if one exists and also prevents starting replication if it's anything other than enqueued state or it just created the record (as enqueued).

I thought maybe there was something in submitting that allowed multiple submissions for the same resource, but I think this happens from curators publishing, so probably it happened when curators somehow clicked multiple times in rapid succession to publish an item somehow.  I guess we should probably have the publishing step block them with a spinner to only allow clicking once until a response to the request is received.

### DIFF
--- a/spec/models/stash_engine/resources_spec.rb
+++ b/spec/models/stash_engine/resources_spec.rb
@@ -1485,6 +1485,12 @@ module StashEngine
         @resource.send_to_zenodo
       end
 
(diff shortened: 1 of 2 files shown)


ryscher:
Looks good. Hope it fixes the issue.

PR #648, opened 2021-10-04
Very small changes to dryad emails so they read better #1471

sfisher:
Fixes missing end tag for list item, period other small things.


ryscher:
Thanks for cleaning these up.

PR #651, opened 2021-10-04
Add ISSN to API requests that only contain journal title

ryscher:
When API users only send journal title, locate the ISSN if it is available.


sfisher (approved):
Thanks for doing this.  It's annoying that they don't give an issn.

PR #652, opened 2021-10-04
Explicit destroy of the frictionless reports before removing the file

sfisher:
A user couldn't delete because there were two copies of frictionless reports for a CSV.  The destroy_all failed because of constraints.  IDK how they got two reports, but, added an explicit destroy of frictionless reports before destroying a file to prevent this issue.

Also put in the change we've been running with for a while anyway to limit size for CSV checking to 5MB since some of the larger ones with very odd formats cause frictionless to take minutes to process.


ryscher:
Looks good. One more step towards making it bulletproof.

PR #656, opened 2021-10-06
Improve logic for emails in Salesforce cases

ryscher:
Curators complained about email addresses not being included in the salesforce tickets. This is due to the submitter (resource.user) not having an email address present.

This change updates the logic in the Salesforce module to be parallel to the logic used in UserMailer, so it will use the first author's information if that is available. It is a slight duplication of code, but I didn't think it was enough to warrant moving everything into a separate helper class.


sfisher (approved):
Looks robust!

PR #657, opened 2021-10-07
Fixes Datacite metadata generation for funder ids

sfisher:
Generate funder ids for fundref if they are set in our database for a contributor of type funder with an id.


ryscher:
Looks good, works in my testing. Thanks for adding this. I know it's a pain to work through that Datacite mapping code.

PR #670, opened 2021-10-19
Added additional output so we can see Frictionless times to complete validation

sfisher:
This is a minor change that outputs completion times in seconds to the log for frictionless validation.


ryscher:
Looks straightforward.

PR #686, opened 2021-11-01
Improve logic for curator assignment

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1472

- Allow auto-assignment of curators in more situations, so all resubmissions are assigned to a curator
- When a curator is no longer with Dryad, choose an appropriate backup curator.


sfisher (approved):
This looks pretty good and thanks for all the tests!

PR #687, opened 2021-11-01
Add datacite contributor identifier_type

sfisher:
This adds the identifier_type specifically to the contributors table.  Before everything in contributors was assumed to be a funder.

Also adds both identifierType and identifier to the API.

Fixes the UI where it it submitted so it has correct identifier type as part of the form.

Added tests. examples. migrations.

It's optional, so if people use the API the old way then these aren't set and they are not required so shouldn't break compatibility.

### DIFF
--- a/documentation/apis/sample_dataset.json
+++ b/documentation/apis/sample_dataset.json
@@ -24,8 +24,10 @@
     } ],
     "funders": [ 
 	{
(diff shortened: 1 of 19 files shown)


ryscher:
Looks good, tests fine for me. Thanks for adding this.

PR #689, opened 2021-11-03
Improve robustness of Salesforce connection

ryscher:
This should minimize the impact of issues like https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1513

- update password, since Salesforce required a change
- catch exceptions when initializing Salesforce connection


sfisher (approved):
Thanks for fixing this, Ryan.

Do we need to do anything to redeploy this or maybe you've already done it.

Catching StandardError is a bit broad, but it might not be clear what typical errors the salesforce gem gives.  Maybe we can revisit later to catch http or password type errors if they bubble up that way.

PR #690, opened 2021-11-03
Duplicate submission warnings

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1351

When an author submits a new dataset (new `Identifier`), if that dataset has a title substantially similar to the prior dataset, flag it as a possible duplicate. A warning is shown on the Review page of the submission system, though it can be ignored. If the dataset is submitted anyway, a warning is added to the curation activities.


sfisher (approved):
Thanks for doing this, Ryan.

I tested and it works great with either or both problems (duplicate or no data files).  Thanks for the tests.

PR #692, opened 2021-11-04
Adds the research facility using a ROR and autocomplete

sfisher:
Ok.  I added the extra items for this PR and it should be OK now.

I also had to incorporate the Sheffield tenant in here because we need both of these things on Stage before release.

- The facility form and the javascript is the main code here.  It got a bit tricky to get it to reset the ROR id in the hidden field correctly.  It sets it when an item selected from the list.  However it should remove it again if changes are made to the text input field.  I finally settled on detecting typing of printable characters or the delete key in the input to reset it on those changes.  I tested a bunch to be sure the hidden ROR id resets at the right times and I think it's pretty bulletproof.
- Also changes some of the javasascript attached to the controller action since it now has to handle two separate forms (funder and sponsor). That js changes the form from a POST to a PUT after initial creation and to do updates after that.
- Adds tests for filling in the UI.


ryscher:
Thanks for adding this.

I think it's good enough to merge, but there's one tiny bug that could be fixed by modeling the JavaScript more on the format that is used for author affiliations:
- Fill in a partial facility name
- Click on one of the results
- Without clicking on another field or tabbing out, switch the focus to another application using a keyboard shortcut (Cmd+Tab, or Alt+Tab)
- The name is incorrectly saved in the DB without the ROR.

sfisher:
BTW, @ryscher This functionality you note above seems pretty much the same as the Author affiliations which I based my code on.  It also wasn't correctly resetting the ROR if someone went back and typed some more onto the affiliation name there (which I can't reproduce now).

In any case, the author affiliation often always save correctly, either, if it is tabbed to another window before exiting, so I don't think that problem is fixed there, either.  It sometimes saves the ROR and sometimes doesn't.

You can watch it with a query like this (fill in the resource id) at the end.

```
SELECT auth.`author_first_name`, auth.`author_last_name`, af.long_name, af.ror_id FROM stash_engine_authors auth
LEFT JOIN dcs_affiliations_authors aa
ON auth.id = aa.`author_id`
LEFT JOIN dcs_affiliations af
ON aa.`affiliation_id` = af.`id`
WHERE resource_id = 3530;
```

It seems like it *often* saves the author ror if it is an existing record that is CMD-tabbed out of (but not always).  It *usually doesn't* save the ror if it is CMD-tabbed out of in a new record (click the + sign for a new author and do the affiliation on that one).

For the research facility, if *usually doesn't* save the ror if CMD-tabbed in a new record.  It *usually* saves the ror for an existing record being updated.

The author ror seems like it might be *slightly* more consistent, but it also seems to give unpredictable results.  So if it's critical that these are fixed we probably need to look at the code for both places since neither are really 100% reliable for this scenario.  I tested this on the dev server.

PR #693, opened 2021-11-05
Pulls in extra funding info we got from Ted

sfisher:
We got a spreadsheet from Ted with additional funders for some datasets.  These were to be added to the datasets.

Some had IDs and some didn't.

- If they have an ID filled by Ted then use that ID.
- If no filled ID, then do an exact match against our Fundref dump of names/ids.  If still no ID, try to find any records in our database with that name string for funder and an ID filled in for it.
- Create the funder record in our database.  We hope it has an ID, but if it doesn't we add it without one.

It seemed like only a small percentage of items without IDs that Ted gave us were direct matches against fundref names or something already in our database with an ID.  I'd guess maybe 10-20% at most.

If someone decides they want to do some more manual matching I can export the unmatched items again as a spreadsheet.  Last time we did this, I went through the top 250 or so items and did some manual lookups on things I could easily figure out and it took me a few hours.  Trying to go through all of them will be tedious and lots of time and work.

IDK if we'll ever get all of them to match because the fundref database is limited and people put lots of small funders that may not be in the fundref database.  Also, there seem to be lots of english-version names for funder names in other languages which are a little hard to figure out sometimes without a lot of googling or speaking those languages better.  Also, lots of acronyms (some from foreign languages/govts).

### DIFF
--- a/stash/stash_engine/lib/tasks/funders.rake
+++ b/stash/stash_engine/lib/tasks/funders.rake
@@ -86,30 +86,35 @@ namespace :funders do
 
   desc 'imports Tedsheet which is something someone gave us with IDs for some funders'
   task import_tedsheet: :environment do
+    util = Funders::Utils.new
+
     CSV.foreach('/Users/sfisher/Downloads/funderOutputCrossrefFixed__20211006.csv', headers: true) do |row|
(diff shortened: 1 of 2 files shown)


ryscher:
Thanks for archiving these scripts.

PR #694, opened 2021-11-05
Recalculate payment when journal changes

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1146

Not a lot of code changes, but some subtle interactions here...
- When journal information is changed, and the journal was set to pay, the payment information will be recalculated.
- Payment is *not* recalculated for other types of payment (e.g., if we manually apply a waiver)
- Payment is *not* recalculated for datasets that are already in the system at the time a journal discontinues its payment plan. As long as the journal associated with the dataset remains consistent, we don't change it out from under the depositor. We only change it when the depositor actively switches journals.
- This PR also removes a stub feature for invoicing journals through Stripe, which we never used.

### DIFF
--- a/spec/models/stash_engine/identifier_spec.rb
+++ b/spec/models/stash_engine/identifier_spec.rb
@@ -38,8 +38,8 @@ module StashEngine
       end
 
       @fake_issn = 'bogus-issn-value'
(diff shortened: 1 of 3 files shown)


sfisher (approved):
Thanks for updating this, Ryan.

PR #695, opened 2021-11-08
Optimize curation so that it loads only results needed for display

sfisher:
It looked like this originally loaded all results from the database (about 66,000) and then dumped most of them out of the array to display 10 items in Kaminari (pager).

I had to point my development against production and test a bit against a bigger database to see the slow times more clearly.  (62 seconds to load the first time and then maybe some caching or something.)  After optimizing with LIMIT then that query was about 500 milliseconds.

- Only returns the results from SQL needed to get current page and a few pages after (for making the pager look better).
- No longer has a "Last page" just like there isn't a "Last page" of Google. I don't think there is any compelling reason to rerun a full search with COUNT in order to get an exact last page number.  If people want the end, they can sort something in the reverse direction or limit or search in some smarter way.
- Had to hack in a bunch of nil values onto the front of the Array since Kaminari is kind of annoying and expects them.  I looked for workarounds for a few hours and didn't find a lot else that was a really better solution.  BTW, github has tickets complaining about Kaminari and some slow queries in their codebase because people used it badly there, so we're not the only ones and they wanted Rubocop warnings about it so people didn't use it for paging on some complex SQL queries. ;-)
- Changed the menu item for dataset curation to limit to the current user and submitted status since that is what they generally start from.
- Changed the curator list to display as `lastname, firstname` since it sorts by lastname and without it displayed that way it made it look like it was sorting in a random order.
- Fixed a bunch of tests (mostly caused by the new filter when navigating with the menus).
- Made sure it does the full query for CSV since that shouldn't be paged and dumps everything.

### DIFF
--- a/spec/features/stash_datacite/dataset_versioning_spec.rb
+++ b/spec/features/stash_datacite/dataset_versioning_spec.rb
@@ -105,7 +105,7 @@ RSpec.feature 'DatasetVersioning', type: :feature do
         before(:each, js: true) do
           sign_in(@curator)
           find('summary', text: 'Admin').click
-          click_link 'Dataset Curation'
+          visit stash_url_helpers.ds_admin_path
         end
 
         it 'displays the proper information on the Admin page', js: true do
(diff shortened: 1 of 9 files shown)


ryscher:
Thanks Scott! The limit made it much faster. Too bad Kaminari still has to have the full array.

PR #697, opened 2021-11-10
Do not auto-assign curator if dataset was never in curation

ryscher:
Fix for issue reported by curators. When a submitter versions a dataset before it reaches curation, we do *not* want to auto-assign a curator.


sfisher (approved):
This looks good, Ryan.  I'll merge and retag and deploy to dev/stage for inspection before doing it on production.

PR #698, opened 2021-11-12
Perhaps fixes double-downloads by curators

sfisher:
This cleans up a bunch of things that might have some effect on double downloads that curators mentioned happens intermittently but I was never able to replicate exactly based on what they mentioned. Though I was able to get some double downloads by machine-gun clicking on the download button. This fixes some of that and gives visibility on the waiting for download.

- Takes off the Rails UJS disable/auto-enable functionality for the button, which I think was messing with when it disables and re-enables the download.
- Created a new spinner icon for the download button that shows when it's disabled or in progress, switches classes around to show it.
- After a download begins or is canceled, wait 5 seconds before re-enabling the same download button to prevent impatience clicks while the person waits for the browser to show the download (in bottom or with popup "Save as" dialog).


ryscher:
Thanks, this works well and should help to prevent problems.

PR #699, opened 2021-11-12
Instructions for re-indexing solr as needed.

sfisher:
Update to instructions about clearing and re-indexing SOLR data.


ryscher:
Thanks for tracking this down.

PR #702, opened 2021-11-16
Index file types

sfisher:
This gets the file extension facet working.

The two config files need to be copied over to a solr server, but this is already done on dev if you're testing there.  If you run this branch against the dev solr instance it should work and allow faceting by file extension.

Added instructions for how to do it all since we do this pretty rarely and it takes a little time to re-figure-out each time without instructions.


ryscher:
This looks great. It will be interesting to see what it looks like with all the crazy file extensions that exist on the production server.

PR #703, opened 2021-11-17
Updates to curator download tool

sfisher:
Allows curators to download more easily and updates to work better under windows (and move minimal API config out of the code).

- Updates minimal instructions in README.md
- Looks for `C:\DryadData\api_config.yml` (I made it a hidden file in Windows) or `File.join(Dir.home, '.api_config.yml')` (also a hidden file in home directory).
- Moves to prompting user for the DOI/landing URL rather than having to type it as part of command.  I think may be easier for curators.
- Save path is based on a `base directory` + `username` + `sanitized doi` + `version number`.  Added the version number since I can imaging curators downloading more than one version and not clearing out the old one before downloading again and then they would have a jumble of files in there if it doesn't separate by version.
- Removed that one nasty file and commented out test that windows hates with the `*` in it.  If we really need this, then I can figure out a way to put it back.
- It says where it saved things at the end of the script.

On the windows machine, I did some stuff in my account.  I'll check with you about what I need to do with other accounts or test with them.

- I re-downloaded the https://rubyinstaller.org/downloads/   Ruby+Devkit 2.6.8-1 (x64) and installed it again as administrator.  I chose to install the number 3) with MinGW since I think it's needed for compiling some ruby gems that might have native-ish code.
- I pulled this branch to the machine under "\Program Files\dryad-app" for testing.
- `bundle install`  from the `script\file-download` subdirectory. (as administrator)
- I added a very simple batch file under `C:\DryadData\SoftwareShortcuts\api_downloader.bat` that changes directory, runs the file and then tells people to press a key to continue when it's done downloading.

PS.  Here is a kind of large dataset for testing downloading which I tried.  https://datadryad.org/stash/dataset/doi:10.5061%2Fdryad.vt4b8gtqm

---

Ok, I tried this under a new remote desktop account and it seems to work fine without any additional installation.  Steps to use:

1. In explorer, go to `C:\DryadData\SoftwareShortcuts` and double-click the `api_downloader.bat` file to execute.
2. Copy/paste the landing page URL (or type in DOI) into the window.
3. Watch it download.
4. Navigate to path like `C:\DryadData\testy2\doi_10.7272_Q6DZ06J3\v4` to see files.

### DIFF
--- a/script/file-download/Gemfile
+++ b/script/file-download/Gemfile
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
 
-gem 'byebug'
+# gem 'byebug'
 gem 'http'
 gem 'ruby-progressbar'
--- a/script/file-download/Gemfile
+++ b/script/file-download/Gemfile
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
 
-# gem 'byebug'
+gem 'byebug'
 gem 'http'
 gem 'ruby-progressbar'
--- a/script/file-download/Gemfile.lock
+++ b/script/file-download/Gemfile.lock
@@ -3,6 +3,7 @@ GEM
   specs:
     addressable (2.8.0)
       public_suffix (>= 2.0.2, < 5.0)
+    byebug (11.1.3)
     domain_name (0.5.20190701)
       unf (>= 0.0.5, < 1.0.0)
     ffi (1.15.4)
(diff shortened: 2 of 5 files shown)


ryscher:
Thanks for fixing this! I briefly showed it to the curators, and they were happy about it.

PR #704, opened 2021-11-17
Make API documentation more approachable for new users

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1508

- Expand the README in the `api` directory to include more overview and some simple examples.
- Move information about accounts and OAUTH into a separate document that can be referenced independently of the submission document
- Other minor cleanup


sfisher (approved):
This looks good to me.  One thing occurs to me since I notice that the <domain> isn't filled in.

We may want to suggest people use dryad-stg.cdlib.org while they're writing code and testing scripts, especially if they write any information or change data since we don't want a lot of garbage data in our production system and then move along to datadryad.org when their integration is stable.

Or maybe that is something that belongs elsewhere or in the email if someone requests api credentials.

PR #705, opened 2021-11-18
Add another relationship type of Data Management Plan

sfisher:
This turned out not to need a lot of other changes since it is handled correctly already.

Some of the relationships need transformation for Zenodo.  I checked code and this seems OK.


ryscher:
This is great. Glad it was simple to add.

PR #709, opened 2021-11-19
SOLR config changes wouldn't work on stage

sfisher:
The previous commits worked fine on development, but they caused errors on stage.

I still don't know why it worked differently in different environments.  I spent 2 or 3 hours doing diffs and trying to figure out what was wrong.

I finally had to get the production config which worked on stage and go through and retype all my changes in by hand.

This is already deployed and restarted on stage/production solr servers.

When deploying on production, you just need to do this after the deploy to update the indexes to have file information.

```
RAILS_ENV=<env> bundle exec rails rsolr:reindex
```


ryscher:
Odd that these additions were required. They didn't seem to make a difference on my local install.

PR #710, opened 2021-11-20
Standardize autocomplete

sfisher:
This makes a function that will attach autocomplete and fill in a hidden id when things are selected.

I added this for author affiliations and the research facility.

We can look into other places it might be useful later.  I'm unwilling to do any more of this right now and I've already worked a 10 hour day without lunch, or going outside or getting away from a screen.  If people need things done for a shorter sprint (since I'm out next week) they will need to plan accordingly.


ryscher:
Thanks Scott, this works well, and it's nice to have everything standardized. I had some problems with the merge conflicts, so I will merge it from a separate branch as https://github.com/CDL-Dryad/dryad-app/pull/713

PR #716, opened 2021-11-24
Make `end_with?` safe

ryscher:
We have been receiving some email errors relating to use of `end_with?` on a nil object. Unfortunately, there is javascript involved, making it difficult to get an exact stack trace to determine which `end_with?` is causing the problem. So I found every instance of `end_with?` in the code and made them robust to nil objects.


sfisher (approved):
Thanks for doing this, Ryan.

Yeah, I saw those errors.  I wonder if it was something to do with the asterisk on unidentified items.  In any case, this should make it more robust everywhere.

PR #717, opened 2021-11-29
small change to instructions for removal

sfisher:
Tells how to embargo and also adds "delete dataset" to heading in markdown so it's easier to search for in github.


ryscher:
Thanks for the clarifications.

PR #718, opened 2021-11-29
Update documentation for APIs and curator PC

ryscher:
- Added an example to the API README
- Moved the instructions for the Curator PC into their own document
- Added more documentation about using accounts on the PC, and how they may install software


sfisher (approved):
Looks good.  Thanks, Ryan.

PR #722, opened 2021-12-02
Add FOS keywords to API

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1534

Update API to include the Field of Science keywords. Only one FOS is allowed per dataset, and it is treated separately from the other keywords. Also, API users can only use the FOS values that are already present in Dryad; they may not add new ones.


sfisher (approved):
This looks good, Ryan.

It's annoying when we have FoS in the keywords and have to differentiate between fos and non-fos keywords, but this works well.

PR #723, opened 2021-12-02
CSV export of datasets with funder or author affiliation in string

sfisher:
This gets a bit more complicated then the SQL we have to do something similar since we're adding funders.  This has minimal info in the output, but seems like we didn't need more right now.

Getting from text instead of RORs since there were tons for Max Planck and not everyone correctly selected a ROR value, either.

The rake task just gets called like

```
bundle exec rails reports:from_text_institution name="Planck" RAILS_ENV=production
```

The other module has most of the code.
- big-ish query gets the datasets that have the string in author affiliation or contributor.
- latest version for each dataset
- returns `doi, pub_state, title, author_affiliations, contributor_affiliations`
- Only affiliations matching the string are output separated by pipes in the appropriate column, since we're possibly cramming more than one thing into a single field.

This was kind of one-off for now and I wasn't sure if there would be other requirements.  I haven't done tests yet and mostly SQL and some simple CSV output so just tried it manually.  If we need tests, I can add them, LMK.


sfisher:
Sorry for the bazillions of commits trying to get the tests to pass on github actions.

ryscher:
Looks great, works in my testing. Thanks for figuring out the stupid Node problem!

PR #725, opened 2021-12-04
Re-verify uploaded files at zenodo after "success" and before we delete from S3

sfisher:
We've had a couple of datasets where all files seemed to be removed from S3 but Zenodo didn't actually seem to have the files even though I believe we got success messages and matching md5s from them in a response to an upload.  Maybe their system was unstable or 🤷 

This adds another verification between our database list and zenodo's list (by re-querying the dataset from them) to be sure the names and sizes match before deleting from S3 and declaring success.  Raises error if there is a problem.

Also cleans up a @resource that the class was initializing with and was never used (and fixes some of the tests to match).


ryscher:
Thanks for adding all these extra protections!

PR #728, opened 2021-12-07
Curation workflow updates

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1480, https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1552, https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1556

- Reworking of the logic for auto-assigning curators and statuses after a dataset is processed by Merritt. See the included documentation file for an overview of the new rules.
- When a curator is unassigned, retain the current curation status, unless the status was `curation`

### DIFF
--- a/documentation/technical_notes/curation_workflow.md
+++ b/documentation/technical_notes/curation_workflow.md
@@ -1,0 +1,36 @@
+
+Curation Workflow
+==================
+
+Dryad attempts to assist curators with their work while retaining as much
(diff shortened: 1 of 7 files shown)


sfisher (approved):
This workflow stuff is complicated.

I like that you added date_last_curated and date_last_published and seems useful for bunch of other cases, also.

PR #729, opened 2021-12-07
Trying to get rid of babel warnings

sfisher:
They happened in the puppet deploy on stg, but this changes fixes them.


ryscher:
Great.

PR #730, opened 2021-12-08
Announces downloads for screen readers

sfisher:
Adds an aria live region and updates it with an announcement when the download begins.

If you want to see it happening, you can change the class of the element with id="accessible-dl-msg" in _download.html.erb from `screen-reader-only` to something else and it displays text on the screen.  Or edit the class in your browser inspector.  It is right by the download link on the page in the same div.


ryscher:
Looks good, works for me. Thanks!

PR #731, opened 2021-12-15
Complete revision of validations, verbose and linky

sfisher:
The accessibility team wanted changes to validation to link to all the specifics.  See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1567

The main things to notice from the user side.

- Validation still happens on the review page initially, but it provides links with all the errors and will go to the page that has problems and re-displays the errors on the top of the other page after following a link.
- Highlights most text-field fill issues and scrolls to the area on the page when a link is clicked.
- Any of the file issues are harder to trigger.  To get S3 to not have a file, you could change the filename in the database to test, for example.

On the technical side.

- The `Completions` class really only has warnings and other less critical errors now.  Also removed all the stuff in it about "3/5 items completed", etc, since it is no longer used anywhere and was just a bunch of confusing clutter.
- Removed some things from some dataset presenter class that was no longer used.
- The validation code is now in a DatasetValidations class.  When things error they need much more specific information: 1) error message with \{ \} which will get replaced with a link, 2) a URL for the page where the error lives, 3) HTML/CSS ID of the items that may be involved in the error which means they'll be highlighted and scrolled to view when clicking on the error message.

Some new css classes and a small amount of Javascript.  The submit button still disabled until errors are cleared.

Let me know if you find any issues.

PS.  I couldn't highlight the CKEditor field since it's some kind of special rich text editor and couldn't figure out, so it just highlights the label if abstract isn't filled.


sfisher:
Thanks for catching those things, @ryscher .  I'll look into the style to be a red border.  I also didn't realize that the popup for the other stuff went away.  I'll look when I get a chance.

PR #732, opened 2021-12-17
Require at least one README file and at least one data file

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1563

Every dataset must include at least one README file and one non-README file that is uploaded as a "data file".


sfisher (approved):
This works well for me.  Makes me upload a README.  :-)  I imagine the tests were a pain since it means you need to add a couple files to mosts tests to get a review page without the error messages.

I did have a weird thing where I had both a "readme.txt" and a "README.md"  It still warned about lowercase for the first, but I think that is fine since people shouldn't have both, really and they should probably make all of them uppercase if they have more than one.

PR #733, opened 2021-12-17
Require Research Domain

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1561


sfisher (approved):
I tested this out and it works well.

Thanks for hassling with all the annoying testing things that break when you make something newly required.

PR #734, opened 2021-12-18
Fix popup with errors

sfisher:
This changes two things and gets both the errors and the popup dialog working at the same time.

1. Javascript would cease functioning but give no errors after a function that showed errors on the screen.  It was because it was attempting to access an element that doesn't exist on all pages, so puts a check for the element with jQuery before the rest of the if statement or doing anything with it.
2. Makes the border color red for error info.

PS.  I also figured out that adding a task to for github actions to use a pre-compiled node means it spends a lot less time to run the tests since it doesn't compile node-sass manually from scratch every time if it's already present.


ryscher:
Thanks for fixing this!

PR #735, opened 2021-12-21
Minor updates to README requirement

ryscher:
Updates requested in the sprint meeting:
- add cutoff date, so READMEs are only required for datasets created today or later
- simplify error message when the README is incorrectly capitalized


sfisher (approved):
Thanks for those updates!

PR #739, opened 2022-01-04
Fixes submit button not being disabled on errors in some circumstances

sfisher:
This fixes the issue we saw in some cases where it would show errors for the title not being filled, but still allowed submission after the other "accept" buttons were checked.

The error was fairly easy to fix, but then it caused lots of errors with other tests where readme.txt files weren't set up or the accept buttons weren't checked.  There were also some tests that were extra slow since they were loading the review page 3 times or more.  I think this fixes the issue and those tests.


ryscher:
Looks good. Sorry the readme requirement is still making things messy. Hopefully all of the tests can handle it now.

PR #740, opened 2022-01-04
Treat tenant_curators as curators too

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1590

When performing a check of whether a user is a 'curator', including auto-assignment of curators for versioned datasets, treat a `tenant_curator` role the same as a `curator` role.


sfisher (approved):
Great.  Thanks for fixing this.

PR #741, opened 2022-01-06
Moving the existing file upload code into subdirectories

sfisher:
This should make it easier to organize and keep it separate since we'll probably see a proliferation of files soon and the components make sense to organize and nest a bit.

Updated references to it to get it working again.


ryscher:
Looks great. Thanks for organizing these.

PR #743, opened 2022-01-15
Get the simple title form working with react and resolve integration issues

sfisher:
- Restructures code a bit so we have separate areas for groups of components.
- Fixes webpacker so it adds it all and doesn't require to enter through the components as Cassiano had it set up.
- Title field that uses Formik for managing forms and finally figured out how to get reference so I could trigger the internal submitting action onBlur.
- Adds a basic test for it in Jest.  I can add a bit more later.
- Fixes a bunch of broken Javascript and broken Capybara tests. Cleans up some unneeded and confusing Javascript.
- Adds to Capybara setup so you can trigger seeing the window with CAPY_SHOW environment variable.
- Allows controller to accept JSON.
- Fixes current_user which was making me log in again every time I changed code.

Sorry, this took a while, but there were a lot of annoying details to set up and figure out.

We probably also want to add a linter for Javascript.  I'll add a separate ticket.


ryscher:
Looks good, just remember the prod server will need Formik installed.

PR #746, opened 2022-01-18
Fix for nil contributor names

sfisher:
For some reason, we now got a few nil contributor names and they were causing metadata generation errors.


ryscher:
This looks good.

PR #750, opened 2022-01-20
Add Hindawi to list of publisher members

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1613


sfisher (approved):
Thanks for adding this, Ryan.

PR #751, opened 2022-01-20
Adds linting for javascript in the React code

sfisher:
I spent a bunch of time configuring ESLint and also fixing Cassiano's old react code.

For some components or modules I put in ignores for some rules.  For the 600-line `UploadFiles.js` container, I put in a .eslintignore since it has one item preventing validation, but when/if I could figure out how to fix that I have no doubt it will give me hundreds of other things.

We'll need to go back and fix some of that code in the future, but for now, I'm hoping I can get pretty clean code for the new things we build.

This doesn't check code outside of react right now since that code is a mix of very old and different syntaxes and seems to currently work.  When we rewrite some of that we can look at modernizing it.


sfisher:
Ok.  I think I got this in a usable state and runs under the Jest tests in here to check the React section of Javascript.

ryscher:
Looks good, works for me. I think we might want to disable the no-console rule, though.

PR #752, opened 2022-01-25
Adds the descriptions to react and with TinyMCE instead of ckeditor

sfisher:
This took a while to get right, especially tests.

This modifies the abstract, methods and usage descriptions to use TinyMCE.

I'll need to deploy on dev and test that the key warning doesn't show up since I put one in.

It builds on the previous pull requests which put in the ESLint javascript linting.

This is tested pretty extensively in our capybara feature tests, but I can add something to Jest, also, if you want.


ryscher:
Looks good, works well.

Note: Some versions of Webpack and Node don't work well together, and produce an error about "digital envelope routines::unsupported". If this happens, your shell should have this variable set: `export NODE_OPTIONS=--openssl-legacy-provider`

PR #753, opened 2022-01-26
Updating datacite-mapping version for bundler and rake

sfisher:
The new version of datacite-mapping has newer versions of bundler and rake under development dependencies, which don't affect the way the code runs, afaik, but makes the dependabot warnings about dev dependencies go away so everyone is happier.

Also some other gems got updated with the bundler update in our code.

I tested submitting and publishing an item and no changes, as expected.

PS.  Puma updated, but I had to roll it back to the current version we were using before because the new version made tests fail.


ryscher:
Looks great, thanks for tracking down the Puma issue.

PR #755, opened 2022-01-28
Move API engine into main application

ryscher:
- Moved all models, views, and controllers to the top-level application
- Merged the old routes file and Gemfile into the top-level files
- Although some files look completely "new", they don't actually contain new code. This is a result of the file moving, and the indentation changing slightly due to the new usage.
- Moved the database migrations, though I'm not 100% certain they are replayable as-is in the new location.
- Removed a lot of "duplicate" files that are no longer needed when we're not using an engine.


sfisher (approved):
Wow.  This was a ton of work, @ryscher .  Thanks for dealing with all this and all the details to move this around.

I read over everything, ran the request tests and did a bunch of spot checking.  It seems like everything works ok.  I even tested with some private without logging in to be sure it prevented viewing and looks good.

Though on a tangent it says `Identifier cannot be viewed. Either you lack permission to view it, or it is missing required elements`.  I wonder if we should just give a 404 or something else for those so that people cant see if datasets exist or not if they shouldn't see them.  That is completely unrelated to this ticket, though.

Thanks for all the work.

PR #763, opened 2022-02-03
Add AAAS to publisher members

ryscher:
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1632


sfisher (approved):
Thanks for adding these to the list.

PR #764, opened 2022-02-07
Clarify directory name for putting submissions on hold

ryscher:
This directory changed slightly with the Linux 2 servers. Adding an absolute path to make it easier to find.


sfisher (approved):
Thanks, Ryan.  Looks good.

PR #765, opened 2022-02-09
Improve precision of API searches using modifiedSince

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1620

Previously, API searches for `modifiedSince` used the `timestamp` field in SOLR. This field is updated every time SOLR is reindexed, so it produces a lot of false positives. This PR changes the behavior to use a new field, `updated_at_dt`, which uses the `updated_at` field from the resource being indexed.


sfisher (approved):
This looks good.  Do you need to run a solr update or anything for this when we deploy?

ryscher:
Yes, a reindex will be required.

PR #767, opened 2022-02-12
Show more journals on journals page

ryscher:
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1637

Expand the journals page to include all journals that have sent a metadata email in the past year, regardless of payment status. 

This does two queries (one for payment status and one for metadata status), then combines the results and issues a third query. It could be made more efficient using a single SQL statement, but that statement would be much more complex and difficult to maintain, and we don't have enough journals for the speed to be a concern at this point.


sfisher (approved):
Looks good, Ryan.  I also tested and it works great.

PR #768, opened 2022-02-14
Treadway report -- Curation status dates

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1650

New report of the datasets, indicating the first date that each curation status was reached. This report is much more simplistic than related reports of curation activity, but hopefully it will still be useful.

(Also moved the `curation_timeline_report` into the section with other curation reports, but I didn't actually change its contents.)


sfisher (approved):
Thanks for putting this together for the reporting they wanted.

PR #771, opened 2022-02-15
Fix GMail authorization

ryscher:
GMail messages have not imported since January 19th. Apparently, upgrading the Yaml gem breaks something in the current Gmail gem.

To test:
`bundle exec rails journal_email:validate_gmail_connection RAILS_ENV=local`
(you will either receive an exception or a message stating it is working)


sfisher (approved):
Thanks for finding this and rolling it back.  What a pain.

PR #772, opened 2022-02-15
Update staff page

ryscher:
Mariah is no longer with Dryad.


sfisher (approved):
Thanks for removing this.

PR #773, opened 2022-02-15
This adds bold appearance for strong text

sfisher:
There is something weird about our default font that doesn't show bold.  I can display bold by adding a shadow style which seems to work fine in both Chrome and Firefox.  The dataset at \<dev-domain\>/stash/dataset/doi:10.5072/FK2QR4XG4M shows an example.

### DIFF
--- a/stash/stash_engine/app/assets/stylesheets/stash_engine/ui.css
+++ b/stash/stash_engine/app/assets/stylesheets/stash_engine/ui.css
@@ -5524,6 +5524,10 @@ table.c-table_institutions ul li {
     margin: 0;
     font-size: 1.2em; } }
 
+strong {
+  font-weight: bold;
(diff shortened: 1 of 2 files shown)


ryscher:
Thanks, this is better. It's still not incredibly noticeable. I see elsewhere on the landing page, where we have bold content, we also increase the font-size to 1.1em. Maybe that would help more?

sfisher:
Ok.  Added size 1.1em.  Maybe your browser is different but it looks pretty dark on my mac.

ryscher:
Thanks for adding this. It does look even more bold. Looking again, when I remove the 1.1, it still looks pretty good, so I wonder if my browser wasn't rendering the shadow before. So it's fine whichever way you prefer.

PR #774, opened 2022-02-15
This gives us a quick and dirty way to download a large number of files from S3

sfisher:
A basic rake task to download files from S3 since it becomes unwieldy to retrieve a large number of files from S3 by hand.

A user uploaded 15,169 individual files.  Our limit is 1,000 files.  This will give us a way to download them and zip them for her if we need to do that and create a zip container for her.

(The review page will not load because of the large number of files, but even if it did--it wouldn't allow submission of over 1,000 files.)

This only works on new files at S3 and goes through each and downloads into a directory named after the resource id in Rails.root.

BTW, tested this manually on an item with 10 files like:

`RAILS_ENV=local_dev bundle exec rails dev_ops:download_s3 3575`

### DIFF
--- a/lib/tasks/dev_ops.rake
+++ b/lib/tasks/dev_ops.rake
@@ -1,8 +1,10 @@
 require 'yaml'
 require_relative 'dev_ops/passenger'
 require_relative 'dev_ops/download_uri'
+require_relative 'dev_ops/download_s3'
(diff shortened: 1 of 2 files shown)


ryscher:
Thanks for figuring out the problem and a quick way to solve it!

PR #778, opened 2022-02-18
Improve note when resource owner auto-assigned by API

ryscher:
Previously, most auto-assignment was performed using an "edit code". Since #610, it is more common for the auto-assignment to happen at the time the dataset is initially created. So this PR updates the note to inform curators about the owner that is assigned at creation.


sfisher (approved):
Thanks for the update, Ryan.

PR #780, opened 2022-02-22
Adds 1) generic autocomplete component and 2) single ror autocomplete

sfisher:
- Generic autocomplete component called "GenericNameIdAutocomplete" since it supplies the UI and methods to autocomplete and return both the text (name) and Id with the autocomplete and with label and text box.  Does display and functionality such as debouncing requests so we don't spam APIs.
- Ror-specific autocomplete which has stuff specific to ROR lookups.

I added explanation of how it works to the generic file or looking at the Ror-specifics helps, also.

I updated our tests so they're passing again and has a basic test for loading the component.  However to get better tests working for the complex component, I'll need to mock the API calls in Javascript and also have some better way to trigger behavior from the component and check the DOM results (which I think means adding Enzyme style tests, but not 100% sure).  I'll add that in another PR.

I'll also go through and replace more than the one autocomplete in future PRs.


ryscher:
Looks good, works for me. Thanks for all the work getting this sorted out!

PR #781, opened 2022-02-22
Update connections to Linkout & Labslink

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1651

The content that we send to NCBI/EuropePMC depends on the presence of PubMed IDs in our database. Although we had previously created code to update the PubMed and Genbank IDs, we were not regularly running the code, and some bugs had crept in.

This PR does: 
- add `seed_pmids` to the monthly process, so we're actually updating the links to items that have PMIDs.
- improve `seed_pmids` to find PMIDs from more forms of article DOIs
- add `seed_genbank_ids` to the monthly process
- restrict the seed processes to only query datasets that have been created in the past year, so we are not constantly querying datasets that will never have corresponding content in PubMed
- move the `link_out:publish` process from weekly to monthly, since the content doesn't change quickly


sfisher (approved):
Thanks for catching this and doing the update.  This is really helpful.

PR #782, opened 2022-02-25
Adding some testing for the autocomplete components

sfisher:
- Setup and configured React Testing Library https://testing-library.com/docs/react-testing-library/intro/ .  Originally went toward Enzyme since many tutorials and things recommended it, but needs hacks for React 17 and will not work in React 18 (and hasn't been well maintained apparently).
- Turned off ESLint checks for console.log
- Tried to test actual dropdown functionality itself, but very difficult to get working, so tests loading and text input and resetting of ROR ids when changed.
- Lots of npm package changes, some test data for ROR and some moving the structure around.


ryscher:
Thanks for getting this working. Looks like it will be very useful.

PR #785, opened 2022-02-28
Move stash_discovery engine into main application

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1619

- moved all functional files
- merged routes and Gemfiles
- removed a lot of placeholder stuff


sfisher (commented):
I tested this out and it seems like it all works except the map doesn't load consistently now.  Sometimes there is a blank space on the right where the map should be and sometimes it loads.  Mostly I'm seeing a blank space and then if I refresh the page again it loads.

We can try a deploy on stage to see since maybe it will act different there.

ryscher:
Update looks great. Thanks for fixing this!

PR #786, opened 2022-02-28
Autocomplete for Facility

sfisher:
This uses the same data source (RoR) but submits differently since the form was different and submits by json.


ryscher:
Looks good. Thanks for getting it working.

PR #787, opened 2022-03-01
Autocomplete for Funder

sfisher:
This gets the autocomplete working for funder, similar to the ROR dropdowns.

I ran into testing problems suddenly from tinyMCE, so this also includes tests to delay until tinyMCE is really fully loaded.

Also had to modify the old capybara tests and add a couple JS tests for it.


ryscher:
Looks good, tests fine.

PR #788, opened 2022-03-01
Preprints report

ryscher:
A quick report on datasets associated with preprints. This is a little messy, because although we currently have a relationship `work_type` for "Preprint", we formerly only had the DataCite relationship types, so we have to explicitly look for older relationships that contain likely prefixes.


sfisher (approved):
Thanks for doing this.  It sounds like this is useful for the presentations.

PR #790, opened 2022-03-09
Adds the whole funder section as a react component

sfisher:
This is the full component for all the funders which takes in about 3 subcomponents.

Everything is working as far as I can tell and I added Jest tests for some of the actions which turns out to be a bit tricky because of asynchronous  actions, renderings and javascript but I think I finally got the hang of it better.

I also had to add some checks for some of our functions that call out to some global things in the page like showing the twirly for saving or the "are you sure?" dialogs since they were not present in the limited react testing framework test rendering.

Also fixes Capybara tests so they work in rails and rspec.  I also found and fixed some more ID bugs for out lists late in the process.  In all looks pretty OK, I think.

This seems like a lot of code to do something that was less code before (even if it was spread around a lot of places in rails).  I guess it gives us flexibility for flashier interactions with the UI.


ryscher:
Thanks for rearranging this! All tests good for me.

PR #791, opened 2022-03-09
Correct "something went wrong" for funder admins

ryscher:
Normally, `funder_id` and `name_identifier_id` have either a value or an empty string. This protects against instances where the field has a `NULL`.


sfisher (approved):
Thanks for adding this, Ryan.

PR #795, opened 2022-03-15
Transforms keyword autocomplete and list into React

sfisher:
This recreates the functionality of the keywords and adds tests.


ryscher:
Looks good, works fine. Thanks!

One minor thing I noticed -- in the existing feature (before this branch), when you are typing a keyword and hit Tab, the focus goes to the next keyword. Using this branch, the focus instead goes to the next field. I'm not sure whether this change is intentional, and I'm also not sure which is better.

PR #797, opened 2022-03-16
Fixes Funder add/remove weirdness

sfisher:
This now creates a new funder in the database so it has a proper ID that doesn't change.  (which caused lots of weird problems with rendering).

Simplified code and fixed tests.  This seems very reliable now.


ryscher:
I'm getting some strange behavior:
- When I click a value in the autocomplete, it is not filled in in the UI, and not saved in the DB.
- When the "Granting Organization" field has focus, and I click on "add another funder", the initial click only changes the focus, and a second click is needed to actually make a new field appear. But then if I refresh the page, there are fields for each button press.

sfisher:
Thanks Ryan.  I'll look into this.  It seems to work most of the time but I'm ocasionally getting things where it doesn't seem to trigger or it seems to trigger more than once maybe.  This will be fun to figure out.  :-/

sfisher:
Ok, @ryscher .  I fixed a number of things.

1. The "Are you sure?" dialog was keeping old bound events for the buttons so it would fire delete events more and more repeatedly over time.  Removed events before re-attaching them so there is only one.
2. Fixed the "Are you sure?" css so that it doesn't make the dialog appear on the top of the document, but it should show on the right scroll position on the screen.
3. I could duplicate your problem with list acting wonky by entering a funder name and directly clicking the "add another" item.  This causes a race condition where it saves the data from one item and creates a new item at the same time. Both modify the list or contents of an item in the list at the same time.  Whichever updates the state last "wins" and wipes out the other action's state. It turns out there is some "prevState" specialness for cases like this.  See https://stackoverflow.com/questions/67309672/how-to-prevent-react-state-overwrites-due-to-asynchronous-code-and-websocket-rac and https://stackoverflow.com/questions/54807454/what-is-prevstate-in-reactjs .  I guess it takes a function and then the setState function batches or combines updates so none of the state updates interfere with the others.

I was never able to see any problems getting the values chosen from the autocomplete saving.  Are you still seeing this?  You need to press `Enter` to select an item.  Tab is the key that navigates out and enter is the key to choose an item from the list.

ryscher:
The issues with add/remove are resolved, and the "Are you sure?" box is greatly improved. However, now the funder autocomplete is not displaying for me at all. Wondering whether it is some weird issue with cached JS....

ryscher:
Ok, now that Crossref is functional, the autocomplete populates again. It does work correctly when you navigate with the keyboard (down arrow and Enter). However, if you use a mouse to click on the value, it does not work.

sfisher:
Hi @ryscher,

I do not see this problem (not saving when clicking an item) when testing on my local machine.  I also just deployed this revision to our development server and I'm not seeing it there, either.  I tested in chrome, firefox and safari on my mac and never saw a problem saving.

Did you have a stale page after restarting the server or your session, perhaps? The page has an authenticity token defined in it and if the authenticity token doesn't match the current one defined for the user session then all CRUD operations are denied by Rails to prevent cross site request forgery (csrf) or other security problems.

Can you test this on our development server and let me know if it works for you?

Otherwise, I'll need more information since I'm not able to duplicate this problem.

ryscher:
Thanks for reworking this. The issue is fixed for all of my tests now.

PR #798, opened 2022-03-18
Move ROR functionality into the local database

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1667

Updates include:
- Create local model to store ROR organizations, `StashEngine::RorOrg`
- Add an importer that downloads the latest ROR data file from Zenodo and updates the content of the model. Note: the importer has a hardcoded the directory for temporary downloads of the ROR files. This is similar to crons and the notifier, since we want to retain the downloaded file even when the Dryad code is redeployed.
- Move functionality from the old ROR utility class into the RorOrg model
- Simplify ROR autocomplete code, since the model handles sorting and we do not need to re-sort data multiple times
- Removes the ROR Mock from the tests in favor of using "actual" data with the model
  
The rspec tests will run without any special adjustments, but to test manually, you will need to import ROR content, which can take several hours:
`bundle exec rails affiliation_import:update_ror_orgs`


sfisher:
Hi @ryscher,

I tried testing this manually and I didn't know if there were extra steps I needed to take to populate the database.

I ran it and it completed in about 5 seconds and there was nothing populated into the database when I tried it on the dryad development server.

```
RAILS_ENV=development bundle exec rails affiliation_import:update_ror_orgs
...
Downloading ROR data file...
Validating file against expected checksum...
Processing file contents into database...
```

Any thoughts?

ryscher:
Yes, you will need to wait for the import to complete... will probably take a few hours. (It will output an update every 1000 records)

sfisher (approved):
Thanks for doing this, Ryan.

It looks really good and works well now.  Crazy that they changed the filename for the Ror download.

PR #799, opened 2022-03-22
Default funder crossref

sfisher:
Defaults id type to crossref funder id, even if they don't actually fill an id.  crossref_funder_id as identifier type is needed even if no identifier present in order to satisfy DataCite mapping and this seems to be what we'd done up to this point. (builds on the other funder fix pr)


ryscher:
Looks good. Not sure why GitHub is having trouble tracking the difference with this branch; it boils down to a one-line change.

PR #801, opened 2022-03-23
Allow journals to specify that their datasets should default to PPR

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1664

Adds a setting to `stash_engine_journals` that allows journals to specify that any dataset associated with that journal should default to PPR. When a journal name is set or changed, the journal setting will be used to determine the status of the `hold_for_peer_review` option (i.e., whether the PPR checkbox is checked).


sfisher (approved):
This looks good.  Thanks for adding this and adding to the documentation about it.

PR #802, opened 2022-03-24
Database field and controller method for changing order

sfisher:
This makes the migrations for adding author order to the database and creates an ordering method in a controller where you just send the ID of the author and the order for each one. Also tests.


ryscher:
Looks good. Thanks for keeping it simple.

PR #804, opened 2022-03-24
Changes for ordering

sfisher:
- Makes default order by the author order id if present
- Adds to api as the field "order" and allows viewing and setting and adds to swagger
- Adds test to model (shuffling authors)

Looked in our UI and it orders correctly on test data I put in manually in all the UI pages and also outputs to Merritt/Datacite and other places correctly since it's not default order if the order gets set.

(I put in names as originalnumber for the original number and reordernumber for the new order.)

Screen Shot 2022-03-24 at 3 19 09 PM


ryscher:
Looks good, and works well! Thanks for getting this working.

PR #805, opened 2022-03-30
More trigger troubles

sfisher:
I wasn't sure what you did to get the others to run.  I searched the code if you added "trigger" somewhere but didn't see it.  I eventually came up with this solution to add them manually for tests.

Also apparently frictionless has put in breaking changes in their latest versions (4.28.0, maybe) by changing the output and made our tests fail, so I had to lock the library to an earlier version with pip in the setup.


ryscher:
Thanks for tracking this down! Sorry it happened to you as well; I had hoped that my experience was a one-off. But I'm glad you found a way to prevent it for the future.

PR #806, opened 2022-03-30
Remove old debugging array

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1712

This extra array had been left in from previous debugging. Although Scott was seeing extended load times for it, I was only seeing a fraction of a second, so I didn't notice that it was still in place.


sfisher (approved):
Thanks Ryan.  I didn't think this was used for anything, but wasn't 100% sure.

PR #812, opened 2022-04-11
Require at least one funder

ryscher:
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1716

Require a funder to be entered. 

Since this changes the help text in GenericNameAutocomplete, the note about "N/A" is also shown in the help text for affiliations. This seemed cleaner than making it customized to the funder field, but I'm certainly open to feedback on this approach.


sfisher (approved):
This looks good to me.  I tried this branch and it seemed to work.  Thanks for figuring this out.

PR #814, opened 2022-04-12
Corresponding author changes

sfisher:
If you merge the other pull request first it should show a cleaner diff of changes just for the corresponding author stuff.

- Changes what is passed into some of the react components
- Adds a new method for comparing authors and the resource owner's orcid. (it's called `owner_author` on the resource model)
- Opening a new dataset to the metadata edit page adds a new author with the owner's orcid if it's not there already. Or if an existing author has the exact same first/last name then it adds the orcid to that author instead.
- Validation requires that one of the authors has the same orcid as the dataset owner and that there is an email address for that author.
- Updated emailer to choose that author to send emails to rather than the first one.
- Lots of test updates since our test data wasn't always very careful in creating owners and authors so they both had matching orcids and needed to submit.

I'm a little concerned about the requirement of an author having a matching orcid to the dataset owner for submission, especially from the API since I'm not sure that this is always fulfilled, especially from publishers.


ryscher:
Thanks for rearranging this at the last minute! All looks good, and works as far as I can test.

PR #815, opened 2022-04-13
Update salesforce cases to use owner_author

ryscher:


sfisher (approved):
Thanks for catching this, Ryan.

PR #818, opened 2022-04-13
Move StashEngine engine into main application

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1618

- Moved all models, views, and controllers to the top-level application
- Merged the old routes file and Gemfile into the top-level files
- Removed a lot of "duplicate" files that are no longer needed when we're not using an engine.
- Updated references to routes and relative paths

Most of the changes involve updating paths for the files and routes in new locations. I chose to explicitly use `stash_url_helpers` in many places where it was previously included automatically, since this simplified tracking down some of the routing issues.


sfisher (approved):
Holy cow!  These are a lot of changes.  Thanks for catching all the StashEngine:: things that don't work anymore.

I tried out the main submission path and landing pages and all is fine.  Geoblacklight has some problems loading, but that is similar to main right now, so is a problem to fix before deploy.

I think before release, we should also try out all the weird code paths like merging accounts, curation things, orcid registration, etc.  However I think we can do this over the next while and before release and we can enlist Daniella or others to help, also, before release.

PR #820, opened 2022-04-13
Errors loading asset on stage with rails helpers

sfisher:
This caused the page not to load because it wasn't able to get the asset and sprockets is annoying.

I tried putting the image file many places and nothing worked, so I finally just ignored the asset pipeline and put the icon in the public folder where it bypasses the asset pipeline.

Also removed the ckeditor which adds a bunch to the precompile since it's no longer needed.


ryscher:
This looks good. Thanks for dealing with this annoying issue. 

Note that the ckeditor stuff is removed already in the engine branch... hopefully git is smart enough to not cause conflicts over it.

PR #821, opened 2022-04-13
Security updates 2022 04

sfisher:
lots of gem/npm updates.


ryscher:
Looks good, works as far as I can see.

PR #822, opened 2022-04-13
Hide help text 

sfisher:
so people with screen readers can still access it the display. was disliked and the default message there from Dragon Drop might be useful to screen readers.

Also unbolded the title since it was the only one that was bold and none of the other labels.


ryscher:
Looks good, works for me. Thanks for doing these last-minute fixes!

PR #823, opened 2022-04-14
Leftover changes that went into release

sfisher:
This should have the last minute changes for our deploy. It seems like the Gemfile rollback fixed the geoblacklight problems until we upgrade them again.  :-/

BTW, v0.7.25b is the tag that we deployed and seems to be error free for this sprint.


ryscher:
I'm going to leave this unmerged for now. It looks like this is almost entirely gem downgrades which were subsequently reinstated by https://github.com/CDL-Dryad/dryad-app/pull/825

PR #824, opened 2022-04-14
Retry recent Merritt Errors every 5 minutes

sfisher:
This picks up on recent Merritt errors (last day) and tries to resubmit those items. Runs for something like 16 hours before it stops retrying.

This was a stopgap measure for Merritt LDAP issues causing some submissions to fail intermittently, but not consistently and this keeps thing moving along until everyone can get together to troubleshoot.


ryscher:
Looks great. Thanks for adding this so we have an alternative for working around intermittent Merritt problems.

PR #825, opened 2022-04-15
Big batch of gem updates

sfisher:
I went through and updated the gems I could that didn't seem problematic.

I am no longer getting issues with geoblacklight.  I think it might have been because of one of the support classes being a different version from the rails version.

I updated rails a tiny version and also updated bundler (since it kept getting stuck).

I believe it all works, but we should look carefully as we also test out the stash engine changes and go through all the paths.  I tested login and the most prominent pages.

PS. also tested deploy on development and it works (after puma restart).


ryscher:
Looks like these all work now.

Note for future me: I was originally getting geoblacklight errors when reviewing this code, but the root case was that sshuttle had not properly shut down, which kept my test machine from reaching the solr server.

PR #827, opened 2022-04-18
Update email validation regex

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1708

We already had email validation in place, so the issue described in the ticket must have not been processed by the UI (e.g., updated directly in the DB or migrated from the v1 system).

But I think it's useful to use the "official" email validation regex instead of our own.


ryscher:
Ugh. Withdrawing this, since it misses some cases we were previously catching. Weird, Ruby!

PR #828, opened 2022-04-20
push people to datadryad.org from www.datadryad.org

sfisher:
This will simplify our life and mean we don't have to check everything dependent on domain names to support the www domain name, also. (from https://stackoverflow.com/questions/22235977/ruby-on-rails-redirecting-www-to-non-www-version-of-site )


sfisher:
I'll check the tests on this.

ryscher:
This should be helpful. Thanks for simplifying the hostnames in the tests.

PR #832, opened 2022-04-27
Fix search results error

ryscher:
Correct issue with navigating from search results to landing pages.


sfisher (approved):
Thanks for updating this, @ryscher .

PR #833, opened 2022-04-27
Took out the YUI dependency

sfisher:
YUI (Yahoo UI) css minifier isn't the rails default.  We originally switched to it like a decade ago because the uglifier css minifier had a bug where it didn't correctly encode something when minifying and caused errors.

The time we did that was probably around 7 or more years ago, so I'm really hoping any bugs in Uglifier encoding and minification have been fixed by now.

I'd rather get rid of the extra dependencies that we don't really need when we can.  I already took YUI out of our Gemfile but forgot to change up the css minification in some environments.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,3 +1,2 @@
 module ApplicationHelper
-
 end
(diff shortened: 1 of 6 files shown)


ryscher:
Looks good, though I guess it only applies to stage and prod.

PR #835, opened 2022-04-27
Fixes the css compressor

sfisher:
This only happens on stage/production, so I had a heck of a time getting it right.

We're already using sassc from a bootstrap dependency and it can compress assets.  It is faster and the default if present.  I removed the lines for this in the environments and got it to successfully deploy on stage after a bunch of attempts.


ryscher:
Looks good. From the test history, it looks like the keywords test did pass, so you might want to try reinstating it again, but this is fine to merge regardless.

PR #837, opened 2022-05-02
Comment out counter process

ryscher:
It sounds like the counter process still needs a little babysitting, so I propose commenting it out, in preparation for the time when the server will actually run the monthly cron script again.


sfisher (approved):
I think this is fine for now since we have to babysit it each month when it get interrupted by maintenance on the first week and we have to restart it and then it takes a week more to finish.

I don't think we ever envisioned it taking so long to run.  It gets longer each week of the month as more data accumulates. It's starting to seem a little crazy to try to send it weekly, but I guess it's OK as long as we're fine with one core of our cpu on the server processing counter stats all the time. 🤷‍♂️

PR #838, opened 2022-05-05
Preliminary information auto-populate choices into React

sfisher:
This replaces our current code though most of the server-side logic has been left alone and I fit our UI into it.

- Changing the radio button for input type saves the setting into the dataset (ajax), so we don't have to guess what the user selected based on archeological evidence from lots of different values that might or might not be saved all over our database.  Also re-renders with the appropriate form instead of the confusing combined form with bits being replaced we had before.
- Even though there are a lot of similarities between the two import options they're separated out in case handling diverges (such as querying different data sources for autocomplete) and small differences in handling and settings.
- Typing and blurring away from elements or autocompleting will save changes to the database.
- Clicking the button below attempts to import metadata and if there are problems displays a message about results.
- No longer bug where you have to click outside form before clicking import button.
- I added a bunch of stuff to javascript libraries to resolve a bunch of npm library conflicts that prevented the ui library from working on my machine.  I did these with `yarn add -D <library>`.  This makes them development dependencies so they shouldn't need to be loaded on real web sites, but may be useful to add to our lock files so that people can reproduce getting the ui library working by just doing `yarn install`. Also had to reconfigure a few config files for ui library since they no longer had correct settings for paths or our environment in some cases.

Some test data I tried out:

```
manuscript example:
Journal: Ecological Applications
MS number: 20-0939

DOI Article examples:
Journal of Ecology
10.1111/1365-2745.13903

American Chemical Society (or it is Crystal Growth & Design according to crossref).
10.1021/acs.cgd.0c00978

Cold Spring Harbor Laboratory
10.1101/2022.03.21.22272672

Nature
https://doi.org/10.1038/s41586-021-04228-1
```


ryscher:
Generally looks good. I found two minor issues:
- When I import metadata from a DOI, the DOI appears in two duplicate Related Identifiers.
- If the dataset has a primary article, DOI, and a new one is imported, the original one remains and fills the field at the top

Not sure whether either of these is urgent.

PR #839, opened 2022-05-05
Suppress sending relationships to DataCite

ryscher:
When metadata is sent to DataCite, do not include the related identifiers. This is a temporary change, allowing Datacite to change relationship handling in their system. This change should be rolled back when we complete https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1750


sfisher (approved):
Thank you for finding and killing this, @ryscher .

PR #842, opened 2022-05-09
Some small instructions on how to make a corporate author

sfisher:
We need to keep doing this for the CTP and I need to root around and figure out how to do it each time, so will save us time to document.


ryscher:
Always good to document these things. Thanks!

PR #843, opened 2022-05-10
User Management page -- main features

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1724.

- Refactors the user management page to make it more obvious which code belongs to user management and which code belongs to dataset management
- Updates the process for setting roles, using all of the available options and displaying their details
- Adds a popup to select each user's tenant
- Adds a set of selection checkboxes, which can be used for any operations that affect multiple users
- Adds a function to merge two users
- Adds a read-only popup for information about users that have admin privileges over journal content


sfisher (approved):
Thanks for doing this, @ryscher .  This turned into a lot more work than we realized at first.

I love being able to search for users, change roles and am able to merge users here.  I think Jess will like it, also.

I read through the code and tested the functionality and looks good.

If we ever decide to do more work on this in the future, I think we could probably streamline and remove some of the "at a glance" info about how many datasets if we wanted.  I also notice that if I click a user's name and then see the datasets then the "see history" button doesn't seem to do anything (or maybe it just reloads the same page?).

These things are just odd details on a page no one else but us and Jess use and are probably leftovers from previous work we did.  They are not worth spending any time on unless the use of this page expands or changes.

Also thanks for figuring out explicitly documenting the rails popup functionality since it will be useful for the future.

PR #844, opened 2022-05-10
Allow curators to have HTML code editing

sfisher:
It's last on the toolbar under the ... icon to expand.  Only curators and above get it and uses the built in tinyMCE option for this.

I didn't allow it for normal people since allowing to edit all code may make malformed HTML (which the plugin seems to eliminate) and also allows people to do things like insert images or other functionality which we may not want.

We once had someone who made an entire paper to post for free in the old Dash system by editing code, encoding photos inline and the rest.  It had photos of churches and other things. Allowing things like photos is a can of worms since then we also need to host them, encode them inline or be afraid that externally supplied photos will go dead and break the rendering.

So I think maybe it's best to keep normal users from having HTML editing.

PS.  Javascript is stripped from the html when things are saved.

### DIFF
--- a/app/javascript/react/components/MetadataEntry/Description.js
+++ b/app/javascript/react/components/MetadataEntry/Description.js
@@ -8,7 +8,7 @@ import PropTypes from 'prop-types';
 import {showSavedMsg, showSavingMsg} from '../../../lib/utils';
 
 export default function Description({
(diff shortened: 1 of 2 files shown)


ryscher:
This is great new functionality. Thanks!

PR #845, opened 2022-05-11
Detect if previously public and don't send doi if so

sfisher:
PS.  You may not want to merge this into the main branch.  I can probably just deploy from this branch temporarily and if it's not merged into main for our next full release then you don't have to take it back out.

Adds a function to resource to detect if one was previously public.

Adds logic to skip updating doi if it was previously public.


ryscher:
All looks good, and the test deploy on stage seems fine. I'll leave the PR here for future reference.

ryscher:
Closing this, since it is no longer needed. Relevant parts were copied into #848.

PR #848, opened 2022-05-13
This changes the identifier relationships to DataCite (and others)

sfisher:
This is for tickets #1750 and #1752 .  However not all the zenodo batch metadata update resubmission code is ready yet (moving to another ticket).

See the markdown file [documentation/sql_queries/updates_to_relationships.md](https://github.com/CDL-Dryad/dryad-app/compare/reverse-identifier-relationships?expand=1#diff-bbd1e07a2305861826412510fc091dbd45f07b3d071a3ef43329795286f0d724) which documents the final changes based on the emails and other tickets entered here.  The file also has three SQL queries that can be run when we're ready to update legacy data.  These database updates are fairly simple.

I searched through all code for these and took into account context to carefully update the values and the text.  The relationship for us adding things to zenodo changed according to #1752 (supplemental and software) and also the inverse relationships for them linking back to us which we submit the metadata to them.

It's not an extreme number of code changes, but took a while to be confident I'd found them all and had things right.  See also their https://support.datacite.org/docs/relationtype_for_citation and emails.

### DIFF
--- a/app/controllers/stash_datacite/publications_controller.rb
+++ b/app/controllers/stash_datacite/publications_controller.rb
@@ -99,7 +99,7 @@ module StashDatacite
       ri = RelatedIdentifier.create(resource_id: @resource.id,
                                     related_identifier: standard_doi,
                                     related_identifier_type: 'doi',
-                                    relation_type: 'cites', # based on what Daniella defined for auto-added articles from elsewhere
+                                    relation_type: 'iscitedby',
                                     work_type: 'primary_article',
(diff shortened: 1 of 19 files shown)


ryscher:
This looks good and tests fine. I have added the utility method from #845, since we probably won't be merging that. I also added a revert of the code that suppressed sending relationships to DataCite (#839).

I will leave this unmerged until DataCite confirms that everything is working correctly on their end.

PR #855, opened 2022-05-25
Don't let crossref set dataset publication date

ryscher:
I think this will prevent the issue described in https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1753

The mechanism for importing CrossRef metadata had been setting the publication_date on the resource, but the date it was receiving was the publication date for an article, which would not necessarily be the same.


sfisher (approved):
Thanks for tracking down what was happening with this.  I imagine that it might be hard to figure out what was changing the dates.

PR #856, opened 2022-05-25
Validate URLs that to go to CloudFlare servers

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1742

CloudFlare wants to set cookies and run JS to ensure you are using a "real" browser. We don't want to do all of this just to validate a URL. We will assume that any URL encountering these hoops is a valid URL.


sfisher (approved):
Thanks, Ryan.  How annoying that cloudflare does this.

PR #857, opened 2022-05-27
Rake task to update Zenodo metadata

sfisher:
The basics:

- Does a join from stash_engine_identifiers to zenodo copies to get only items that have had something happen at Zenodo.
- Checks  & updates latest submission for identifier on finished items with a deposit id in the three categories of data/software/supplemental.
- sets submission type so relationships get set when generating metadata from zenodo libraries.
- For closed items (published) it re-opens, updates metadata and re-publishes to put them back to the published/closed state.
- For open items (not published), it just updates the metadata

### DIFF
--- a/lib/tasks/zenodo.rake
+++ b/lib/tasks/zenodo.rake
@@ -1,5 +1,6 @@
 require_relative 'zenodo/stats'
 require_relative 'zenodo/metadata'
+require 'byebug'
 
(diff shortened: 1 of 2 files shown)


ryscher:
Thanks for adding these scripts. I'm sure they will be useful for more than just the current updates.

PR #859, opened 2022-05-31
Remove dataset migration on login

ryscher:
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1316

Remove the UI that asked users to "migrate" their data, along with much of the underlying machinery that would be confusing for future development projects. I'm leaving the `migration_token` column in `stash_engine_users` in case it is needed for any future debugging, though I doubt it will be.

To test: 
- set the `migration_token` to null for a user
- (optionally) set other properties null for the user, such as `tenant_id`
- login as the user


sfisher (approved):
Thanks for removing this, Ryan. The code will be a lot cleaner without this in place.

PR #860, opened 2022-05-31
Update list of fee waiver countries

ryscher:
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1780

Update countries to correspond with the latest World Bank list of low and lower-middle income countries.

While the World Bank dictates the list, we match country names against organizations in ROR, so we use the ROR spelling. Hence it's not an easy one-to-one matchup, and some countries appear at a different alphabetical place than their World Bank names.


sfisher (approved):
Thanks for updating this list.

PR #861, opened 2022-06-01
Set frictionless to version that is merged with their latest "component" for UI

sfisher:
I merged into a separate repository that we're using for our customized fork of the frictionless code and made it available by NPM. This updates to use this version which works with the updated frictionless python libraries (4.38.0).  I did a few superficial tests and seems to work fine and also our tests pass.

Fixed some runtogetherwords in our error messages for frictionless.

Fixed a deprecation warning for rack attack.

Also related to https://github.com/CDL-Dryad/components/pull/5, so please look at and accept that one along with this, also.


ryscher:
Glad you got this to update!

PR #863, opened 2022-06-03
Disallow html uploads from URLs

sfisher:
The items coming back as 'text/html' are probably going to be missing dependencies and render poorly at best.  At worst and the most common case is that they are probably user errors (putting in a html page for their paper which should go in related works instead) or they are trying to use a file sharing service but they didn't put a good sharing URL or else the sharing service is giving us an HTML page that contains the file or is giving us a html page saying it couldn't scan the file for viruses because it is too big.


ryscher:
Looks good, tests fine for me. Thanks for blocking these.

PR #864, opened 2022-06-06
Fix bug in calculation of invoiced dataset size

ryscher:
When calculating the "current" size of a dataset for invoicing, we were accidentally including the size of files whose `file_state` was `deleted`. In practice, this rarely affected anything, because the curator usually versioned the dataset in a final pass before publication. However, in the rare case where a dataset incurs overage charges AND the user replaces a very large file in the final version of the dataset, the overage charges could be quite large due to counting both the new and deleted versions of the large file.

This PR corrects the calculation by only including files that are `created` or `copied` in the version that is invoiced.


sfisher (approved):
Good fix.  Thanks!

PR #865, opened 2022-06-06
Withdrawn stats update

ryscher:
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1786

Updates to statistics and reports that use the `withdrawn` status based on new shared understanding of the meaning.


sfisher (approved):
Thanks, Ryan.

It seems like a good change based on people's new understanding of what withdrawn means.  It seems like creating daily stats is complicated.

Do we feel like we have the statuses we need for reporting and are they easy enough to work with?

PR #866, opened 2022-06-06
Lower the rate for hitting NCBI

ryscher:
After running these processes manually, it appears that NCBI has become more strict in rate-limiting, so we should hit them slower.


sfisher (approved):
Thanks for making this more reliable so we don't get throttled.

PR #867, opened 2022-06-08
Allow superusers to change email addresses

ryscher:
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1788

On the Manage Users page, allow superusers to edit the email address for a user.


sfisher (approved):
Thanks, Ryan.

This looks good.  I tested changing my emails and works great. The other features in there to set the role and institution are really useful, also.

PR #868, opened 2022-06-09
Fix submission queue page

sfisher:
- Gets rid of the option to pause acceptance of submissions and just gives instructions
- Adds a checkbox to each line in the list and allows checking items to reset/re-enqueue for submission
  - Javascript to gather ids
  - controller resets items and always re-sends them automatically from whatever the current server is

To test manually, you can do something like:
1. Create submission with a URL on the internet and validate it so that URL goes into the database as a file.
2. Go into the database for your resource and modify the file URL in the database so that it is bad and will fail in Merritt.
3. Submit it to Merritt and wait for it to fail.
4. Go back into the database and correct the URL so that it is good again.
5. Go to the submission queue page and check the failed item in the list and click the button to resend it.
6. It should go through merritt now.

I think this should make re-sending/starting items easier after either freezing submissions for a while or else if something goes wrong and we have lots of errors.  Just check the ones you want to resend and click submit.

### DIFF
--- a/app/controllers/stash_engine/submission_queue_controller.rb
+++ b/app/controllers/stash_engine/submission_queue_controller.rb
@@ -40,24 +40,20 @@ module StashEngine
     end
     # rubocop:enable Metrics/AbcSize
 
-    # it is a little weird that these are GET instead of POST requests, but we may want to make it simple to use these URLs from curl or elsewhere
-    def graceful_shutdown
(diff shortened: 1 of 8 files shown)


ryscher:
Thanks, this is great. It already helped me test the processing on my new server as I was setting up the proxy through the FTP server.

PR #869, opened 2022-06-09
Limit schema.org description length and strip html

sfisher:
This fixes the warnings that Google gives us from the page.

https://search.google.com/test/rich-results  (and then put in the url to test)

Before it gave us a warning because of the length and the html.  I deployed to dev and tested with https://dryad-dev.cdlib.org/stash/dataset/doi:10.5072/FK28P64B5G and it didn't show errors after the changes (but did before).


ryscher:

PR #870, opened 2022-06-10
Fixes the caching of gems based on gemfile.lock

sfisher:
Sorry for the zillions of commits.  I had to keep making small changes and pushing to ensure that github actions was running against the pull request correctly.

We had a weird configuration before. and this uses different instructions from https://github.com/ruby/setup-ruby and updated versions.  There is a gem that does the gemfile caching properly these days and seems to work well.

I made a bunch of commits for hard to see problems in my yaml indentation.  😏

I thought maybe there was also a way to cache the precompiled assets so it didn't take so long, but there isn't much information about how to do this and they don't have something like the gemfile.lock that indicates when something has changed.  I guess you can recompile and hash it, but that is probably no faster than recompiling it by itself.

But at least the gems only need to be re-compiled and reinstalled when they change which saves us 2 or 3 minutes on most runs.


ryscher:
Thanks for working through this. I know it's a pain to keep waiting for the actions to run when you're debugging these changes.

PR #871, opened 2022-06-13
This adds the content urls for the item if files are public

sfisher:
Adds tests to be sure it's outputting url with the id in them for the data files.  Note: only outputs ContentUrls for datasets that have been published and have files available to download by everyone.

Also tested at https://search.google.com/test/rich-results to be sure the schema.org was being parsed and didn't have errors. Looks good.


ryscher:
Thanks, Scott. It's useful to have these added URLs in the metadata.

PR #873, opened 2022-06-14
Update to prevent possible sql injection

sfisher:
Only allow whitelisted fields and the two orders into ORDER clauses.

https://stackoverflow.com/questions/17859880/is-activerecords-order-method-vulnerable-to-sql-injection

Also didn't seem like there was a user permissions scope applied to the journal list.


ryscher:
Thanks for catching this!

PR #874, opened 2022-06-15
Updates to installation docs

ryscher:
Changes required to get the current codebase installed on a new machine.


sfisher (approved):
Thanks for updating all these again while going through the install again.  We haven't updated this in a while.

PR #875, opened 2022-06-15
Allows an individual API item to show the general catalog of what changed

sfisher:
This only outputs on items that have more than one embargoed/published version (the same way the UI does and compares those versions).

It doesn't show the changedFields for lists so that we don't have to load so many things in long lists and just the individual item view.

Some examples (using the dev database) that show the extra field:

http://localhost:3000/api/v2/datasets/doi:10.5072%2FFK28P64B5G
http://localhost:3000/api/v2/versions/3660

Some examples that are lists and don't:

http://localhost:3000/api/v2/datasets
http://localhost:3000/api/v2/datasets/doi:10.5072%2FFK28P64B5G/versions

### DIFF
--- a/app/controllers/stash_api/datasets_controller.rb
+++ b/app/controllers/stash_api/datasets_controller.rb
@@ -27,7 +27,7 @@ module StashApi
 
     # get /datasets/<id>
     def show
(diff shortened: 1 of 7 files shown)


ryscher:
Works great on my machine. Thanks for filling in this last missing piece of metadata.

PR #876, opened 2022-06-16
Fixes issues with subjects (keywords) entered in the Field of Science box

sfisher:
This fixes some weird keywords that could have been entered as free text in the field of science (fos) subject box.

In most cases, changing the field of science deletes previous ones, but for the non-standard ones that people typed by free association they become a custom subject that then couldn't be deleted easily.

There were also some problems when subjects were deleted leaving orphans in the join table between resource/subjects (though these didn't show or cause practical problems because of the inner join).

We can correct these orphans after we deploy with

```
DELETE sub_res.* FROM dcs_subjects_stash_engine_resources sub_res
LEFT JOIN dcs_subjects subj
ON sub_res.`subject_id` = subj.id
WHERE subj.id IS NULL;
```

This correctly removes the association to a subject rather than removing the subject (which can be available for more joins or autocompletes in the future).

### DIFF
--- a/app/controllers/stash_datacite/fos_subjects_controller.rb
+++ b/app/controllers/stash_datacite/fos_subjects_controller.rb
@@ -10,7 +10,9 @@ module StashDatacite
       respond_to do |format|
         format.json do
           # this removes the current associated fos subjects, but doesn't delete subject entries from subjects table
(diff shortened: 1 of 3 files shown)


ryscher:
Thanks for working through these complicated relationships.

PR #878, opened 2022-06-24
Lockout simultaneous submit

sfisher:
See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1820 about why this change and how someone was able to submit two versions (waiting on queue to set one of the states).

I also removed Merritt from the configuration file and hard coded it into the startup instead.  Seems kind of dumb to make it configurable like we're going to be switching between more than one storage repository frequently when we've never changed it in like 6 years.

PS. I attempted a bunch of updates to see if I could move github ubuntu version to 20.04, but it caused mysql problems and I reset my attempts and it will need to be a separate ticket that might take a while.

### DIFF
--- a/app/controllers/stash_datacite/resources_controller.rb
+++ b/app/controllers/stash_datacite/resources_controller.rb
@@ -64,7 +64,7 @@ module StashDatacite
       resource = StashEngine::Resource.find(resource_id)
       resource.identifier.update_search_words!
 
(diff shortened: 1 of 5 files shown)


ryscher:
This is useful. I hope it helps prevent future issues. Moving he Merritt info directly into the code is a nice simplification. Definitely no need to have this configurable until we actually add support for multiple backend repositories.

PR #880, opened 2022-06-27
Tracks state when changing "tabs" in preliminary info

sfisher:
See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1770

- Added code for tracking state across the "Preliminary info" "tabs."
- If people have added identifiers for manuscript or published article then default to that "tab" first.

To test that it retains settings for radio buttons:
- Start new dataset.
- Choose a different radio button choice.
- Navigate out and then re-enter dataset from "My Datasets" page and see that previously selected choice is still selected.

To test that it forces to the "tab" where an identifier is entered:
- Start a new dataset.
- Choose either of the two first choices and type in something for the identifier (MSID or DOI).
- Navigate back to "Other or not applicable"
- navigate out to the "my Datasets page" and then re-enter.
- Note that it forces to the first "tab" where you entered an identifier, even though you chose the last tab before.

To test that it now remembers the updated identifier you typed in after switching tabs.  (Before it would revert to showing the page-loaded value if you switched tabs, but saved correctly and it would show correctly if reloading the page)
- Start dataset
- Go to the first tab and enter a manuscript ID of your choice.
- Go to the second tab and enter a doi of your choice.
- Switch back to the first tab and note that it didn't revert the ID in the display and still has what you typed in.
- Switch to second tab and observe the same.

To test that the population still works correctly and I didn't break stuff you can try these items on development for auto-populating:

```
manuscript example:
Journal: Ecological Applications
MS number: 20-0939

DOI Article examples:
Journal of Ecology
10.1111/1365-2745.13903
```

### DIFF
--- a/app/javascript/react/components/MetadataEntry/PrelimArticle.js
+++ b/app/javascript/react/components/MetadataEntry/PrelimArticle.js
@@ -154,7 +154,10 @@ export default PrelimArticle;
 PrelimArticle.propTypes = {
   resourceId: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
   identifierId: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
-  publication_name: PropTypes.object.isRequired,
-  publication_issn: PropTypes.object.isRequired,
-  related_identifier: PropTypes.string.isRequired,
+  acText: PropTypes.string.isRequired,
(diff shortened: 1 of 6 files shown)


ryscher:
Great, thanks for cleaning this up. I know it's tricky to anticipate how people will want to use it.

PR #897, opened 2022-07-20
Changes logic to assign random curator rather than the oldest

sfisher:
Description above.  Should work so long as we keep the list of current curators accurate.


ryscher:
Looks good. This should be relatively rare, but I know Ari was getting overwhelmed with content.

PR #900, opened 2022-07-25
Previews a CSV from the file list

sfisher:
This does a range request and retrieves up to 2k of a CSV for preview in a special modal dialog box.

I added tests for the model to be sure it works when retrieving using webmock.  For the actual UI, it would be a lot harder to set up the tests, but I think it's going to be very obvious if it doesn't work.

To test, upload a CSV file and then click the magnifying glass icon next to it in the landing page.


ryscher:
Looks good. I worry that people will have a difficult time finding this. Hopefully, the upcoming site redesign will make it more prominent.

PR #901, opened 2022-07-26
Adds filter by tenant and tests

sfisher:
This seems to work and the sort by name already seems to be working.


ryscher:
Looks good, works well. Thanks for adding this!

PR #902, opened 2022-07-28
fix: ORCID link stuff not showing up for curators

sfisher:
Previously the special links only showed on published datasets for curators to give people invite links.

This gets passed through and now shows for curators on any that don't already have an ORCID and whatever state the dataset is in.  Turned out to be a simple change but just needed to trace few a few UI levels to figure out.


ryscher:
Looks great. I tested with datasets in several statuses, and the links all seem fine.

PR #903, opened 2022-07-28
Don't allow setting back to private for peer review after publishing

sfisher:
I guess some users were doing this after publishing and it doesn't really work right to do this, anyway, since the data will be out there on the internet about the dataset (in zenodo, in solr search, maybe in Google results and maybe referenced by other works).

Should avoid making something private or removing after being public unless good reason and not just at author's whim.


ryscher:
This is a very sensible change. One more protection against craziness!

PR #904, opened 2022-07-29
published back to PPR instructions

sfisher:
@ryscher  Can you look these instructions over to be sure they seem correct.  I think the other instructions may be fine in other circumstances (removing temporarily or completely or re-embargoing).

We're getting semi-frequent requests for this lately (like one or a couple a week). IDK reasoning on all of them, though the latest I saw seemed to be published on purpose by a curator and then something changed.


ryscher:
Looks great. Thanks for creating this more specific task list.

PR #906, opened 2022-08-01
Remove Daniella from our configuration

sfisher:
This removes daniella from emails and other things from our app.

Sorry, this has leftovers from a previous branch, but the diff of those should disappear if it is merged ahead of this PR.


ryscher:
Looks good, thanks for the updates!

PR #907, opened 2022-08-01
Environment cleanup for dev/stage

sfisher:
I tested this on development and removed extra crap.  It will not run on production (exits early if tried).

- The rake task `dev_ops:clean_datasets` deletes published datasets that do not have a corresponding file in Merritt (tests a single file from each dataset).
- Afterward, run the updated script `ClearData.clear_solr` which I needed to change since the old method no longer worked.  It clears all records from SOLR.
- Then run the rake task `rsolr:reindex` to repopulate the records for search.

This should make the things in the public UI work and get rid of most of the test data that didn't actually have files uploaded.

Some instructions from previous pull request should go away if other branch merged first.  I accidentally started my branch off that one instead of main.

### DIFF
--- a/lib/stash/indexer/solr_indexer.rb
+++ b/lib/stash/indexer/solr_indexer.rb
@@ -10,7 +10,7 @@ module Stash
 
       def initialize(solr_url:)
         # rsolr gives lots of other config options, but this is probably all we need for now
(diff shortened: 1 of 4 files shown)


ryscher:
Thanks for the cleanup. Note that Github doesn't always update the diff when overlapping PRs are merged. To fix this, you must "Edit" the PR to have a different base branch and then change it back. Supposedly this behavior is by design, but it's weird that they don't have a button to do this.

PR #908, opened 2022-08-02
Only show 5 lines of csv if it's not very wide

sfisher:
Only show 5 lines of the csv.  This uses a regex for end of line so we don't have to know if it's unix or windows line endings.  Splits and reconstitutes the preview lines as Unix file endings.


ryscher:
Looks great, thanks!

PR #909, opened 2022-08-02
Make buttons in modal have outline

sfisher:
I reused one of the button styles from the main stylesheet when buttons appear inside the modal dialog.


ryscher:
This is much better. Thanks for the last-minute cleanup.

PR #911, opened 2022-08-05
Improve automatic workflow actions caused by journal API calls

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1842

Fixing a couple of issues arising from interactions of the journal API and our automation:
- Always provide the name of an auto-assigned curator, so that if they don't pick it up immediately, we can still determine who was assigned.
- Don't allow journal API calls to withdraw a dataset that had been previously published.


sfisher (approved):
Thanks for updating this. The curators will be happy for weird cases like this one.

PR #912, opened 2022-08-08
Authorization Code grant type for OAuth2 and RSpace

sfisher:
The amount of code changed here, isn't huge, but it took a while to create and debug the changes and to be sure the login process worked correctly and work through documentation and a sample app.

- Doorkeeper config updated to allow authorization code type.  Had to add some user detection stuff in the resource_owner_authenticator.
- Added a new API require type `require_login_wo_tenant` that is useful for pages where someone is half-logged in (logged in but user account isn't complete since they haven't selected a Tenant yet). Had to do this to prevent some weird redirect loops in some circumstances where user had account but not tenant.
- This type of authorization puts the id of the user they get permission for attached to the `resource_owner_id` in `oauth_access_grants` and `oauth_access_tokens`.  This isn't the case in Client Credentials Grant.
- The `optional_api_user` is more complicated since in the case of Client Credentials Grant it's operating on the application owner. For Authorization Code Grant it's operating on the user that is `resource_owner_id ` for the access token/grant.
- If using the Authorization Code Grant, the user's system role isn't relevant and it's reset to `user` (but the user isn't saved, so just the case for the current request).  So a third party only gets the user-level permissions if proxying for a user that has approved.  (IE, if that user is a curator, the outside app doesn't get any special curator permissions from the Authorization Code Grant even if their other normal user permissions for that user might be Curator.)
- A bunch of samples and explanation.
- I added a test to the dataset API requests controller to ensure that a Authorization Code Grant isn't given extra superuser views into non-published versions, even when the proxy user for the grant is a superuser.

Let me know if you see problems.  While this isn't a huge amount of code changed, it's all new functionality and kind of tricky to test.

BTW, the other repo is at https://github.com/CDL-Dryad/api_test for test code.  The links from the readme are to the main branch here, so will not work until it's merged.  You can replace the branch name in the URLs if you want to make sure they work.


ryscher:
Looks good, and works well for me. The one thing that tripped me up is that the `redirect_uri` used in the client application must match the `redirect_uri` that Dryad has stored in the `oauth_applications` table. It would be good to add this to the documentation.

sfisher:
OK.  I updated the documentation.

ryscher:
Thanks for the updates!

PR #913, opened 2022-08-09
Fixes encoding display issues

sfisher:
This fixes the encoding issues by stripping out bad things that will cause rails to barf.

At first I tried a character detection library to convert more cleanly, but it just caused more issues than it fixed.

This also fixes a problem where some previews not limited to 5 lines with `/$/` regex.

I have a dataset on dev called "Testing lots of bad CSVs" which has about 20 csvs that the semrush crawler found and errored on.  They all work now.

It's a tiny PR but took a long time to find the right thing to do to make it reliable.


ryscher:
Thanks for making this more robust!

PR #915, opened 2022-08-19
Create merritt_stalled.md

sfisher:
Gives query to try to detect stalling in Merritt where it never appears in the OAI feed or their UI despite a success message from SWORD.


ryscher:
This is great to make us more aware of Merritt's current status. Thanks!

PR #916, opened 2022-08-23
Frictionless report api access at /api/v2/files/<id>/frictionlessReport

sfisher:
This allows a frictionless report to be viewed and set in the API for a file.

This is really for our own use since we likely won't have people setting Frictionless reports besides us.  It is also a niche thing for people to view since only tabular files will have a report and is more of our validation than something most people are excited to see.

Also, frictionless reports may be available for all file types (including those uploaded to zenodo) which we do not expose the file objects for other people's files through our API and only expose our own files (but reports can be seen or set for any file).


ryscher:
Looks great. Thanks for finding a relatively streamlined way to communicate with the Lambda process.

PR #917, opened 2022-08-24
Api token handling

sfisher:
You put in your application id and secret in the one row in the api tokens table and the class takes care of making sure your token for the same environment API is current and not expiring soon (re-get token as necessary and store in the row and update expiry time).  ApiToken.token


ryscher:
This is a sensible approach to the problem, and it gives us some insulation from Doorkeeper. Thanks for both the solution, and giving me some extra time to think about it.

PR #918, opened 2022-08-25
Zenodo logging to database

sfisher:
This passes the zenodo copy id for the database table so we can log the whole zenodo request/response cycle into our error_info field.  Also expands the size of the field for all the retries and crap that may be in there.

These are a lot of changes, but mostly finding the chain of calls to pass the id for the item through multiple laters and revisions of tests to include it.


ryscher:
Looks great. Sorry it was so much work adding this.

PR #921, opened 2022-08-31
Add curation note to log the user who edits a dataset

ryscher:
This should help us to avoid the confusion described in https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1950


sfisher:
That's OK with me to disable it.  Once in a while I feel it's helpful to do it the way it suggests but a lot of the time it doesn't make things better or might even make things worse.  I'm ok disabling it.

sfisher (approved):
Thanks for adding this editing note to the history.  This makes a lot of sense.

PR #927, opened 2022-09-15
This seems to fix the "more >" links in geoblacklight.

sfisher:
I had to try lots of different combinations to get the load order or combination of conflicting libraries right.  It wasn't correctly loading bootstrap for some "modal" thing they had but seems to now.


ryscher:
Looks good, works for me. Thanks for tracking this down!

PR #929, opened 2022-09-15
1978 fix small errors found while reviewing docs

ahamelers:
Typos & other small errors fixed in some docs


ryscher:
Wow, thanks for catching all of these issues! It's always great to have a fresh set of eyes on the pages.

PR #930, opened 2022-09-15
Closes #1967

ahamelers:


ryscher:
There is also a modification date in "Dryad Code of Conduct Working Group" that should be updated. I'm not sure whether both are strictly necessary, though.

ahamelers:
> There is also a modification date in "Dryad Code of Conduct Working Group" that should be updated. I'm not sure whether both are strictly necessary, though.

I missed that, and I think only the working group one should be updated! I'll fix this

PR #935, opened 2022-09-20
1974 add dryad contact page to website

ahamelers:
Closes #1974

- 🤞 that the facebook link is correct, someone who knows or has facebook, please check

- I added a small level of spam prevention/obfuscation for screen scraper bots on the email addresses for individuals.

### DIFF
--- a/app/assets/javascripts/stash_engine/ui.js
+++ b/app/assets/javascripts/stash_engine/ui.js
@@ -1660,4 +1660,13 @@ function joelsReady(){
   $('.js-pubdate__year1').text(year1);
   $('.js-pubdate__year1').attr('datetime', year1datetime);
 
+  var emails = document.getElementsByClassName('emailr')
+
+  for (var i=0; i < emails.length; i++) {
+      emails[i].onclick = e => {
+        var email = e.currentTarget.textContent.split('').reverse().join('')
+        e.currentTarget.href='mailto:'+email
+      }
+  };
+
 }// close joelsReady()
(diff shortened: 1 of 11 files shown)


ryscher:
Dryad's Facebook page has been unused since 2019, so I think we should omit the link. Otherwise, this is great, especially the email obfuscation.

ahamelers:
Okay, sounds good. I've removed the link. Not sure if there's some plan to expand/reinvigorate the Facebook page, but if that's the case we can always add it back in after that happens!

ahamelers:
The email address obfuscation can be reused anywhere 😄

sfisher:
This looks good, @ahamelers  .  Thanks for doing this.  I had forgotten that ui.js file existed (but it's fine to put things there).  You'll see a bunch more js files under separate sections (named like the controllers) that might have specific functionality to a part of the site, but I think they all get compiled into a giant ball of minified Javascript (that should be cached after the first load).

Joel (a UI design person) put in a bunch of those modernizr functions and some other things that are in there.  We'll also want to evaluate if they're needed anymore (I suspect most aren't needed anymore since they were put in probably 5+ years ago).

I don't think some parts of our site will really work with less than ES6, but when a lot of this stuff was first created, people might not have had a browser with ES6 yet in cases.  At this point, I think we can expect people to have a fairly up-to-date browser (and not Internet Explorer).  Plus these days, if they don't have an up-to-date browser they are exposing themselves to lots of vulnerabilities and lots of risk on the Internet and they have bigger problems than our site not working.

sfisher (approved):
Looks good.  Thanks Audrey.

PR #936, opened 2022-09-20
Make a role called limited_curator

sfisher:
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1949

Doesn't allow editing of dataset and locks out some other changes.  Mostly used for viewing things.

The way we were doing permissions was inconsistent, so tried to make most use the helpers we have defined for roles.

Ran into a few problems where I don't think things were right and ran into some testing problems which is why it took so long.  I think it's all working correctly now.


ryscher:
Looks good, and everything works as expected. We might want to deploy to dev/stage so Jess can verify.

PR #937, opened 2022-09-22
Remove extra period that can cause problems for the URL

ryscher:


sfisher (approved):
Thanks, Ryan.  Yeah, I think some email clients try to auto-link urls or else copy/paste can do something weird.

PR #941, opened 2022-09-27
Closes #2009, removal from #2005

ahamelers:
Also relocates Jess to the section with the other Heads


sfisher (approved):
Thanks for the updates for these.

PR #942, opened 2022-09-27
Move GMail authentication to the web app

ryscher:
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1939

Replace the outgoing command-line authentication with a web-based system. The new system performs nearly the same task, but with Google's preferred mechanisms.

To test on a new server:
- go to http://servername/stash/gmail_auth and follow the directions
- after successful authentication, check for the token file one directory *above* the server's root
- verify the token file with `rails journal_email:validate_gmail_connection`


sfisher:
Hi Ryan,

I tried this process on our dryad-dev.cdlib.org server and got the message

```
Access blocked: Dryad v2's request is invalid

You can't sign in because Dryad v2 sent an invalid request. . . .

If you are a developer of Dryad v2, see [error details](https://accounts.google.com/).
Error 400: redirect_uri_mismatch
```

I assume this is just because this isn't a registered callback redirect URL with OAuth2.  I also don't have the gmail account info (which is fine, I think if I just try it with my own account and then delete the credential info file afterward).

I think it looks OK, but wasn't really able to fully test.  I'm ok going ahead with it since it's not a user-facing page and if there are problems no one will see or care.

ryscher:
Sorry about that -- I just now added all of the redirect URIs. I hadn't added them before because I was re-creating the auth credentials for every test.

sfisher:
Huh.  I'm still seeing the problem, but I'm fine merging this since it's not user-facing.  I'll go ahead with the merge.  I'd think you'd need to add https://dryad-dev.cdlib.org (or maybe the full callback path to the google config for oauth).

It's certainly something we can work out during the coming sprint if it's any problem and user's will not experience it as a problem.

PR #943, opened 2022-09-27
Add separate GMail credentials for each server

ryscher:
Extension to #942, because Google won't supply a refresh token when you request access from a different server with the same credentials. Sigh.


sfisher (approved):
Thanks for the updates.

PR #944, opened 2022-09-27
Fixes for Merritt Sword cutting off at 66 minutes

sfisher:
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1999

This was tricky to troubleshoot and test.  In order to be sure this worked, I set the timeout in the http_helper.rb file to 1 second (instead of the 3600 seconds it lists below).  Then I went through the UI on my local machine and submitted an item.  This worked out for me in that it was long enough to submit to Merritt Sword and then timed out before it got a success response.

I was able to examine the repo_queue_states_table.  It stayed on "processing" until it got a "completed" response back from the completion by the OAI-PMH feed (from the normal dev server).

Also in the stash_engine_submission_logs it writes "Presumed eventual success -- check in OAI-PMH for completion" instead of a normal success message.

### DIFF
--- a/lib/stash/repo/repository.rb
+++ b/lib/stash/repo/repository.rb
@@ -77,6 +77,8 @@ module Stash
         self.class.update_repo_queue_state(resource_id: resource_id, state: 'enqueued')
         promise = submission_job.submit_async(executor: @executor)
         promise.on_success do |result|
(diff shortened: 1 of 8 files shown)


ryscher:
Thanks for getting this in at the end of the sprint!

PR #945, opened 2022-09-27
Remove editing internal data for limited curators

sfisher:
This is an add-on to the curation history that disallows the editing icons for internal data for limited_curators.


ryscher:
Looks great!

PR #946, opened 2022-09-28
Text updates

ahamelers:
Closes #1979 and #2011


sfisher (commented):
This looks good to me!

PR #950, opened 2022-09-29
Auto-merge users with duplicate orcid/email

ryscher:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1923

To test: 
- In `stash_engine_users`, set two users to have the same ORCID.
- Login with that ORCID.
- The login should succeed, and the two user accounts should be merged into one.


sfisher (approved):
Thanks for doing this, Ryan.

This should help out in those cases where they have an obvious duplicate.

PR #951, opened 2022-09-29
Fixing testing issues that appeared

sfisher:
They looked like WebDriver errors and there were many.

I updated gems to the extent I could.

I updated tests to use README.md instead of README.txt in most default data for tests.  I'm not sure why this didn't appear in earlier tests.


ryscher:
Thanks, Scott! I verified that the new gems work in my setup.

PR #952, opened 2022-09-29
Cedar audrey

ahamelers:
Sorry for all the noise from merging main into this branch 😄


sfisher:
Hi Audrey,

It looks like some tests have begun failing with all the changes so you'll want to look into the failures.

1.  The Rubocop tests should be fairly easy to fix by doing this from the application root `bundle exec rubocop -a`.  That will probably autofix many of the things that Rubocop is complaining about.  You might have to run it a few times to fix most things.  If there are still things it doesn't correct you can look at the lines it's indicating and read descriptions of the suggestion and either correct or if you think it not something that is really not worth doing you can enable/disable rubocop around an item with comments of `# rubocop:disable <category>/<item>` before and `# rubocop:enable <category>/<item>` afterward.

2. The other failures I'm now seeing are trickier, but there were some fixes for some tests in the main branch so be sure the origin/main branch is merged into your branch so you pick up any of those test fixes.

The feature tests run a full web browser that is remote controlled with Capybara/Selenium Webdriver and exercise the application and tests many features.  Occasionally there are timing or other issues that might be inconsistent.

I'm seeing that some of the items are now failing because of the TinyMCE Rich Text editor.  Perhaps something has changed the way it loads, the load order or it is delayed and not loading or something.  There is a helper to try to get it to behave at spec/support/helpers/tinymce_helper.rb which has been pretty reliable for 6 months or so but it looks like something has changed enough that it now sometimes fails because of whatever changes.

You can see what the browser is doing by running a test like this and it will pop it open so it is visible and run through tests (below).  (BTW, you can run all tests locally by doing `rails default` and that runs all tests.  To run tests locally you'll need a local test database on a local MySQL server with that `travis` user able to get in.  Travis was the name of the continuous integration service we used to use.)

Maybe something like this if you just want to run the tests for one file.
```
CAPY_SHOW=true bundle exec rspec spec/features/stash_datacite/new_dataset_spec.rb -fd
```

If you only want to run a single test then put a colon with the line number on the end like
```
CAPY_SHOW=true bundle exec rspec spec/features/stash_datacite/new_dataset_spec.rb:48 -fd
```

(you can leave off the -fd if you want but it shows each individual test by name and with colors so you can see what tests it's running instead of it just doing a dot for each test.)

If you put `byebug` in the middle of a test it will stop execution and you should be able to see what the web browser looks like at that point if you're using the "CAPY_SHOW" option to see what is happening.

Byebug pauses both the execution of the test and also the web server gets paused.

If you want a debugging console where you can test out Capybara commands and other things then I'd recommend pry-remote instead.  Info at https://github.com/Mon-Ouie/pry-remote .

- `require 'pry-remote'` at the top of the test
- put `binding.remote_pry` where you want a breakpoint
- in another terminal window or tab type `pry-remote` and hit enter when the breakpoint is encountered during the running of the test.

It will connect to the server remotely and not pause the web server (but just pause the test in the middle) and you can execute Capybara commands or other things one by one to automate the web browser or inspect variables or whatever.  When you exit from the remote session you opened with pry-remote (CTRL-D or exit) then the execution will continue for the remaining test(s).

Sorry you're running into some issues, but I hope this is helpful.

ahamelers:
Hi @sfisher , this is not a merge request to go to main, but to merge some work I've done back into @ryscher 's `cedar-base` branch. The rubocop errors are about comments and other work yet to be done in his cedar_controller, that I didn't mess with because the comments indicated he had an idea of work yet to be done there and I didn't want to mess up his process, and it also wasn't in the ticket he assigned to me 😄

ryscher:
Thanks Audrey! I'll merge this to `cedar-base`, and we can continue from there.

sfisher:
Sorry!  I didn't look carefully at the branches.  Still, maybe some of this troubleshooting will be helpful in the future.

ahamelers:
It definitely will!

PR #956, opened 2022-10-03
Closes #2012

ahamelers:


ryscher:
Looks good. Thanks for tracking down the issue here!

PR #957, opened 2022-10-03
Curation stats ppr_to_curation

ryscher:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1932

Adds a column to the Curation Stats that indicates the number of datasets transitioning from `peer_review` to `curation` status on a given day. After deploying this change, run `rails curation_stats:recalculate_all`


sfisher (approved):
Thanks for doing this, Ryan.

I tested the branch by updating the database and populating off our dev database.  We don't have lots of activity but it appeared to be working.

Looks good to me.

PR #959, opened 2022-10-04
Cedar audrey

ahamelers:
A couple of integral fixes to merge to the `cedar-base` branch:

- Loading error fixed
- Previously inserted metadata loads into opened editor reliably
- Updated time now included in default save route (I did the very easy version of this and just updated the templateInfo as well whenever auto saves occur)


ahamelers:
Added another commit that changes the dialog to a react component & only reloads and initializes the editor if the chosen template changes

ahamelers:
@ryscher I've added a cancel button but it needs some better definition of what it cancels—right now it resets any changes made since the page was loaded. At which points should the saved data become the SAVED data, and not be cancelable? When the save button is clicked only?

ryscher:
Thanks, Audrey! Merging this into the main Cedar branch.

PR #962, opened 2022-10-05
Fixing problems caused by Turbolinks in geoblacklight and subsequent pages

sfisher:
Turns off turbolinks so it doesn't keep the specialized geoblacklight css/javascript around for subsequent pages that might possibly be navigated to.

Also removes a stray backtick


ryscher:
Looks good, but I had to re-run the tests several times before they passed.

PR #949, opened 2022-09-28
Improve documentation for unpublishing datasets

ryscher:
- fix issue where the command-line process for Zenodo embargoes wasn't working
- remove confusing duplicate section in the documentation
- improve titles of the other sections to make it clearer when to use them

PR #963, opened 2022-10-05
More cleanup on zenodo embargo rake task

ryscher:
Additional cleanup associated with #949


sfisher (approved):
Thanks for updating the documentation.

PR #965, opened 2022-10-06
React reorganization

ahamelers:
- Reorganization of files
- Changes to linting, removal of individual file lint disabling & linting of code
- Some linting of CSS
- Refactoring of UploadFiles code


app/javascript/react/components/MetadataEntry/Authors.jsx
```diff
@@ -235,20 +233,13 @@ export default function Authors({
...
-            e.preventDefault();
-            addNewAuthor();
-          }}
+          type="button"
+          onClick={addNewAuthor}
         >
-          Add Author
-        </a>
-        {/* eslint-enable jsx-a11y/anchor-is-valid */}
+          Add author
```
sfisher, line 241: The reason for the failed test is that it's trying to click on "Add Author" and changing the failed test to "Add author" (with correct capitalization) will fix this test failure.

sfisher (commented):
Thanks for making all these changes and making things more consistent.  Everything looked good to me and I also did a little bit of manual testing just to be sure and everything looked to be working.

app/javascript/react/components/MetadataEntry/Authors.jsx
ahamelers: Ah, I caught this in the jest tests but didn't realize it would affect Rspec too. My first Rspec edits to make!

ryscher:
I remember some accessibility discussion about `<a role="button"` vs `<button`. @sfisher do you remember why we had these as anchors?

sfisher:
I think it's ok to change them to buttons as long as the styles and appearance remains intact.  I believe it's a legacy of some of the original designs.  Also while these don't really create any back-end state changes or cause http requests they do change some of the display forms (adding a new unsubmitted form to the page) so they have state-"lite" changes that are really just about appearance rather than data changes.

The old wisdom was that links were for GET requests that didn't change state outside of moving to a new page or refreshing display and buttons should be expected to be state changes (PUT/POST/DELETE requests) which did change state and were dangerous for crawlers and bots to do. With lots of Javascript and SPAs then it's less clear what is actually changing really changing state and what is just changing appearance and is more like "navigating."  It seems like many people put more into buttons even when they're not really updating state but only the display (similar to a link).

ahamelers:
My understanding of accessibility issues around links and buttons is that `<a href>` should be used for links and not for other actions. Using link tags for actions can cause confusion for non-sighted or visually impaired users, because doing so can cause unexpected behavior with both screen readers (which generally use the `a` tag to create lists of links) and similarly with keyboard triggers (which can be different for links vs buttons).

PR #968, opened 2022-10-07
Reduce our timeout for Merritt-SWORD

sfisher:
It looks like Merritt maybe reduced their timeout to 1 hour (from previous 66 minutes) at which point we now seem to be getting a "500 internal server error" with an HTML page response from SWORD. (About 3 this week in this condition.)

Reducing our timeout allows us to ignore a spurious error code for a submission that runs long (and which otherwise is generally just a submission which has gone asynchronous and is more of a notification of state change than an actual error).


ryscher:
Thanks for working through these issues!

PR #969, opened 2022-10-10
Closes #1655

ahamelers:
Adding 1000 max file limit to upload files


sfisher (approved):
Thanks for looking at this, Audrey.

It looks good and I tested it out by creating a bunch of dummy files and trying to add them all.

ahamelers:
You're much more dedicated than me @sfisher, I tested it by changing the maxFiles variable to 5 😆

PR #973, opened 2022-10-12
Frictionless bug

ahamelers:
Fix frictionless bug, re-enable frictionless alert close button, fix all console errors & warnings


ahamelers:
Also fixed some spacing, and made buttons, table rows, etc. sentence case.

sfisher (commented):
Thanks for all the hard work and the big bunch of cleanup.  This is useful going forward.  You even managed to find and cleanup some code from the recesses of time.

I found one very small oddness about the validation and the warning box with the X in the corner.  As long as the warning is present on page load then it can be closed with the X.  But if the validation process runs on the page and it's added to the page dynamically after validation then the X doesn't close it correctly.

It's a very minor thing.

PS. I can also give you a bunch of test tabular files I got random places from the internet if you want them (there are also some in our system with various levels of linting problems and sizes of files).

ahamelers:
> I found one very small oddness about the validation and the warning box with the X in the corner. As long as the warning is present on page load then it can be closed with the X. But if the validation process runs on the page and it's added to the page dynamically after validation then the X doesn't close it correctly.

That's a fascinatingly weird one. Is the other branch still running on staging with the old code? If so, it seems that this is an existing issue, since I can recreate the same problem there. I can still try to fix it, of course 😄 , or we can make a separate bug issue for that.

sfisher:
Yes, that's an existing issue which we can look at later and it's certainly not a critical thing.  I didn't realize it did that currently.

PR #974, opened 2022-10-12
CEDAR editor

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1805

Adds the CEDAR Embeddable Editor into the submission system. On the first page of the submission process, there is a control that allows users to select a metadata template and launch the editor. When the editor launches, it is placed in a modal dialog and initialized with our desired configuration options.

The editor itself lives in a Javascript file, which we store in our public directory. This file is not delivered as part of the normal Rails asset system, because it was greatly increasing the time to precompile the assets during the deployment process (10 minutes or more!)

Configuration: The editor will only appear when the `app_config.yml` references metadata templates. Currently, some sample templates are configured in the `development` environment, but they are not configured in any other environment. The base URL (`sampleTemplateLocationPrefix`) is currently set to the CEDAR servers, but it might be best to change this to our own servers when we have production-ready templates to use.

There is a standalone HTML page in `documentation/external_services/cedar_tester/`, which can be useful for testing the functionality of the CEDAR editor within a modal dialog. We might not need to keep this for the long term, but it's useful to keep until we have finalized all of the functionality.


sfisher (approved):
This certainly seems an OK state do a merge to me, especially considering it will not be configured for use for a while.  It will be easier to deal with and long separated branches are harder to keep up to date.  (I assume if we ever decide to rip it out then that won't be too hard.)

I know there are a number of outstanding things with the Cedar folks (including what useful templates they have ready).  I'll add a couple observations (which you've probably also noticed before) but we may want to log them into whatever information we're tracking if they're not there.

- It looks like skipping the authenticity token is happening right now, but would be nice for security to have it passed into the action.
- I found the [x] button in the corner randomly doesn't close the dialog.  It mostly does, but sometimes seems like it doesn't work.
- The old form shows briefly if you remove one and try filling in a different one (this is a pretty small thing and I doubt a lot of people will be switching forms like that).
- I didn't really track all that is going on with the integration.  You might be able to survive without formik and IDK if it would make things simpler since I think it's basically just an Open button and could just save whatever the json is directly without going through a form?  I'm not sure about this and would need to understand all that's going on to know.
- I got a couple of cancel buttons.  I think by closing and deleting and reopening a new form again.  I know the cancel button is a quandary since it saves as we go and it's kind of questionable, anyway or we don't know what it really does.
Screen Shot 2022-10-12 at 3 48 10 PM

It's also interesting to see what their forms do since I don't see much guidance, autocomplete, controlled vocabulary and select lists to guide and limit people to put in higher-quality data.  "Universal form generator" projects keep on coming (what happened to the XForms W3C spec and some similar things in the past?).  We'll need to evaluate our needs and the experience we want people to have with Dryad as we move towards a real release.

I think it's fine to merge what we have as a stepping stone with a bunch of stuff noted for the future.

I'd also defer to Audrey if she has feedback since she's more of a React and JS developer than I am.

PR #975, opened 2022-10-12
Add documentation about numbers for the annual report

ryscher:
Also cleans up the reporting and converts to sentence case.


sfisher (approved):
Thanks for the updates!

PR #976, opened 2022-10-12
Documentation about ruby app setup for our Frictionless Lambda

sfisher:
This gives instructions on how to test with it locally or setting up another environment that is configured to have a different public URL if needed.

I wasn't sure where to put this in our docs, but put it under the AWS services since it was related.  I can move it if you have suggestions of better places it should live.

### DIFF
--- a/documentation/external_services/amazon_aws_notes.md
+++ b/documentation/external_services/amazon_aws_notes.md
@@ -76,6 +76,17 @@ bottom of that page).
 ]
 ```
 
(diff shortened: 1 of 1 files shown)


ahamelers (commented):
Thanks for updating these docs! This looks easy to follow

ryscher:
Thanks, Scott. The information about new environments will be useful in other places (e.g., there are similar issues in needing a server name for #983), but I think it is fine to keep this information with the AWS notes for now.

PR #981, opened 2022-10-21
Copy 30-minute cron to 1 hour

sfisher:
I made a copy of this script from every_30.sh to every_hour.sh .  Both exist right now.  Then we will need to get Ashley to change our crons in puppet (or can we do it and where?) so we get rid of the 30 minute and replace it with an hourly trigger of the hourly script in cron.  Then we can do another PR to remove the every_30.sh script.

The only thing in the 30 minute script is the database backup right now so I don't think we need to keep it. If we want to keep the 30 minute cron we will want to empty the script of the command that runs the database backup.

### DIFF
--- a/cron/every_hour.sh
+++ b/cron/every_hour.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+: ${1:?"Need to pass in environment (e.g. development, stage, production)"}
+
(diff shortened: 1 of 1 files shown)


ryscher:
This is fine. I suggest we simply switch to a 1-hour cron. I don't see the need for both 1-hour and 30 minutes.

PR #982, opened 2022-10-23
Remove Author.primary in favor of Resource.owner_author

ryscher:
The `Author.primary` had not been updated when we introduced `Resource.owner_author`, so it was possible for the incorrect author to be selected when generating a Stripe invoice. This corrects the problem, and fixes the underlying cause for https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2053

### DIFF
--- a/app/models/stash_engine/author.rb
+++ b/app/models/stash_engine/author.rb
@@ -23,11 +23,6 @@ module StashEngine
       enable
     end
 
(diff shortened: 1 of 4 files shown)


ahamelers:
Looks good & no problems working. Thanks for doing this!

ahamelers (commented):
Looks great, thanks!

PR #983, opened 2022-10-25
Web-to-lead form for Salesforce integration

ryscher:
Adds a new page, `/stash/interested`, which collects information from people who are interested in partnering with Dryad. 

This page uses a Google reCAPTCHA. The "key" is stored in plaintext in the config, because it is equivalent to an account ID; it must be stored as plaintext in the resultant HTML. The secret associated with this key is not stored anywhere in Dryad; it is stored in Salesforce to validate the results of the CAPTCHA.

When the form is filled out, the user will be redirected to the `contact_thanks` page, but only if Dryad is running on a server with a defined Rails environment (dev, stage, prod). Other servers pass the servername "localhost" to the Salesforce instance, and Salesforce doesn't know where to send the redirect.

### DIFF
--- a/app/views/layouts/_contact_thanks.html.erb
+++ b/app/views/layouts/_contact_thanks.html.erb
@@ -1,0 +1,5 @@
+<h1>Thanks for Contacting Us</h1>
+
+<p>Thank you for contacting Dryad. We will be in touch with you shortly.</p>
+
+
--- a/app/views/layouts/_interested.html.erb
+++ b/app/views/layouts/_interested.html.erb
@@ -31,30 +31,30 @@
   value="ryan@datadryad.org">                               -->
 <!--  ----------------------------------------------------------------------  -->
 
-<label for="first_name" class="c-input__label">First name</label>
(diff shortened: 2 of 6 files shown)


ryscher:
@ahamelers This is done now. Unfortunately, I couldn't make all of the field names generic, so I had to put stupid ID numbers in the config file. I'm writing a ticket to improve this in the future, but the current implementation should be fine to get us started.

ahamelers:
One more! I think a bare minimum of form items (maybe "Email", "How would you like to participate") should be required and marked as required, and the submit button should be disabled until the required elements (and the captcha?) are complete and valid (i.e., the email address looks like an email address)

ahamelers:
>  It looks fine for an initial pass. Is this going to go on the contact page or linked from the contact page? I wasn't clear where it would get linked in or if it's really going in for public use or if this is just to allow Salesforce to continue testing.

I see that my required input suggestion is in the follow up ticket #2093

If this is for public use, I do think the need for required form elements should be moved up—if real people can submit blank forms, that's not helpful for anyone!—but if it's just meant for testing currently and won't be expost, it's fine to leave that for that next ticket.

PR #985, opened 2022-10-27
Cedar audrey

ahamelers:
Merged in main branch and deleted cancel button.

The very last commit is the code deleting the cancel button—the others are commits from merging cedar-base and main into this branch.


ryscher:
Thanks Audrey. I'm wondering whether we should automatically close the dialog a second or two after the Save button, but otherwise, this is good.

PR #987, opened 2022-10-27
1902 stripe fee waivers

ahamelers:
Closing ticket #1902. I believe this is complete, but I haven't been able to test the sending to the development stripe locally. If we get it loaded on dev, can someone tell me the fastest route to 1) act as a user with a waiver and 2) make an invoice?

### DIFF
--- a/app/models/stash_api/dataset_parser.rb
+++ b/app/models/stash_api/dataset_parser.rb
@@ -45,6 +45,7 @@ module StashApi
       TO_PARSE.each { |item| dynamic_parse(my_class: item) }
       @resource.identifier.payment_type = @hash['paymentType']
       @resource.identifier.payment_id = @hash['paymentId']
+      @resource.identifier.waiver_basis = @hash['waiverBasis']
       parse_internal_data
       @resource.identifier.save
       @resource.identifier
--- a/app/models/stash_engine/curation_activity.rb
+++ b/app/models/stash_engine/curation_activity.rb
@@ -318,7 +318,9 @@ module StashEngine
     def ready_for_payment?
       resource&.identifier&.reload
       APP_CONFIG.payments&.service == 'stripe' &&
-        (resource&.identifier&.payment_type.nil? || resource&.identifier&.payment_type == 'unknown') &&
+        (resource&.identifier&.payment_type.nil? ||
+          resource&.identifier&.payment_type == 'unknown' ||
+          resource&.identifier&.payment_type == 'waiver') &&
         (status == 'published' || status == 'embargoed')
     end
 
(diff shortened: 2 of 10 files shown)


ryscher:
It is a little tricky to test this locally, because it depends on having the `stash_engine_ror_orgs`. These can be imported using `bundle exec rails affiliation_import:update_ror_orgs`, but you need to have the same directory structure as the "real" machines.

I deployed on dev and tested like this:
1. Reset the tenant for my account to "dryad"
2. Create a new dataset
3. In the affiliation for my name, use an institution in a waiver country. I often use "Islamic University of Gaza".
4. Complete the submission of the dataset.
5. When Merritt finishes processing, go to the Admin Dashboard and publish the dataset (set the status first to "curation", then to "published".
6. Check the payment status in `stash_engine_identifiers` and the invoice in Stripe.

Currently, I am not seeing an invoice created. The payment status in the database is set to `waiver`, which I think makes sense, but it doesn't record the invoice number -- I wonder if we should store the invoice number instead of the waiver country.

ahamelers:
> The payment status in the database is set to `waiver`, which I think makes sense, but it doesn't record the invoice number -- I wonder if we should store the invoice number instead of the waiver country.

Excellent catch, it looks like there is a function that checks whether `payment_id` is nil and doesn't submit to stripe if it's got content and the `payment_type` is also not 'unknown'.

```ruby
    def ready_for_payment?
      resource&.identifier&.reload
      APP_CONFIG.payments&.service == 'stripe' &&
        (resource&.identifier&.payment_type.nil? || resource&.identifier&.payment_type == 'unknown') &&
        (status == 'published' || status == 'embargoed')
    end
```

So I've removed the record of the country from `payment_id`. Should it be recorded somewhere else? Also, the function to send the invoice overwrites the `payment_type`, changing it from 'waiver' to 'stripe'. Is this okay, or should I leave it as 'waiver', or change it to something like 'stripe-waiver'?

ryscher:
> So I've removed the record of the country from `payment_id`. Should it be recorded somewhere else? Also, the function to send the invoice overwrites the `payment_type`, changing it from 'waiver' to 'stripe'. Is this okay, or should I leave it as 'waiver', or change it to something like 'stripe-waiver'?

This is a question for Mark... I'll ask in the ticket.

ahamelers:
I added a new column to the identifiers table (as stated in the ticket). As part of the db migration I copied the current `payment_id` for any existing `payment_type == 'waiver'` to the new `waiver_basis` column, so it actually will appear in the shopping cart report! I also updated the shopping cart report to include this column. These will now appear there and in the database as **payment_type**: waiver, **payment_id**: stripe invoice #, **waiver_basis**: country name. The content of waiver_basis is meant to be expandable/alterable for other methods of granting waivers in the future.

ryscher:
Thanks for the updates! This all works as expected, and the resultant invoices have great detail. I'll merge this and ask Mark to confirm our test invoices.

PR #989, opened 2022-10-31
Update Salesforce API to use new Reason field

ryscher:
Partial fix for https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1991

As part of the updates to Salesforce, a new field was created, allowing curators to select multiple Reasons for creating a case. This update includes that field in the Dryad UI, but falls back to the older fields if the new field is not present.

### DIFF
--- a/lib/stash/salesforce.rb
+++ b/lib/stash/salesforce.rb
@@ -50,7 +50,7 @@ module Stash
     end
 
     def self.find_cases_by_doi(doi)
-      result = db_query("SELECT Id, Status, Reason, Case_Reason_Other__c FROM Case Where Subject like '%#{doi}%' " \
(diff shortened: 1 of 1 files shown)


sfisher (approved):
This looks good to me, Ryan.  I'm not sure exactly how to test but the changes seems ok.

Salesforce has a really weird API since it looks like those things are just SQL queries using a salesforce client.  I wouldn't have expected that.

ryscher:
It's definitely a weird API, and the SQL capabilities are fairly restricted, but they're super convenient.

I tested this on dev using my personal Salesforce instance. We don't currently have an "official" dev instance of Salesforce, but I think we can repurpose one of the instances that the consultants were using. I'll make a new ticket for that.

PR #990, opened 2022-11-01
Dataset funder dashboard see 1964

sfisher:
See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1964 and lots of comments on the ticket since the main text of ticket was more of an broad overview than a concrete/actionable ticket.

It sorts/filters/pages/autocompletes.  I added a request test since it's easier to directly test functionality by url params.  The sub-class report has the mega-query and things to make it work.  I hope this is a bit less ugly than the curation one since it doesn't do so many intermediate things.  It's a bit weird the way ActiveRecord can use any model to do any arbitrary query, but it's nicer than manually populating arbitrary fields and the fields returned become temporary active-record-like fields.  Doing this level of complexity of query in ActiveRecord is quite difficult.  We could think about having a view or other SQL things, but then we'd also have to manage that view and migrations for it and other stuff.

Most of the stuff that gloms onto the query is in the controller which is the place that has best access to all the page querystring parameters, so I think it makes sense there and I think the logic isn't bad to follow.


sfisher:
> A few small issues:
> 
> 1. At first, when I limited funders to "National Institutes of Health", I didn't see the entries that are funded by NIH institutes. I had to manually add the `name_identifier_id` to the appropriate `dcs_contributor_groupings`. Not sure when this was added, but we will need to be careful to ensure this is set up on all servers (and included in the install instructions).

This is added now in a rake task and added to setup information.  The information has been in the dev/stage/production database for a month or so already.

> 2. Since we're touching all the menus, it would be good to convert them to sentence case.

I ended up changing the menus as you suggested. 
- Then as Audrey suggested I needed to change all the pages.  So searched for `<h1>` `<h2>` and `<h3>` and the HTML titles to change all to title case on all pages since it's a lot easier than looking at each menu, manually finding the corresponding view view manually and navigating to 20 different pages.
- Sorry, I'm not going to sort through every HTML page on our site and proofread every page and also change all buttons, links and everything else I can possibly find.  Yeah, I "touched" most pages with titles and headings, but this is kind of crappy to force a cascade of editing everything on the site for a completely unrelated pull request.
- As predicted a bunch of tests failed so I had to go correct tests, but some are now sentence case and some are old title case (because they're buttons and etc) and it's hard to tell what it's supposed to be when editing tests.
- Also intermittent test failures with names generated by factorybot when it put apostrophe's and other things in names so finally figured out to HTMLEncode those entities on the intermittent failures when checking in the request test.
> 3. (very minor) The autocomplete list for Funder does not order the suggestions the same way as the Funder field in the submission system.
- I think this should be corrected in the controller when/if we stop using CrossRef Funders to fix the intermittent query problems.  It would be one place to do it in a single controller (that returns funders from our database and queries and sorts them in a useful way) rather than in Javascript which is different between some areas and also will have slightly different functionality.
- Also sorry, I'm not going to put this in React just for this one single thing on a page that doesn't really use it for anything else.  jquery UI is old, but it has the advantage of being much easier to use and this has accessibility stuff folded in. It also doesn't require some new dependencies which I don't think we need to add any more.

ryscher:
Thanks Scott! Adding the code/docs for inserting the correct NIH structure will help a lot in the future when we're bringing up new machines. 

Sorry that my comment about the menus ballooned so much. I was only expecting a quick update to the menus themselves and the associated tests. But it's great that we have these updates in place now.

ahamelers:
Sorry for all the work, Scott! I just meant we should have the headings and their pointers match as an argument for making the menus sentence case—not that you should do it all right now! After the various upgrades are done we can tackle sentence casing in a more systematic way.

PR #991, opened 2022-11-02
Remove bogus Kaust email

ryscher:
A user at Kaust was accidentally notified of some testing I was doing in the dev environment. This PR removes them from the dev config, so they are no longer notified of test activity.

Normally, I would add them to the production config, but this user seemed surprised to be getting any message, and I can find no record that they ever asked to be added to notifications, so it seems best to avoid adding new notifications.

### DIFF
--- a/config/tenants/kaust.yml
+++ b/config/tenants/kaust.yml
@@ -25,7 +25,7 @@ default: &default
     entity_id: https://waseet.kaust.edu.sa/idp/shibboleth
     entity_domain: waseet.kaust.edu.sa
   default_license: cc0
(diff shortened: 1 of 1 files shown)


sfisher (approved):
looks good.  We must've put it in when we were doing setup and someone gave us the email.

I agree, if the user doesn't want to get production notifications then no reason to add.  I think only a few institutions or people are interested to monitor the activity by emails on their Dryad account (from what I've seen recently).  Seems like the admin view is good enough for most people (if we set someone up as admin when adding).

PR #993, opened 2022-11-02
Testing to see if tests fail

sfisher:
And a couple of changes I had to make to get it working on my system with new ruby/bundler/rubocop versions.

### DIFF
--- a/.github/workflows/rspec_requests.yml
+++ b/.github/workflows/rspec_requests.yml
@@ -20,7 +20,7 @@ jobs:
 
     - name: 'Set Ruby and Bundler Versions'
       run: |
-        echo "RUBY_VERSION=2.6.6" >> $GITHUB_ENV
+        echo "RUBY_VERSION=2.7.6" >> $GITHUB_ENV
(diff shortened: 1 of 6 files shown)


ahamelers:
Thank you Scott! I didn't think to upgrade rubocop. I'll check out your changes on my system

ahamelers (approved):

.github/workflows/rspec_requests.yml
```diff
@@ -49,23 +49,6 @@ jobs:
...
-
-    - name: 'Install frictionless python package'
-      run: pip install frictionless==4.38.0
-
-    - name: 'Install frictionless python package for excel'
-      run: pip install frictionless[excel]==4.38.0
-
-    - name: 'Install frictionless python package for json'
-      run: pip install frictionless[json]==4.38.0
-
```
ahamelers, line 68: @sfisher Just double checking, did you mean to delete these lines? Will tests for frictionless still work without them?

sfisher: Yes, sorry, I should've commented that I noticed these extra things that weren't needed anymore while I was editing the github action tests.  We're no longer using the Frictionless stuff with a local Python environment.  I noticed as I was in there updating the Github stuff.  It doesn't really hurt to install all the python and libraries for frictionless but makes tests take longer to run.  I maybe should've put it in another PR but it was part of the same file I was editing already.

PR #994, opened 2022-11-04
error over 50GB replication to zenodo immediately (see #2097)

sfisher:
This errors immediately instead of taking up the queue slots for 24 hours before erroring on large submissions for data files to zenodo.

This should make less distractions and wasted time on our part until/unless we decide to put work and collaboration into fixing this.


ryscher:
Looks great. Hope this reduces the effort of managing the Zenodo connections.

PR #995, opened 2022-11-05
Workarounds for Merritt-Sword returning errors for non-error states

sfisher:
Something changed in the past month or so and now we get `500 Internal Server Error` when large items are ingested in Merritt, even though they succeed later.

David also said he'll send us 503s for some items for this case, also.  So adding that in.

By looking through the logs, I found out that these specific 500 errors seem to have java.net.SocketTimeoutException in the document body.  So I'm looking for this string to ignore the error. Other 500 errors can mean other things have gone wrong in other ways in Merritt so I don't really want to catch every one (so I'm re-raising others that don't have that string the body).

I hope this will make our life less annoying and so we don't have to babysit these constantly and make manual corrections all the time and be one less distraction.

### DIFF
--- a/stash/stash-merritt/lib/stash/merritt/sword_helper.rb
+++ b/stash/stash-merritt/lib/stash/merritt/sword_helper.rb
@@ -20,8 +20,12 @@ module Stash
         else
           do_create
         end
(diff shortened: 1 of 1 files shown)


ryscher:
Thanks, hope this improves the issue.

PR #996, opened 2022-11-07
PPR detail report

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2043

Detailed report of the times and contexts for datasets to be in PPR status.

Instead of one row per dataset, this report creates one row per PPR status, so we can focus on each time datasets enter and exit PPR.

To run: `rails identifiers:ppr_detail_report`


sfisher (approved):
This looks ok to me if it's good for Jess.

PR #997, opened 2022-11-08
Copy CEDAR config to stage environment

ryscher:
Allow the CEDAR editor to display in the stage environment, so the CEDAR team can see it and perform tests.


sfisher (approved):
This makes things appear in the stg environment for CEDAR for testing?  Looks good to me.

PR #999, opened 2022-11-09
Revise frictionless callback to make 404 errors when file is missing

sfisher:
I thought this was in place before but maybe it didn't work since the render of the 404 wasn't the last thing in the method?  I'm not sure why it didn't catch these.

- Moved render to the end of the method
- Added a test for a situation where the file or resource is now missing when the results of validation triggers the callback

This should get rid of the errors like below that we see in our email.  I believe people upload files that validate by Frictionless.  They get impatient and decide to delete them instead of waiting for validation to finish.  So when the callback from the Lambda fires with results the file is gone and can't update results.

```
A NoMethodError occurred in frictionless_reports#update:

  undefined method `resource' for nil:NilClass
Did you mean?  rescue
  app/controllers/stash_api/frictionless_reports_controller.rb:43:in `require_file'
```

### DIFF
--- a/app/controllers/stash_api/frictionless_reports_controller.rb
+++ b/app/controllers/stash_api/frictionless_reports_controller.rb
@@ -39,8 +39,8 @@ module StashApi
 
     def require_file
       @stash_file = StashEngine::GenericFile.where(id: params[:file_id]).first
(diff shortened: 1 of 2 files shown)


ryscher:
Looks good. I was not able to reproduce the problem by uploading and deleting files, but this can only improve it.

PR #1000, opened 2022-11-09
Get rid of code that redefines a constant

sfisher:
This was originally for testing a tenant, but we can just test that it creates a tenant and also test our test config values to be sure they're set up.

Audrey, I'll put this as a pull request against the main branch, but if you want it in your upgrade branch, also, you may be able to just merge this branch in to your branch manually (or cherry pick the two commits).

Also removed google login since we haven't use it in years.


ahamelers (approved):
Fantastic, thank you!

PR #1001, opened 2022-11-11
Sample beginning code to convert Excel to CSV

sfisher:
This converts the files in a directory and with the sample output.  This is just example starting code I needed to check feasibility so we'll need to build more in later.

We will also need to understand where in the workflow this happens and how users understand when new files suddenly appear and also gets scanned for errors with Frictionless afterwards and they get asked to correct all the CSVs they didn't upload.

None of this is put into any kind of Lambda yet which will take a bunch more work and I'll add tickets.  The lambda will also likely either need to interface directly with our S3 buckets or else callback to our own API to add files.  We'll have to work out the details for that later.


ryscher:
Thanks, Scott. I grabbed a random Excel file from Dryad and it worked perfectly.

PR #1004, opened 2022-11-15
Salesforce updater

ryscher:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1958

Automatically update Salesforce to reflect changes in Dryad.

This includes:
- Switches configuration to use the "partial" sandbox instance that includes real Dryad
  data, instead of Ryan's personal development instance.
- Adds `update_case_metadata` to update all relevant metadata fields in a
  Salesforce case
- Improves lookup of tenant/member institutions in Salesforce, trying the tenant RORs, but falling
  back to the tenant's name.
- Adds calls to automatically update the Salesforce metadata when curation status or active
  curator changes.
- Since the above changes would trigger Salesforce calls during rspec tests,
  adds the mock Salesforce calls to many more tests.

To truly test this, an account is needed in Salesforce. Previously, we had a very restricted set of licenses, but I think this may have improved. I will find out.


ryscher:
I have created a new Salesforce account that can be used for testing this functionality. The test username is `devs@datadryad.org.cloudparti` -- "cloudparti" indicates that this user accesses a sandbox instance, which contains a partial copy of our Salesforce data.

To test: 
1. Login to Dryad. Use the "test" login option, so you can assign yourself an ORCID that is assigned to the test user, "0000-0000-0000-0001". Select the role "superuser".
2. Go to the Activity Log for a dataset. 
3. Press the button to "Create salesforce case".
4. You will probably need to login to Salesforce at this point -- use the `devs@datadryad.org.cloudparti` username, and the corresponding password from the Dryad credentials file.
5. A case should be created with basic metadata copied from the dataset. Look it over, and leave this tab open.
6. Go back to Dryad, and change the dataset (title, journal, article DOI, curator, curation status, etc.)
7. If needed, wait for the dataset to process through Merritt
8. In Salesforce, refresh the page, and verify that the Dryad changes appear in the Salesforce case.

sfisher:
Thanks Ryan,

This looks like a lot more work that it seems especially since you had to update so many tests.  Sorry there wasn't an easier option.

I wasn't fully able to test since I still wasn't able to log in with that account and the passwords in the credentials files.

PR #1005, opened 2022-11-16
Ruby, rails, and node upgrade

ahamelers:
Upgrade code to Ruby 3.0 and Rails 6.1

Do not accept pull request until this branch is running and tested on dev and staging


sfisher:
This looks really good to me, @ahamelers .  There were a ton of changes but I've gone through a lot of the major functionality on dev and it seems good to me.  Feel free to put on the stage server next week (as far as I'm concerned and it sound like Ryan is fine with it also).

PR #1007, opened 2022-11-17
Static updates

ahamelers:
Closes #2107 and #2112. Also some work toward #2114


ryscher:
Thanks for cleaning all of this up, Audrey. 

I have a question about the capitalization of "Private for peer review". I was a little surprised by this change, since I consider it to be a proper noun, which I would expect to retain the capitals. On the updated "Join us" page, I see full capitalization for the proper nouns "Community Engagement" and "Publishing Service". Or would you think those latter examples should be changed as well?

sfisher (approved):
This look good to me.  Thanks for making these updates and I hope it wasn't too tedious.

ahamelers:
> Thanks for cleaning all of this up, Audrey.
> 
> I have a question about the capitalization of "Private for peer review". I was a little surprised by this change, since I consider it to be a proper noun, which I would expect to retain the capitals. On the updated "Join us" page, I see full capitalization for the proper nouns "Community Engagement" and "Publishing Service". Or would you think those latter examples should be changed as well?

I've not seen "Peer Review" cased as a proper noun—a quick search shows neither Elsevier nor Wikipedia nor anyone in between capitalize the phrase as a matter of course (unlike for example "Open Access" which seems to be more title cased than not). 

Personally I would make those "Community engagement" and "Publishing service" but I did generally just copy the new "Join us" page text from Jen's google document—I presume she has decided the Dryad section/department names should remain in title case? Which makes sense, to me, "Head of community engagement" on the staff page would look a bit sad (and what if we ever want to turn it into an acronym?!)

sfisher:
LMK if you think this is OK to merge or still would like to see those changes before we merge.  I'm just going through and merging the things that look ready.

ahamelers:
It's good with me, how about you @ryscher ?

ryscher:
Yes, all good. Merging.

PR #1009, opened 2022-11-18
Getting rid of the notifier and OAI & replacing with Merritt submission status

sfisher:
See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2067

This creates our own polling of the Merritt endpoint to check for status updates of submitted items indicating an item has gone through.

For it to accept that it's gone through then it needs to 1) get a response with info and 2) the response needs to include the Merritt version for this version of the submitted dataset.

To test on a machine or server:

- Run as a rake task like `RAILS_ENV=development bundle exec rails merritt_status:update`
- As long as it continues running it will keep checking for updates until it receives INT/TERM signals which can be sent by CTRL-C or `kill -15 <pid>`
- It outputs to file `<Rails.root>/log/merritt_status_updater.log`
- PID is written to file `<Rails.root>/log/merritt_status_updater.pid`
- It warns if PID already exists so you don't run more than one copy and makes you set the environment.

I ran this on dev by running/disowning and putting in the background and seems to be working.

I also added the state: `provisional_complete` as an option to the repo_queue_state table status.  This state means we got a "success" response from Merritt-SWORD (which we get with some datasets when SWORD thinks it's done but we don't get it for all datasets, especially large ones).  We don't really know that it has successfully gone into Merritt until the standard "complete" status.

You may want to tail the log to follow the output or status updates.


ryscher:
Looks great, works perfectly. Feel free to merge this whenever you're ready (not sure if you wanted to wait for Ashley's part)

PR #1013, opened 2022-11-21
Small style change to fix scrolling in modals

ahamelers:
Force break for modal header so no side scroll is added. Particularly for the tabular data check when the modal header might include a long file name (pictured).

Screenshot 2022-11-21 at 16 44 28


sfisher (approved):
Thanks for making this change.

PR #1015, opened 2022-11-22
make them uncompressed for zip version downloads

sfisher:
This should save Merritt some CPU and other resources in creating zips.

Tested on my machine and worked fine as far as I can see.


ryscher:
Thanks, this works for me. Hope it makes things easier on Merritt.

PR #1018, opened 2022-11-22
Update docs after VE upgrade

ryscher:
Docs associated with https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1824


sfisher (approved):
Looks good to me.  If we want curators to generally save somewhere else then we may want to look at defaults in the yml file that `script/file-download/download.rb` uses since I think it defaults to saving to the C drive under the user storage area instead.

PR #1023, opened 2022-11-29
Moves the pid file location

sfisher:
Thanks for catching this as a better location, Ashley.


ryscher:
This is great. Thanks, Scott!

PR #1024, opened 2022-12-02
NIH updater

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2007

Updates funder information by looking up grant records at NIH, parsing out the Institute or Center that led the funding, and updating the contributor record.

To test:
1. Go to the NIH Reporter tool and find a grant. https://reporter.nih.gov/
2. Create a dataset in Dryad with the funder name "National Institutes of Health" and the award number (project number) for your grant.
3. Run the tool: `rails identifiers:nih_funders_clean`
4. View the dataset and verify that the funder has been updated to the entity listed as the first "Funding IC" on the grant page.

Note that this code does not currently update records at Datacite. I'm wondering whether to include that process here, or run a broader update of Datacite.

### DIFF
--- a/lib/tasks/stash_engine_tasks.rake
+++ b/lib/tasks/stash_engine_tasks.rake
@@ -266,6 +266,8 @@ namespace :identifiers do
   task nih_funders_clean: :environment do
     # For each funder entry that is NIH
     StashEngine::Identifier.all.each do |i|
+      next if i.latest_resource.nil?
+
       i.latest_resource.contributors.each do |contrib|
         next unless contrib.contributor_name == 'National Institutes of Health'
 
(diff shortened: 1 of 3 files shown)


ahamelers (approved):
Great work Ryan, very neatly done.

lib/tasks/stash_engine_tasks.rake
```diff
@@ -261,6 +262,26 @@ namespace :identifiers do
     end
   end
 
+  desc 'Update NIH funder entry'
+  task nih_funders_clean: :environment do
+    # For each funder entry that is NIH
+    StashEngine::Identifier.all.each do |i|
+      i.latest_resource.contributors.each do |contrib|
```
sfisher, line 269: I had an error on the development database, maybe becaues fo some bad data.   When I added the line below just before this line then it fixed it.

```
next if i.latest_resource.nil?
```

sfisher (approved):
I guess maybe sometimes catching that nil is needed.  Otherwise it worked fine and mine updated.

Thanks for adding this.

lib/tasks/stash_engine_tasks.rake
ryscher: good catch -- updated

PR #1027, opened 2022-12-02
Should fix #2085 peer review not being notified

sfisher:
The main ones in that ticket from Jess were created in 2021 and the resource didn't have a `peer_review_end_date` filled in.

This allows notification if the `peer_review_end_date` has passed or if the `peer_review_end_date` is NULL in the database.  It still leaves the stuff about finding the CRON info in the history for finding if the last reminder is blank or if the last reminder was created more than a month ago.

So any items with blank peer_review_end_dates will be notified every month.  Does this seem like correct behavior, @ryscher ?  It seems like Jess would've like these notified, even if they have no end date.


ryscher:
Yes, this seems correct. Thanks for tracking down the issues!

PR #1031, opened 2022-12-05
Fixes geoblacklight by upgrading blacklight

ahamelers:
Upgrading blacklight alone seems to have done the trick! Upgrading geoblacklight doesn't seem necessary at this time as it causes more problems, if we want to do it it would take a little more effort (https://geoblacklight.org/docs/upgrading/Upgrading_to_version_4_0/).

There's also a change not mentioned in this documentation, which is that search_builder.rb would need to change from.

```ruby
class SearchBuilder < Blacklight::SearchBuilder
  include Blacklight::Solr::SearchBuilderBehavior
  include Geoblacklight::SpatialSearchBehavior
```

to

```ruby
class SearchBuilder < Blacklight::SearchBuilder
  include Blacklight::Solr::SearchBuilderBehavior
  include Geoblacklight::SuppressedRecordsSearchBehavior
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -56,7 +56,7 @@ gem 'amatch', '~> 0.4.0'
 gem 'amoeba', '~> 3.2.0'
 gem 'aws-sdk-lambda'
 gem 'aws-sdk-s3', '~> 1.113'
-gem 'blacklight', '~> 7.19.2' # do we really need this should be required by geoblacklight
+gem 'blacklight'
 gem 'bootsnap', require: false
 gem 'bootstrap', '~> 4.0'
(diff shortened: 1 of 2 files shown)


ahamelers:
I actually didn't update the lock file manually, but I did revert one update and do another one to update only the blacklight package specifically—some other gem updated with `bundle update` caused some of the tests to fail. But I must have messed up the commit trying to amend it! Thanks for fixing it.

ahamelers:
In future @sfisher it is fine to do additional commits on the same branch (or ask me to do them) so they're in the same pull request!

PR #1032, opened 2022-12-05
Updated gemfile for 2155-geoblacklight

sfisher:
It hadn't updated on my machine since it hadn't set a version in the Gemfile.lock for geoblacklight.

I updated it there with `bundle update` which updates the items that aren't locked in the Gemfile.


Gemfile.lock
```diff
@@ -432,13 +432,16 @@ GEM
...
     loofah (2.16.0)
       crass (~> 1.0.2)
       nokogiri (>= 1.5.9)
     lumberjack (1.2.8)
-    mail (2.7.1)
+    mail (2.8.0)
       mini_mime (>= 0.1.1)
+      net-imap
+      net-pop
+      net-smtp
```
ahamelers, line 444: It seems like some gem update is breaking tests. Based on the errors there's one of those keyword argument issues in this net-smtp gem. Is it just me, or is this lock file indicating that gem wasn't installed before?

ahamelers:
Okay, I think I've fixed this test—it looks like the way this kind of error works had previous functionality that was broken but is now fixed with the new upgrade. There was a very helpful comment in the test!

ahamelers (approved):
It seems like some gem update is breaking tests. (now fixed)

PR #1034, opened 2022-12-05
Fix the merge users action so the page refreshes correctly (#2156)

sfisher:
This was no longer using UJS because local: false wasn't set (used to be remote: true).


ahamelers (approved):
Thank you for catching this! I thought I got all these correctly, I'll have to double check that commit again.

PR #1036, opened 2022-12-06
Fixing constants and changes to ActiveRecord signatures for linkout #2157

sfisher:
This makes it easier to load all linkout files from a base module.  Also adds correct namespacing on based on the modules/classes.  Also changes the way `ActionView::Base.new` is called so it works with new signature.  Also gives full absolute path to a file which seems to be required now.

After making all these changes, I was able to run the rake task, though I'm not sure what all it does.


ahamelers (approved):
Adding the Stash:: to Linkout it a zeitwerk style change I should have caught, thank you for these fixes!

PR #1037, opened 2022-12-06
Revert mail/smtp because it's now broken after updates

sfisher:


ryscher:
Thanks for getting this working again!

PR #1038, opened 2022-12-07
Make assets serve by Puma from the public directory

sfisher:
For whatever reason, the previous config now doesn't work and setting `config.public_file_server.enabled = true` fixes the asset serving problem on the production server.

I tested this manually by changing it manually on the 01-prd server, restarting puma and testing directly against localhost:3000 on the server by using curl and wget to retrieve files.

Previous setting was getting 404s for the assets listed in the homepage.  New setting was getting 200s and retrieving the files.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -27,7 +27,7 @@ Rails.application.configure do
 
   # Disable serving static files from the `/public` folder by default since
   # Apache or NGINX already handles this.
-  config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
+  config.public_file_server.enabled = true
 
   # Compress CSS using a preprocessor.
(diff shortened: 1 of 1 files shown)


ahamelers (approved):
Great catch, thanks Scott

ryscher:
Thanks Scott!!! I don't see any way of testing this outside the production environment, so I'll merge it in.

sfisher:
Sounds good.

I tested on production 01 with stuff like this.  The asset was retrieved from the source code of the first web page.  It downloaded when I tried it.

```
wget http://localhost:3000/stash
wget http://localhost:3000/assets/stash_engine/application-b75c8ead445421dff833d9d6e7ba55ed1f80973fdecb66823b193c147675c0e8.css
```

PR #1041, opened 2022-12-09
Fix issue with peer_review stats that was counting reminder statuses

ryscher:
Fix to bug found by Bryan.

The curation stat "New Datasets to PPR" was massively overcounting, because the algorithm was including datasets that had also received an automated reminder email on that day. This caused the numbers to creep up over time, because we have datasets where reminders are continually ignored, so they re-send every month.


ahamelers:
Great catch, glad these numbers make more sense now!

PR #1042, opened 2022-12-09
Add more detail to curation notes when auto-changing status

ryscher:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2142


ahamelers:
Looks good, thank you Ryan

PR #1046, opened 2022-12-13
fee increase functionality

ryscher:
Allow fees to increase automatically on a set date. After a reasonable period of time (1 year?), we can move the "new" price to be the regular price.

The easiest way to test this is to set the cutoff date in the past, then create and publish a dataset. You should see the correct (old) price in both the submission system and the stripe invoice. Then change the cutoff date to the future and test again.


ahamelers (approved):
This is working for me, thanks Ryan!

PR #1047, opened 2022-12-14
Text updates associated with fee increase

ryscher:
Some more text updates from https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2082


ahamelers (approved):
Looks great, thanks Ryan!

PR #1050, opened 2022-12-16
Add record to submission notes when a user email is sent

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2153


ryscher:
When a message is sent, these notes can accidentally change the status, which is undesirable. We need some way to retain the status.

For example, after a submission is processed by Merritt, it should be in `submitted` status, but the action of adding the note pushed it back to `in_progress`:
Screenshot 2022-12-16 at 1 19 53 PM

ahamelers:
@ryscher , I've tried explicitly passing the status too, let's see if that solves the issue!

ryscher:
That fixed the problem. Thanks, Audrey!

PR #1051, opened 2022-12-16
Reduce duplicate email notifications sent to users for PPR status

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2083

### DIFF
--- a/app/models/stash_engine/curation_activity.rb
+++ b/app/models/stash_engine/curation_activity.rb
@@ -219,7 +219,12 @@ module StashEngine
       return if previously_published?
 
       case status
-      when 'published', 'embargoed', 'peer_review'
+      when 'published', 'embargoed'
+        StashEngine::UserMailer.status_change(resource, status).deliver_now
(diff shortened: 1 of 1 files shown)


ryscher:
It would be useful to get some examples of actual datasets where Jess wants the behavior to change. I made a test dataset and curated it. The act of curating creates a second version (resource), and the logic in this PR will still send an email once for each resource, so I received two emails for this dataset:
Screenshot 2022-12-16 at 1 41 57 PM

That said, in normal circumstances, curators shouldn't be touching things that are in PPR, so when we have multiple emails, it is the result of a curator workflow that we're not expecting. It would be helpful to analyze the sequence of statuses in actual datasets, so the logic has the best chance of working.

ahamelers:
I've converted this pull request to a draft. Hopefully it would be easy enough to change the code to check all previous versions (resources) as well, but I think you're right and it would be good to understand this need better first.

sfisher:
Looks good, but it sounds like we want to know some more about this use case before proceeding.

PR #1052, opened 2022-12-16
Automatic reporting on invoices that are voided in Stripe, but still exist in Dryad

ryscher:
Weekly check to see whether any invoices have been voided in Stripe, but still exist in Dryad. 

When issues are found, an email is sent to the same list as for submission errors. Note that I have updated the `local` environment to use the mailinator address rather than actually sending email to all developers. 

This includes an updated to use the most recent Stripe API. The API version is set explicitly in our code, so other tools that interact with Stripe may use a different version of the API. I set it directly in the class rather than a config file, because any change to the API version will likely affect the structure of the class. The largest effect of the version update is that the objects are now created in reverse order. In the old API, `InvoiceItems` were created first, and then an `Invoice` would collect all pending items for that customer automatically. In the new API, they take the more intuitive approach of creating an `Invoice` first and explicitly adding the `InvoiceItems` to it.

### DIFF
--- a/app/mailers/stash_engine/user_mailer.rb
+++ b/app/mailers/stash_engine/user_mailer.rb
@@ -107,6 +107,15 @@ module StashEngine
            subject: "#{rails_env} Failed to update Zenodo for #{@zen.identifier} for event type #{@zen.copy_type}")
     end
 
+    def voided_invoices(voided_identifier_list)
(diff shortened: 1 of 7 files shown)


sfisher (approved):
Thanks, Ryan.  I was a bit confused about some changes until I read about the changes in the new API version.  Looks good to me.

PR #1053, opened 2022-12-16
Add tenant for University of Rochester

ryscher:


sfisher (approved):
Thanks, Ryan.  I checked the logo and tenant on my local dev machine by changing my tenant_id.  Seems fine.  I also checked that I got to a login screen with shibboleth on stage.

Seems good.

PR #1054, opened 2022-12-16
Add detailed report of the instances each dataset is in_progress

ryscher:


sfisher (approved):
I'm approving since it doesn't affect the public, yet I'm confused if this really does what Jess wants?  Isn't the usual state following "in progress" "submitted" like the one below?

Screen Shot 2022-12-19 at 3 02 18 PM

If that's the case won't it always just be submitted?  And I think maybe Jess really wants to get the final state of the dataset or the state just before the next "in progress." Or maybe the next "published", "withdrawn" or "emargoed" or something like that?

I see that you give the final state for the dataset for the last state, also.

IDK, anyway, it seems to me like this will produce lots of "submitted" after lots of "in_progress" for most stuff since that is the normal following state.

PR #1055, opened 2022-12-17
Fix internal error that prevented curator from being assigned

ryscher:


sfisher (approved):
I'm confused about why it was adding to the string in the first place.  I guess that would fail if the note was nil but otherwise add to the string.

Seems OK to me, but IDK what it was trying to do and if it's supposed to be building up a string somewhere else or why it was that way originally.

PR #1056, opened 2022-12-19
Correct API response formatting issue

ryscher:
Possibly associated with https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2181, though I'm not completely certain until I get more information about the calls they are making.

Apparently, Rails 6 changed how the formats were processed in certain situations, so we must be more specific with the type of format in these situations.

### DIFF
--- a/app/controllers/stash_api/datasets_controller.rb
+++ b/app/controllers/stash_api/datasets_controller.rb
@@ -36,7 +36,7 @@ module StashApi
     # post /datasets
     def create
       respond_to do |format|
-        format.any do
+        format.json do
           dp = DatasetParser.new(hash: params['dataset'], id: nil, user: @user)
           @stash_identifier = dp.parse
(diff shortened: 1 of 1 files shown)


sfisher (commented):
I believe there is another instance of this on line 371, shouldn't this be changed, also?  I can try it and see.

sfisher:
At least the Rails 7 docs say that `.any` without any arguments should be for all of them.  Seems like this might be a bug in the rails version?

sfisher:
The other thing that is weird about this is that we have tests for these that exercise the controller `spec/requests/stash_api/datasets_controller_spec.rb` .  Those should also trigger the same behavior.

I believe they're doing something weird or else have some kinds of headers that trigger a special case or something else.  I wonder if they will let us know and supply a better code example since we have code examples in our tests hitting these controllers with normal types of requests and they worked ok.

PR #1057, opened 2022-12-19
Prevent scraping of emails

ahamelers:


sfisher (approved):
Thanks for keeping the evil bots at bay with the spamming.

PR #1058, opened 2022-12-20
Try upgrading tests on github that we're getting warnings about

sfisher:
I might have to make lots of attempts to get this working on github if there are problems.

### DIFF
--- a/.github/workflows/jest_tests.yml
+++ b/.github/workflows/jest_tests.yml
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v2
         with:
-          node-version: '16'
+          node-version: 16
       - name: Install yarn
         run: |
           yarn policies set-version 1.22.10
--- a/.github/workflows/jest_tests.yml
+++ b/.github/workflows/jest_tests.yml
@@ -4,13 +4,13 @@ on: pull_request
 
 jobs:
   build:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-22.04
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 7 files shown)


ahamelers:
Hi @sfisher, looking at the latest tests compared to some previous pull requests, the ubuntu warning seems to be resolved, but I'm still seeing this one:

> build
> Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/setup-node@v2

ahamelers:
I checked the documentation at https://github.com/actions/setup-node and noticed that the `node-version` there uses an integer, while our YAML files were using a string. I've changed this to integer for us as well, and will check if that fixes the remaining warning!

ahamelers (approved):
The node warnings are now gone! However, the Rspec feature tests have now failed a few times on`rspec ./spec/features/stash_engine/admin_spec.rb:196`, although it passes fine locally

sfisher:
Thanks for fixing that, @ahamelers .

We should keep an eye on this test and if it's repeated we may want to find another way to test it.  Looking at the test, I think there is some data setup outside the web browser directly in the database and then it loads pages in the web browser. In some cases there is some weirdness with capybara waiting for loads.

Alternately I see it's checking (the absense of) a curator name in the test.  If multiple curators are created for testing with `faker` then it's also possible that it's at random duplicating names for other curators.

Alternately, we can look at requests tests if the item is constrained to just testing one controller action (and optionally the rendering associated) which don't require the browser, run faster and seem to be quite a bit more reliable.

PR #1059, opened 2022-12-20
Add dropdown for refinement of funder information

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1993

The implementation is easily expandable for other collections of centers/institutes that should be refined down.

We can change the list of NIH institute names to make them more concise, but would need some more information or potentially testing to decide how a user would most often identify them.


sfisher (approved):
I feel pretty confident that this works.  IDK if there is value in adding testing for this.  We have a really basic rendering test at `spec/javascript/react/components/MetadataEntry/FunderAutocomplete.test.js` .  If not, not easy or useful then that's fine.

A couple of things I notice, which I don't feel have to be part of this pull request, but may be useful to think about or add.

The user can still pointedly ignore the sub-select list and can still submit with only "National Institutes of Health."  If we want to block them, then we should add additional validation to disallow final submission with this value.  This could be another ticket if we like.

I think if we end up adding some more of these main/sub-institutions in the future then it will become increasingly problematic to track and maintain it all if we have duplicated lists in several different places (code, database).

The database is tracking these groupings under `dcs_contributor_groupings` (funders are considered contributors of a certain type) and is used in a  other places.  We only have one grouping right now (NIH) and the `json_contains` field gives JSON for the sub-institutions.  I think you'll see the idea by looking at the fields in the table.

In the future, rather than hard-coding it could do lookups against main institutions to get sub-institutions from the DB.  I know this is a bit more annoying in React, yet we could add a RESTful route and controller for ContributorGroupings in Rails and do a HTTP GET in the React code to get the list so it's more DRY (Don't Repeat Yourself) and has a single source for the grouping lists.  I should've mentioned this as a possibility in the ticket but I think this was written early before some things were arranged and I didn't mention it.

I don't believe either of these things need to be done to accept this ticket, though, since it's better than what we have right now and adds good functionality.

ahamelers:
> The database is tracking these groupings under dcs_contributor_groupings (funders are considered contributors of a certain type) and is used in a other places. We only have one grouping right now (NIH) and the json_contains field gives JSON for the sub-institutions. I think you'll see the idea by looking at the fields in the table.

I didn't know it was tracking these groupings at all! It makes much more sense to use the database than a hand coded list. Looking at the database though, I do think we'd need to keep the names of the insitutes as recorded in CrossRef and the DB, and not use a shortening or acronyms.

PR #1062, opened 2022-12-20
Changes the call to render from within standard Ruby library

sfisher:
to use template rendering instead and also additional options required to do the raw render outside of Rails.  Checked in debugger and it is outputting contents of erb replacements.


ryscher:
Thanks for tracking this down! The files are rendering correctly now.

PR #1063, opened 2022-12-21
Re-adding instructions for unpublishing

sfisher:
since someone removed them.  It's not quite the same as private for peer review, I think.

@ahamelers I updated that instructions you had questions about.  It looks like you managed to find a branch or tag with old instructions that someone had deleted in our main branch.

I've re-put the temporarily "remove my dataset" back in since we do get asked for that sometimes and made some other small updates.  It's pretty similar to "private for peer review."

I'm honestly not super clear on the distinction these days.  I think originally both embargoed and "private for peer review" datasets always had an end-date set, but that might be different these days.  I guess "private for peer review" prevents them being published by curators somehow (or they know not to publish?).

Things that have not been published are also private (and reviewers can be given review links also).

I find the distinctions between these 3 states kind of confusing.  Maybe @ryscher  can clarify where I'm wrong or what the differences/similarities are and if the end-date is still used for PPR.

From my understanding:

- Embargo
  - Always time limited
  - Are already curated
  - Shows metadata but not files
  - It will be published after certain date
  - Private preview links work
- Private for Peer Review
  - Sometimes time limited, sometimes not (what circumstances?)
  - Curators don't look at them until it's "ready"
  - When time expires or else a journal gives an accept (?) notice then it's ready for curation and then curators look
  - Private preview links work
- Not published yet
  - Curators curate soon and aren't waiting on a journal acceptance in order to embargo or publish
  - Private preview links work for submitted versions

### DIFF
--- a/app/views/stash_datacite/licenses/_review.html.erb
+++ b/app/views/stash_datacite/licenses/_review.html.erb
@@ -64,7 +64,7 @@
 			publication of your dataset, you will receive an invoice for
 			&dollar;<%=  Stash::Payments::Invoicer.data_processing_charge(identifier: @resource.identifier) / 100 %> USD.
 			We’re sensitive to the fact that fees for individual
-			researchers are a burden and create inequities. If you'd
+			researchers can be a burden and create inequities. If you'd
 			like to request a fee waiver, please contact <a href="mailto:help@datadryad.org">help@datadryad.org</a>.
 		</p>
 
--- a/documentation/server_maintenance/troubleshooting.md
+++ b/documentation/server_maintenance/troubleshooting.md
@@ -295,7 +295,8 @@ See the instructions for "Private for peer review above" since it's almost the s
 changes.
 
 - In the step of setting "peer_review_end_date" set it to NULL instead.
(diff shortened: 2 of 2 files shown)


ahamelers (approved):
> I updated that instructions you had questions about. It looks like you managed to find a branch or tag with old instructions that someone had deleted in our main branch.

I see! I bookmarked the 'Troubleshooting' page when someone sent it to me, totally didn't notice it was some old version! Thanks @sfisher 

>I've re-put the temporarily "remove my dataset" back in since we do get asked for that sometimes and made some other small updates. It's pretty similar to "private for peer review." I'm honestly not super clear on the distinction these days.

Yes in the case I was dealing with, the resource had been published _unintentionally_ when it was meant to be being curated, I think that is a distinct case from wanting something in PPR.

ryscher:
Thanks for the updates, Scott. I had removed one of the sections in #949, because the two descriptions were nearly identical, and I could never decide which one to use. In practice, I think it makes little difference what status the dataset ends in, as long as it is no longer published. When we get a ticket for this type of removal, curators are actively updating the dataset, and they will change the status if needed.

PR #1065, opened 2022-12-22
Disable ability to set PPR in various circumstances and add explanatory text

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2039

Also tackled sentence casing in the submission page headings


sfisher (approved):
Looks good to me.  I think maybe the `@resource.submitted?` option may not ever be in that state when they're in the edit screen.  🤞   But it doesn't hurt to have it in there just in case.

I also apparently wasn't paying attention to the headings since I didn't realize some were still title case (I thought we'd changed them and a couple are still left).

PR #1066, opened 2023-01-03
Lookup institution subgroupings in database

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2197


sfisher (approved):
I tested this out and seemed to work fine for me.  Is the `group_label` so that you can put something like `NIH Directorate` in and it will show that for the heading instead of just `National Institutes of Health`?

We'll need to remember to populate the correct label into stage/production database when we deploy.

ahamelers:
>  Is the group_label so that you can put something like NIH Directorate in and it will show that for the heading instead of just National Institutes of Health?

@sfisher it's for the select instructional text—according to the mockup, 'NIH Institute or Center'. I should have made adding that text part of the database migration, but it's not too hard to add it to the database afterward

ahamelers:
I added it to the migration 😄

PR #1068, opened 2023-01-05
University of Washington tenant

sfisher:
Adding the university of washington tenant.


ryscher:
Looks good. Note that they use two domains: `washington.edu` and `uw.edu`. We should update the tenant_id for users in both domains when we deploy this.

PR #1069, opened 2023-01-05
Improve robustness of PPR calculation

ryscher:
The PPR statistics were thrown off by the changes in #1050. This adds a bit more checking to prevent double-counting of those messages, as well as any other changes that add repeated PPR messages to the activity log.


sfisher (approved):
Thanks for fixing this, Ryan.

PR #1070, opened 2023-01-06
API for receiving updates of status from external processors

sfisher:
We are running into this problem more and more that items are processed outside the main app.  It seems we will need a generic solution for tracking external processes and possibly for them to make callbacks to notify us of status/completion.

See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2109 .  I moved the rest of the lambda for Excel into another ticket and made this one just about the new API stuff since it took significant time and effort to add all the API actions, security, tests and new Swagger docs for this.

I also noticed some weird things about rails routing which I think I corrected (There was one set of routes repeated and also some things being more verbose than they needed).  Please double-check that those look correct and they weren't duplicated for some reasons I'm not aware of.


ryscher:
The duplicate routes were there because I couldn't get certain things to work any other way. For example: `/api/v2/versions/<id>` works in both the current code and the code with duplicates, but the list operation `/api/v2/versions` produces a routing error if the duplicate is not present. 

However, I'm not certain it is useful to even have `/api/v2/versions`. This URL is currently broken on our existing servers due to some controller issue, and nobody has complained, so maybe it is ok to leave it unrouted?

ryscher:
Regarding the Swagger updates: We have a number of "internal" API endpoints that we haven't added to Swagger, though they are mostly described in the `documentation` folder. We should probably discuss whether we want to document everything in both places, or whether the different documents are targeted at different audiences.

ryscher (approved):
This looks great. It solves the immediate problems while leaving a lot of flexibility for expansion. 

I'll let @sfisher decide whether to merge as-is, or if you want to make more changes to the duplicate routes.

sfisher:
Hi @ryscher 

I tested out the API routes and I didn't see anything that is broken with the duplicate removed.

I tried these on development and they worked.  

```
# list all versions for a dataset
http://localhost:3000/api/v2/datasets/doi:10.7959%2Fdryad.fttdz0c0/versions

# list all file objects for a version
http://localhost:3000/api/v2/versions/2997/files

# download a file for a dataset (by its file id)
http://localhost:3000/api/v2/files/8378/download

# get a frictionless report for a file (if it has one)
http://localhost:3000/api/v2/files/8378/frictionlessReport

# download a full version
http://localhost:3000/api/v2/versions/2997/download

# go directly to a version by its ID (resource_id)
http://localhost:3000/api/v2/versions/2997
```

I'm not sure we ever had something that listed all versions for all datasets and it just listed versions of a specific dataset.  (I don't think the duplicate route I removed would've done that.)

I think it's ok since I can go to all these URLs and the `versions_controller_spec.rb` also exercises them and makes sure the API urls return correct results.

ryscher:
Ok, great. I'm fine keeping it this way.

PR #1071, opened 2023-01-07
Removing leftover bits of OAI-PMH stuff (mostly monitoring)

sfisher:
Tried to find and destroy any leftover OAI-PMH feed stuff.  I think mostly the monitoring.


ryscher:
Thanks for cleaning this up. All works for me.

PR #1073, opened 2023-01-09
Prevent page clicking while form submits

ahamelers:
This solution can be used for other remote forms where submitting the form goes to a new page; forms where actions are completed on the same page will also need the class to be removed when the form action is complete.


sfisher (approved):
Thanks for fixing this.  Looks good to me.

PR #1076, opened 2023-01-11
Use owner_author when calculating waiver

ryscher:
Corrects bug in fee waiver calculation when the submitter is not the first author.


sfisher (approved):
Thanks for making these changes, Ryan.

PR #1077, opened 2023-01-12
Add error message and checkbox to constrain text entry fields

ahamelers:
- Constrain required autocomplete fields to items from the dropdown (i.e., items with IDs), unless the user explicitly clicks a checkbox indicating the text they need to enter is not in the list.
- Automatically match typed text to IDs from the list (without showing an error message)
- Remove the need to enter 'N/A' for no funding; change to a 'No funding received' checkbox.
- Sentence casing of form field labels


ahamelers:
I recommend we do not release this without doing some user testing, perhaps just having it tested by Jess and Bryan, since the checking and editing of these fields is one of the tasks of the curators

ryscher:
Deployed to stage for testing. 

Minor styling question: Is it possible to force the autocomplete and error message to be the same width as the input box? There is currently a little overhang.

ryscher:
I wasn't referring to wrapping nor the fact that the size adjusts. Rather, the fact that the boxes don't align. See below, the red area extends about 5 pixels to the right of the input box. (Again, it's relatively minor; I was just hoping it was something like a padding setting that could be quickly adjusted.

Screenshot 2023-01-17 at 9 47 11 AM

ryscher:
Looks like this can be changed in `_auto-complete.scss`, by replacing the two occurrences of 16 with 20.

ahamelers:
@ryscher , hmm, I think this may be an issue only on some browsers:

Screenshot 2023-01-17 at 17 48 24
Screenshot 2023-01-17 at 17 48 32

Which are you using?

ahamelers:
I've removed the question mark, as suggested by Scott, since this new error message replaces it. So the error box now lines up all the time.

ryscher:
Hmmm, I'm still having the alignment problem in Chrome, Firefox, and Safari. I wonder if we're getting slightly different question marks for some other reason, like locale settings?

sfisher:
I re-tested again and I think this is looking good.  If there are any minor problems, we can revisit, yet I think it's better to get this in and start getting better data.

PR #1078, opened 2023-01-17
Update waiver DPC calculation

ryscher:
When calculating waiver amounts, the system was using the default DPC instead of the calculated DPC for the specific dataset. Correcting that problem.


sfisher (approved):
Thank you for catching this, Ryan.

PR #1088, opened 2023-01-24
Reinstate  peer_review notifications to journals

ryscher:
While reviewing email notifications for https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2195, I realized that journals were no longer receiving notifications when datasets enter `peer_review` status. A review of the `automated_notifications` history revealed that this behavior changed on December 6.

The problem was introduced with the Rails 6 upgrade as a side effect of simplifying some conditional behavior, in commit d567ffd22ba09245c457ed468aa23eee5ebcc269. 

This PR reverses the change that introduced the problem.

### DIFF
--- a/app/models/stash_engine/curation_activity.rb
+++ b/app/models/stash_engine/curation_activity.rb
@@ -219,9 +219,12 @@ module StashEngine
       return if previously_published?
 
       case status
-      when 'published', 'embargoed', 'peer_review'
+      when 'published', 'embargoed'
         StashEngine::UserMailer.status_change(resource, status).deliver_now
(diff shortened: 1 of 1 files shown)


ahamelers:
Looking at the original commit, that was all stuff asked for by rubocop when rubocop was upgraded to handle the new version. So since your change back also passes rubocop checks, we're golden!

sfisher (approved):
Thanks Ryan.  I see why these were combined since reading casually it looked like the two blocks are the same but one is `journal_published...` and the other is `journal_review...` but both seem to be the same length and parameters superficially look the same.

PR #1089, opened 2023-01-24
Fix additional file in Merritt when users change filename case across versions

sfisher:
When a user uploaded identical filenames, but with different case *across multiple versions* then the previous file might not be deleted correctly from Merritt (zip downloads).  This has no effect if a user uploaded different cases of the same filename and deleted them within the same version.  Also a side effect is that it causes weirdness extracting zip files on some file systems with different case names. Some of the filesystems consider different cases to be unique files and some consider them to be the same file (and overwrite an earlier file with a later one if they consider them to be the same).

To manually test:

1. Create a few files with the same name but different case and unique contents so they're easy to tell apart.
2. Submit a dataset with one of the files, wait for it to go through.
3. Copy the Merritt URL from the `stash_engine_resources.download_url` change `/d/` to `/m/` in the URL and open in your browser.  You'll see the files in Merritt.
4. Edit the same dataset again.
5. Attempt uploading the file with different case but the same name.  You should get a message indicating that the file already exists. (this only appeared before if the files had identical filenames with the case being the same.)
6. Remove the existing file and upload the new file with the different case.
7. Submit the dataset and wait for it go through.
8. Check Merritt for the new version to be sure it appears correctly and the old filename is gone.

If you like, you can also check the stash_engine_generic_files table if you wish for the 2nd resource_id you submitted and you should see that it's creating a `delete` action in Merritt for the previous filename which differs from current filename that was added.


ryscher:
Thanks for tracking down this problem. In my tests, the protections aren't triggering. The UI still allows uploading the extra file, which overwrites the database entry for the similarly-named file and does not add a new delete. I'm not sure why this differs from your test results -- I confirmed on both my machine and dev.
Screenshot 2023-01-27 at 4 25 40 PM

sfisher:
@ryscher Can you re-check this for me?  I am not able to duplicate this on my computer in either Chrome or Firefox, dev or my local machine.  Maybe if you're taking steps you think are unusual, LMK.

Perhaps check that 1) webpacker is recompiling (if it doesn't recompile automatically, which is should when deploying on dev), 2) Press shift-reload on the file upload page to be sure cache is cleared of old Javascript files, 3) that puma is restarted on dev if redeploying there.

I just redeployed this to the dev machine, restarted puma and tested in firefox (with a hard page reload just to be sure) and it doesn't allow me. 

Screen Shot 2023-01-30 at 9 55 47 AM

Screen Shot 2023-01-30 at 10 13 57 AM

ahamelers:
I've tested this on dev and it's working for me as indicated by Scott 👍

ryscher:
Ahhh, I see now. I normally have webpack run whenever I restart rails, but it was failing due to a conflict with the node version, and I didn't see the output. Everything works now. Sorry for the confusion!

PR #1091, opened 2023-01-24
Update to github actions setup-node@v3

sfisher:
Like the title says.  I hope this resolves the node 12 warnings since we are setting version 16 in our config otherwise.


ryscher:
I don't see the warning in the tests for this PR, so I think this is fixed now.

PR #1093, opened 2023-01-30
Allow curators to revert metadata to a previous version

sfisher:
This should create a new version a curator can edit and the metadata comes from that version but the files need to be updated if they're not in the curator's ideal state, or they can kick it back to the user to correct the files if the user mangled them.

How to test:

1. create a version 1 of a dataset with wonderful metadata.  Submit the dataset.
2. create a version 2 of the dataset but pretend you're a user who goes through the dataset and fills all fields with "deleted" or something equally ridiculous to ruin the dataset metadata as much as possible.
3. As a curator, find the now-ridiculous dataset.  For the most authentic experience you would log in as a different user than the one that created the dataset.
4. Click the clock for the activity log and look at bottom of the page under "New version based on previous."  Select the version/date and click "<- New edit based on."
5. You should see a new version editing window open (as a curator if you used a different user than the dataset owner).  The metadata should be reverted to the previous version in the new versions you began editing--however the files will reflect the immediately previous version of the dataset.

Ideally a curator should enter information about why they reverted the dataset when they go to submit this new version (notes next to the submit button).

I also added a request test that tests the action of duplicating the dataset from an old version.  The dataset editing page is the same as the one the curator normally uses.

Only one in-progress version of a dataset is allowed at a time, as always.  We don't normally allow multiple or have a good way to reconcile conflicting edits by two people, so the locking mechanism.

The option will not be available if the user still has an in-progress version.  If a curator really needs to destroy an in-progress edit of a dataset they can click the "force edit" option and then discard the in-progress version.  Then they can create another version using this option if that is what they want.

### DIFF
--- a/app/controllers/stash_engine/metadata_entry_pages_controller.rb
+++ b/app/controllers/stash_engine/metadata_entry_pages_controller.rb
@@ -94,7 +94,7 @@ module StashEngine
       duplicate_resource
 
       # now fix the files based on last resource rather than the duplicated one, since we're not copying old files to S3
-      # from all the different services (Merritt, Zenodo)
+      # from all the different services (Merritt, Zenodo) which could be error-prone and take some time
       @new_res.generic_files.destroy_all
       prev_files.each do |f|
         next if f.deleted? # deleted so no longer exists here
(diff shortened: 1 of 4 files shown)


ryscher:
Thanks, Scott, this is great! Its nice to close up some of these loopholes that were causing problems for the curators.

PR #1095, opened 2023-01-31
Display of hierarchical sponsor names

ryscher:
This supports the hierarchical sponsorship of Wiley (https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2191)


sfisher (approved):
This looks good to me Ryan.  At first I was a bit confused by parent org but then realized it was in the journal_organizations model.  It all makes sense.

PR #1097, opened 2023-01-31
Fix display of autocomplete dropdown

ahamelers:
Small style fix for alignment of autocomplete dropdown (missed when removing the question mark display)


sfisher (approved):
Thanks for finding and fixing that.

PR #1098, opened 2023-01-31
Update at-a-glance stats for curation stats page

ryscher:
- Use 72 hours for "untouched" time instead of 48 hours, since curators often don't work weekends.
- Add a stat for the total number of datasets in the queue that have not been assigned to a curator (this is typically only "new" datasets, since versions of older datasets are almost always auto-assigned to a curator)


sfisher (approved):
Looks good to me.  Thank you, Ryan.

PR #1099, opened 2023-02-01
Fix for curation-at-a-glance

ryscher:
Extra fix for #1098. It counted correctly when the curators were explicitly unassigned. But when a curator has never been assigned, the submitter is still the `current_editor`, and we want to count these submissions as well (they're the most common!)


sfisher (approved):
Thanks for this update, Ryan.

PR #1100, opened 2023-02-02
Changing the list of low annd low middle income countries

sfisher:
- Niger wasn't listed before even though it is in the world bank list
- Nigeria is already present in low middle income
- Zambia went from Low middle to low
- Algeria wasn't listed before even though it is in the world bank list
- Lebanon went from upper middle to low middle
- Timor-Leste wasn't listed before even though it is in the world bank list
- West Bank and Gaza wasn't listed before even though it is in the world bank list

### DIFF
--- a/config/app_config.yml
+++ b/config/app_config.yml
@@ -51,6 +51,8 @@ defaults: &DEFAULTS
   merritt_max_submission_threads: 5
   crossref_base_url: https://api.crossref.org
   crossref_mailto: scott.fisher@ucop.edu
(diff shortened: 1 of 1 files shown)


ryscher:
Thanks, Scott. It's probably best to just put in the multiple versions of a name. In the past, ROR only had "West Bank and Gaza" listed as "Palestinian Territory", but that has changed now.

This is ready to merge whenever we determine the cause of the rspec errors.

PR #1102, opened 2023-02-03
prevent multiple email addresses in submission api

sfisher:
Also handles displayname <email@domain> style.

The regex for displayname is pretty loose (just looks for values with an at sign somewhere inside angle brackets) but I think it's good enough and a lot more readable than something extra complicated.

Also strips whitespace if multiple standard emails jammed into one.

Otherwise returns the input as pass through.


ryscher:
Thanks, Scott! This will help to reduce the amount of errors we see when publishers initialize datasets via the API.

PR #1104, opened 2023-02-08
Allow journals to have multiple ISSNs

ryscher:
Although most journals have a primary ISSN, there may be other ISSNs to distinguish print/online, to link the print/online versions, or to subsume ISSNs used by an older version of the journal. 

This update allows storing ISSNs either as a single string value (the old way) or an array of strings. No database migration is needed, because the current field already has enough space to store an array of 10 ISSNs.

This is changing the setting *only* for `Journal` objects. The `InternalData` objects will continue to use a single value, which will typically be the first ISSN found in a `Journal` object.

### DIFF
--- a/app/controllers/stash_api/datasets_controller.rb
+++ b/app/controllers/stash_api/datasets_controller.rb
@@ -491,19 +491,30 @@ module StashApi
     end
 
     def update_publication_issn(new_issn)
-      datum = StashEngine::InternalDatum.where(identifier_id: @stash_identifier.id, data_type: 'publicationISSN').first
+      idatum = StashEngine::InternalDatum.where(identifier_id: @stash_identifier.id, data_type: 'publicationISSN').first
       if new_issn.present?
+        # Ensure we have the standardized journal title and ISSN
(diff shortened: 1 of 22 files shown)


sfisher:
This is looking pretty good to me, Ryan.  What's the best way for me to go about testing this manually?

ryscher:
The best way to test is  to explicitly set ISSNs in the Rails console:
```
j=StashEngine::Journal.last
j.update(issn: [ "1234-5678", "2222-2222" ])
```
Then, you can test any actions with journals to ensure that either ISSN works correctly, including:
- Make a submission and select the journal from the autocomplete, then check that it has the correct title and (primary) ISSN in the Internal data.
- Create a dataset through the API and explicitly set the `publicationISSN`. The internal data should show the correct title and (primary) ISSN regardless of which ISSN you submitted.
- on the Activity log page, manually change the `publicationISSN`, and ensure the title is updated to the correct journal

sfisher:
Thanks Ryan.  This looks good to me.

PR #1105, opened 2023-02-08
Zenodo will not publish unchanged files

sfisher:
Sometimes people edit the files (software, supplemental) that they try to send to zenodo such as deleting files and then re-adding the same files again.  Zenodo will not accept these submissions as another version, even if metadata changed over time.

We don't really know until we publish again if the files at zenodo will change and we try to replicated each version's changes as they come in.  We only know upon publication if we can have a new version at zenodo or not.

If files haven't changed then we just revert all the latest copy "changes" to match the previous deposit at zenodo since we can't add another version.

To test:

1. Add some software (or supplemental) files and publish.
2. Make some more versions, such as not making up your mind and removing a file in one version and then re-adding the file you removed in another subsequent version.
3. Publish the dataset again.
4. It should finish successfully and if you look at zenodo after publication, it should have reverted the new deposition ID and software doi in the ZenodoCopies table to be the same as the previous one.  It should also have removed the in-progress version from the zenodo site.

This should save us from having to correct this problem manually when it happens a few times a month or so.

### DIFF
--- a/lib/stash/zenodo_replicate/zenodo_connection.rb
+++ b/lib/stash/zenodo_replicate/zenodo_connection.rb
@@ -26,7 +26,7 @@ module Stash
       def self.standard_request(method, url, **args)
         retries = 0
 
-        # the zenodo copy so we can log all the requests to the database
+        # the zenodo copy so we can log all the requests to the database, zd_id is just to enable logging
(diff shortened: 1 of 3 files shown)


ryscher:
Thanks for digging into this! It will be nice to reduce the number of errors we see.

PR #1110, opened 2023-02-13
Static page and mobile design update

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2271


sfisher (approved):
It looks good to me.  There are a couple of things I noticed compared to the previous version, though I don't know if they're good or bad just things I noticed were different.

New vs original.  It feels like the logos have more space below them than being vertically centered in the section above the divider line, though maybe it's not true and it's just that the section is now bigger so there is more space above/below.  Makes me wonder if the logos could be slightly bigger.

Screen Shot 2023-02-13 at 4 51 32 PM

Screen Shot 2023-02-13 at 4 51 45 PM

On the file upload, it seems like one of the file upload buttons is bigger than the other.

Screen Shot 2023-02-13 at 4 53 22 PM

Screen Shot 2023-02-13 at 4 53 31 PM


But overall these are minor things and I think the menus and pages fit better and things look nice.  IDK if they really need to be changed unless you or others think they do.

ahamelers:
> Makes me wonder if the logos could be slightly bigger.

I think technically they could; they're set to be a max of 40px height (same as before), we could increase that maximum—it's just a matter of how large they should be!

> On the file upload, it seems like one of the file upload buttons is bigger than the other.

We noticed this before and I forgot to note it down! I'll fix that.

ahamelers:
@sfisher both of these changes now made. I made the two buttons (one of which is a button while the other is a link) more precisely uniform in alignment etc., in addition to the same size, and I both reduced the header height slightly (it is still slightly larger than on the live site, to allow more of a clickable footprint for and space around each of the nav buttons) and increased the allowed size of the logos within that space.

sfisher:
Thanks for all this hard work.  I know how much work all these small changes are.  So many details.

PR #1111, opened 2023-02-13
Update combobox and remove use of bad function

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2289

### DIFF
--- a/app/javascript/react/components/MetadataEntry/GenericNameIdAutocomplete.jsx
+++ b/app/javascript/react/components/MetadataEntry/GenericNameIdAutocomplete.jsx
@@ -123,7 +123,7 @@ export default function GenericNameIdAutocomplete(
                 }
               },
               onChange: (e) => {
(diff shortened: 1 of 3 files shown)


sfisher (approved):
This looks OK to me and seems to work.

I can't say I understand why Downshift has onInputValueChange vs the standard onChange.  I guess from the docs it is for dynamically changing the list shown(?).  ( https://www.downshift-js.com/use-combobox/ ).

Anyway, the result seems better to me.  Thanks for updating it.

ahamelers:
I got the diagnosis of the issue and the solution from the comments on this bug report: https://github.com/downshift-js/downshift/issues/1108

In the future we can see about doing the work to upgrade to a more recent downshift version, where this bug should be fixed. Or I wouldn't be averse to scrapping this combobox and building our own!

PR #1112, opened 2023-02-13
Add CEDAR template for Templeton's neuroscience

ryscher:
Since this is just a draft template, it is enabled in dev and stage environments only. It won't appear on production, but we want it to be included in the codebase so it can continue to be tested on stage.


sfisher (approved):
Thanks for adding this in for them to test.

PR #1114, opened 2023-02-14
Tiered billing

ryscher:
Add a new `payment_plan_type` called `TIERED` that is billed based on usage tiers. 

In the regular code, the only change is allowing the `TIERED` value as an acceptable type of plan. In the reporting tools, there is a new report that is very similar to the `DEFERRED` report, but calculates a total billed value based on the usage tier instead of a flat rate.

Note that I am temporarily leaving the tier cutoffs and prices hardcoded in the reporting tool instead of putting them in a config, because the structure of the calculation may change slightly.

To test, change a journal to `TIERED`, change its associated datasets to have `payment_type = journal-TIERED`, and run the reports:
```
rails identifiers:shopping_cart_report YEAR_MONTH=2023-01
rails identifiers:tiered_journal_reports SC_REPORT=shopping_cart_report_2023-01.csv
```


sfisher (approved):
Thanks Ryan.  This all sounds very complicated.

PR #1119, opened 2023-02-21
Add author notification for datasets withdrawn by journal trigger

ahamelers:
I ended up triggering this based on the 'System' (user_id == 0) user creating the CurationActivity vs normal withdrawals, which are triggered by a curator. Hopefully that works well enough!


sfisher (approved):
As far as I know, triggering off the system user works ok.  If we have something else using that special user to do withdrawing in the future then we will need to change, yet I think this works for now.

ryscher:
I agree, this is a good solution to separate the emails.

PR #1123, opened 2023-02-23
New Merritt Deposit API

sfisher:
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2068

This removes SWORD from the deposit process and instead uses the Merritt deposit API which is always asynchronous.  (It should return acceptance right away but full processing doesn't happen for a while.)

You'll notice I removed a whole lot of classes and re-integrated this as `stash-deposit` gem instead of the `stash-sword` one.  I tried to clean the basic gem up and make it a lot simpler.

I still think there are some questions about the whole pluggable gem design for all these things which is a bit overkill for the way we actually use things and adds a lot of extra complexity.  However, I simplified this specific gem a lot without doing a giant rework of all the Merritt, queuing, promises and other pieces of code that do other parts.

I also left some of the configuration things alone (like the endpoint config).  See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2332 .  I think this should make it easier to revert to SWORD if something is catastrophically bad about the new Merritt endpoints.  Later we can convert SWORD endpoints to just the collection names.

To test:

- Add a dataset and submit it.  Ensure it goes through Merritt correctly.
- Update a dataset and submit updates.  Ensure updates go through and a new version is created.
- This shouldn't change much about our current queuing setup which we can also rework later since there may be less reason to continue using it when all Merritt submissions are now asynchronous.

I'll check with the Merritt people to see what they'd like to do for testing and if release on Tuesday is feasible. I'll mark this a draft PR until then.


sfisher:
@ryscher A quick update about a couple things I added (Friday afternoon)

- It will send out errors and mark things as an error if they get stuck in processing for longer than a day.  I'd rather know about error statuses, but Merritt doesn't seem to give them to us from their api. See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2338 so at least this is better than being blissfully unaware forever.  There were a couple things that had errors in our testing because ARKs in EZID expired (which wouldn't happen in production).  We should see error emails about them from the dev machine tomorrow (Saturday).  Also should show in log `/apps/dryad/apps/ui/current/log/merritt_status_updater.log` and turn to error state in our submission queue.
- I'm sure we will want to know some of the information about the submissions request we sent to Merritt, so I managed to trace it and return that info through about 5 different levels of methods and it's logged into stash_engine_submission_logs in the "archive_response" field.  It looks like below.

```
Submitted to Merritt for asynchronous completion
{"bat:batchState":{"xmlns:bat":"http://uc3.cdlib.org/ontology/mrt/ingest/batch","bat:submissionDate":"2023-02-24T15:26:48-08:00","bat:batchID":"bid-954b90eb-946b-422b-9205-d6118b4763c5","bat:userAgent":"dash_demo_user/Dash Demo User","bat:jobStates":{"bat:jobState":{"bat:localID":"doi:10.5072/FK2NP26C50","bat:jobID":"jid-7d813dec-214b-42cf-ade6-8594ca08588b","bat:retainTargetURL":false,"bat:packageName":"3860-manifest.checkm"}},"bat:batchStatus":"QUEUED"}}
```

ryscher (approved):
Thanks, this all looks great, Scott. I tested many different submission scenarios, and it all worked properly. It will be great to get his changed over.

PR #1124, opened 2023-02-28
2045 February gem/npm library security updates

sfisher:
I did the yarn and bundler updates.

I tested through a submission and looked at landing pages.  I looked at search.  The search "explore" is off but it was like that in the main branch so that change came in from some other branch, I believe.

I believe it all still works with updates.


ahamelers:
This is all working fine locally and in tests, I can't see any problems. Thanks Scott!

ahamelers (approved):

ryscher:
Looks good to me. Thanks!!!

PR #1127, opened 2023-03-03
Adding migration to store container_contents as json

sfisher:
This is a simple change.  There is a JSON database field type for container_contents on generic file model.  I didn't think it was worth creating a new table just for this one new field.

I attempted to follow the pattern at https://www.keypup.io/blog/embedded-associations-in-rails-using-json-fields which seems like a nice way to ensure better data consistency.  I had to delete the extra classes and revert since it doesn't work with our models, maybe because we're using Single Table Inheritance (and we also have nil values some places and not every file has this set).

Data would be like:

```
[
  {
    "name": "karen.xls",
    "size": 12355
  },
  {
    "name": "scan/head.fasta",
    "size": 3873837833
  }
]
```

### DIFF
--- a/db/migrate/20230303003530_update_generic_files_with_container_contents_json.rb
+++ b/db/migrate/20230303003530_update_generic_files_with_container_contents_json.rb
@@ -1,0 +1,5 @@
+class UpdateGenericFilesWithContainerContentsJson < ActiveRecord::Migration[6.1]
+  def change
+    add_column :stash_engine_generic_files, :container_contents, :json
+  end
+end


ahamelers (approved):
This pull request looks good, however  re: your comment, I think the format of the JSON should be a bit different (path instead of name, add mimeType for convenience, even if it's mainly derived from the file extension)

sfisher:
Thanks, @ahamelers, we can put in name as path instead if you prefer.  I'm sure we can guess a mimetype based on the file extension if we like.  I don't think anyone will be using it for any kind of download and file handling since we don't offer direct downloads from within compressed files but it may become useful for some other purpose in the future.  We can use `application/octet-stream` for anything we can't figure which downloads an item as a generic binary transfer and the user needs to decide what program opens the file (which we hope they know if they're in the research domain and have an idea how to use specialist data).

PR #1128, opened 2023-03-08
Fix problem with releasing embargo on the wrong version

ryscher:
Only release embargoes on the most recently submitted version of a dataset.


sfisher (approved):
Looks ok to me.

PR #1130, opened 2023-03-10
Getting contents of zips without downloading all

sfisher:
See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2350 .

Basic usage from rails console:

```
# load file object that is a zip file from database
df = StashEngine::DataFile.find(14121)

# create new ZipInfo and initialize with the presigned_url from file above
zip_info = Stash::Compressed::ZipInfo.new(presigned_url: df.merritt_s3_presigned_url

# get the file entries
zip_info.file_entries
```

- Handles zip32 and zip64
- Should raise errors for problems like not finding signatures or HTTP issues
- Lots of tests using fixture data & relevant bits from a ~5.6GB zip64 file (just saved parts needed for mocks)
- Range requests are barrel of fun
- Lots of people on the internet had partial solutions but were often doing stuff wrong, see the notes in the file which give some good resources and information if you're curious)


ryscher:
Thanks, Scott! Looks like this was a lot of complexity to work through. This works for all of the files I tested. Sorry none of the libraries actually did the job.

PR #1132, opened 2023-03-10
Admin table & CSV report improvements

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2359


ryscher:
This looks good, and I'll go ahead and merge it. 

One minor concern: I think Mark and Jess might have processes that depend on the order of the fields in these reports, so I can warn them that we changed the order slightly. I assume they will need to adjust their processes anyway to use the new fields.

PR #1134, opened 2023-03-10
Update_faq_How_much

jeremywalter:
added text regarding cost waivers to FAQ


ryscher:
Thanks, Jeremy. Note that the link to the World Bank site will need to be updated every year; I'll add a note to our annual task for updating the list of countries.

PR #1135, opened 2023-03-11
Changing the user, key and secret for UI access to AWS.  

sfisher:
Only allows access to the staging area S3 buckets and specific lambdas which are called from the UI.

You'll notice I created a new user and new policies (`s3-ui-staged-content` and `lambda-ui-triggered`) to give fairly minimal permissions for the things I thought might be required.

To test:

- Does file uploading work in S3?
- Are files removed after upload completes?
- Are lambdas still triggered for frictionless?
- Does anything look wrong with these policies on AWS IAM?

(PS.  I tested on my local dev and also on dryad-dev.)


ryscher:
Thanks, Scott. The new permissions all look good to me, and I verified the basic functionality.

PR #1137, opened 2023-03-13
Can you give me UC stats now?

sfisher:
Adding queries from request.   Both totals and by year.  By tenant and also by associated authors.


ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2376

PR #1138, opened 2023-03-14
Prefill journal info and import manuscript using URL parameters

ryscher:
I thought this issue had been fixed. Instead, the feature had never been finished, and the many tickets filed over the years had been removed from the board with the assumption that it was working. Grrrr.

Related tickets include:
- https://github.com/CDL-Dryad/dryad-product-roadmap/issues/456
- https://github.com/CDL-Dryad/dryad-product-roadmap/issues/534
- https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1815
- https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2230
- https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2231

To test:
- Find a Manuscript in the database. If you don't have one, it's best to copy the metadata hash from another instance of Dryad. Otherwise you will need to follow the complicated process of creating and processing an email as described in https://github.com/CDL-Dryad/dryad-app/blob/main/documentation/apis/journals.md
- Determine the Journal associated with the Manuscript and its `journal_code`
- Construct a URL that looks like this: `http://<server_name>/submit?journalID=<journal_code>&manu=<manuscript_number>`
- When you resolve the URL, it should create a new dataset, fill in the journal's name, and prefill any metadata from the Manuscript object.

### DIFF
--- a/app/controllers/stash_engine/resources_controller.rb
+++ b/app/controllers/stash_engine/resources_controller.rb
@@ -56,6 +56,7 @@ module StashEngine
       resource.identifier_id = db_id_obj.id
       resource.save
       resource.fill_blank_author!
+      import_manuscript_using_params(resource) if params['journalID']
       redirect_to stash_url_helpers.metadata_entry_pages_find_or_create_path(resource_id: resource.id)
       # TODO: stop this bad practice of catching a way overly broad error it needs to be specific
     rescue StandardError => e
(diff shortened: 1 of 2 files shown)


sfisher (approved):
Thank you, Ryan.

I've tested this out on development and it seems good to me.  I was able to import a manuscript with it.

PR #1139, opened 2023-03-14
Publish and use updated datacite-mapping gem

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2234


sfisher (approved):
Thank you so much for doing this, Audrey.  I see the gem is available on RubyGems and thanks for the updates here, also.

PR #1141, opened 2023-03-15
Reduce frequency of DB backup checker

ryscher:
We changed database backups from every 30 minutes to every hour in #981. This aligns the status checker with that change, only throwing an alert if the last backup is more than 90 minutes old.


ahamelers:
Great, thanks Ryan! Can we push this into production right away, or will we see these alerts for 2 weeks?

ryscher:
I have made "temporary" changes to production, so you won't see the alerts again unless the servers update for some reason.

PR #1142, opened 2023-03-15
Event data checkers minor problems

sfisher:
See description at https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2378

Making class and underscore case match everywhere for `event_data_citation` .

Removing a couple checks that are either not useful anymore or that we don't currently use directly (Counter completion in 5 days and DataCite EventData service).

Removing these items from the database and rake task that populates the DB keeps them from running.  They are still in some of the classes in case we re-instate later.


ahamelers (approved):
This looks good, thanks Scott. It looks like the rake task destroys the table before inserting the dependencies over again, which is convenient for removing these rows.

ryscher:
Looks good, thanks!

PR #1146, opened 2023-03-21
2351 extract file info from tar.gz

sfisher:
This extracts info from tar.gz while streaming, some retries and adds some tests.

See example in code or in tests for how to try this.


ryscher:
Thanks, this looks great, and works well with my tests.

PR #1147, opened 2023-03-21
Update Salesforce credentials and error handling

ryscher:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2383, part of the solution for https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2120

- Explicitly sets `@sf_client=nil` when there is an authentication error
- On the Activity log screen, omit the Salesforce section if there was an issue logging in
- Update credentials to separate new Dryad System account from Ryan's old account
- Correct issue with credentials in staging environment


app/views/stash_engine/admin_datasets/activity_log.html.erb
```diff
@@ -44,7 +44,7 @@
   </div>
 </div>
 
-<% if current_user.limited_curator? || current_user.role == 'admin' %>
+<% if Stash::Salesforce::sf_user && (current_user.limited_curator? || current_user.role == 'admin') %>
```
sfisher, line 47: I think this should be `Stash::Salesforce.sf_user` instead, right?

sfisher (commented):
Looks pretty good aside from one typeo, I think.  The credentials file will probably have a merge conflict later which is annoying to resolve since I modified it from main, also.  🤷‍♂️

app/views/stash_engine/admin_datasets/activity_log.html.erb
ryscher: Odd. Yes, it's a typo, but it works. Somehow, Rails allows this construction.

I'll fix it anyway, because it's very confusing.

ahamelers:
Thank you Ryan, and great typo catch Scott!

PR #1148, opened 2023-03-24
Rake task and revisions to schema/structure

sfisher:
- introduces ContainerFile class/db schema which is information about a file within a container.  Foreign key is the data_file_id which is the parent container.
  - Zenodo has their own interface for files deposited there.
  - When we deposit to Zenodo it is a convenience for users and I don't believe we want to rebuild the entire zenodo repository interface for files sent to Zenodo from Dryad.
- `populate_container_files_from_last` finds the previous version of the file and copies the information over into this version (useful for unchanged files that continue from version to version).
- Changes to Amoeba settings and inserting copying into our duplication process.  This is all optimized so that it doesn't send 12,000 delete requests (for a zip file that has 12,000 contained files) or send 12,000 queries to copy the files.
- Population script with can be run with `rails compressed:update_contents`.
  -  I decided to cut off the population at 1,000 files per container to avoid insanity when people zip up millions of 80 byte files or something weird like that.
  - This should go through the compressed files it finds in the database that are submitted and unprocessed and add information to the container_files table.
  - Tests.

Builds on previous tar pull request (also included here).


sfisher:
Ok.  I got the problem fixed with Frictionless and amoeba.

ahamelers (approved):
This looks good, and I think it's a really good idea to move this to its own table considering how long these JSON objects can be, and that .* is being used for the parent table in many existing queries.

ryscher:
Thanks, Scott, this looks great!

PR #1149, opened 2023-03-24
Removes extra credentials no longer used

sfisher:
- unused Merritt collections (all should be going to the main merritt collection for dev/stage/prod now)
- Google maps API key
- old dryad access token


ryscher:
I merged `main` into this branch, keeping the version of the credentials that was in this branch. Then added a new commit to update the credentials for salesforce so they were in sync with main.

sfisher:
Thanks for doing this, Ryan!

sfisher:
I'll wait for the test to run and then merge.

PR #1152, opened 2023-03-28
Improve sorting for reports of tiered payment plans

ryscher:
Update the process for navigating the sponsor relationships when calculating results for tiered payment plans (#1114). This allows both hierarchical plans (Wiley) and flat plans (AAAS). The old process would not work, because it used a bottom-up approach, with journal IDs for ordering, so the flat organization could sort between some groups of the hierarchical organization. The result would be that the hierarchical organization would end up with two separate reports.

The new process uses top-down navigation, allowing each organization to complete before processing moves on to other organizations. 

Testing this is unfortunately complex. Ensure you have at least one organization with multiple sub-organizations, where the sub-organizations sponsor journals on a tiered plan. Ensure these journals have at least one dataset published in a given month, then run reports for the month, like:
```
rails identifiers:shopping_cart_report YEAR_MONTH=2023-01
rails identifiers:tiered_journal_reports SC_REPORT=shopping_cart_report_2023-01.csv
```

### DIFF
--- a/app/models/stash_engine/journal_organization.rb
+++ b/app/models/stash_engine/journal_organization.rb
@@ -6,8 +6,33 @@ module StashEngine
     # Treat the 'type' column as a string, not a single-inheritance class name
     self.inheritance_column = :_type_disabled
 
+    # journals sponsored directly by this org
     def journals_sponsored
       StashEngine::Journal.where(sponsor_id: id)
+    end
+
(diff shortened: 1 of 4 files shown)


sfisher (approved):
Thank you for doing this, @ryscher .  Good times with depth-first recursion!

Also, thanks for all the tests since it's good to see it working.

PR #1156, opened 2023-03-30
Upgrades TinyMCE 5 to 6

sfisher:
We need to do this by April 20th when 5 will no longer get updates, including security updates.

- Double check that it is loading and works correctly on our metadata entry page (abstract, methods, usage notes)
- Updated our helper that checks all kinds of crazy stuff to be sure TinyMCE really loaded before filling stuff in.  Some of their interface changed.
- Updated NPM packages with yarn.

This turned out to be less awful that I thought it might be.  Our tests are sometimes hard to get reliable with TinyMCE loaded.

### DIFF
--- a/app/javascript/react/components/MetadataEntry/Description.jsx
+++ b/app/javascript/react/components/MetadataEntry/Description.jsx
@@ -53,13 +53,11 @@ export default function Description({
           height: 300,
           width: '100%',
           menubar: false,
-          plugins: [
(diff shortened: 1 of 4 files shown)


sfisher:
@briri the reason I added you to this PR is because I know you use it, also.  I hope the diff of changes I had to make is helpful.  We're also using a react plugin for it, but I think the configuration is pretty much the same and it just wraps it.

ahamelers (approved):
This works great, thanks Scott

PR #1157, opened 2023-03-30
Improve ISSN handling for journal administrators in API

ryscher:
Journal administrators complained that they did not have the permissions to submit datasets for their journal. The root cause of this problem was the API failing to properly decode an array of ISSNs when computing permissions.

This PR adds a new method to the Journal model, `issn_array`, which forces the ISSNs to be returned in an array (rather than a string that looks like an array), so it can be more easily handled by the code that calculates permissions.

### DIFF
--- a/app/controllers/stash_api/datasets_controller.rb
+++ b/app/controllers/stash_api/datasets_controller.rb
@@ -556,7 +556,7 @@ module StashApi
 
       unless @user.role == 'superuser' ||
              params['userId'].to_i == @user.id || # or it is your own user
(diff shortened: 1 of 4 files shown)


sfisher (approved):
I think automatic conversion might only work with JSON fields in the database (but a single value without an array or object notation like a hash wouldn't be valid json).

This works well.  I think it might break if you had nested hashes using Ruby symbols as keys or other things that didn't align with JSON (unless you also converted them to JS object notation going in).

I don't ever see this field getting more complicated with nested objects so it is OK.

Thanks for fixing this.

PR #1162, opened 2023-04-04
Add new tenant

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2445


sfisher (commented):
Looks pretty good to me.  I can't remember which value we really use for name (short name or long name) or if we use both in different places.

The two weird names came about for cases like "UC Berkeley" vs "University of California, Berkeley" (and UCB for abbreviation).

IDK if the short_name should be "University of Nevada" (does that also mean the Reno campus?) or if they should both be "University of Nevada, Reno"?

(The University of California often refers to Berkeley here since it was the oldest campus but people outside California often don't know that.)

PR #1164, opened 2023-04-04
uninitialized constant for compressed error fix (compressed files)

sfisher:
- Adds error constant that didn't exist
- Make file extension case insensitive, even though traditionally most extensions are lowercase


ryscher:
Thanks, Scott! Lots of great error checking.

PR #1167, opened 2023-04-06
Adding tracker for DataCite COUNTER for views

sfisher:
This just adds some config and javascript to the page so we can begin tracking.


ahamelers (approved):
Thanks Scott!

PR #1168, opened 2023-04-11
Simplify merritt credentials

sfisher:
This moves all the duplicated individual Merritt credentials out of each tenant since all is in the same collection these days.  Moves into the APP_CONFIG instead and cleans up lots of tenant files, refactoring of code and updates to tests.

To test:
- Submisisons to Merritt still work
- Downloads still work
- Copying to Zenodo still works

BTW, this can wait until the next sprint release if we like and no reason we have to rush it in.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -60,6 +60,7 @@ gem 'aws-sdk-s3', '~> 1.113'
 gem 'blacklight'
 gem 'bootsnap', require: false
 gem 'bootstrap', '~> 4.0'
+gem 'charlock_holmes', '~> 0.7.7'
 gem 'concurrent-ruby', '~> 1.1.10'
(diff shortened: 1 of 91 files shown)


ahamelers (approved):
Thanks Scott!

ryscher:
This is a great change. 

The rake task `merritt_status:update` requires changes to be made in a couple more places:
- identifier.rb, line 512
- repository.rb, line 40

sfisher:
Thanks, Ryan.  In addition to the couple more instances you found I found 3-4 more.

I don't know how I missed these since I searched through and looked at things a few times.

i also updated some additional tests.

I'll do a deploy to the dev machine just to be sure it all still works.

PR #1170, opened 2023-04-12
Allow ISSNs to be logged

ryscher:
The Rails [filter_parameters](https://guides.rubyonrails.org/configuring.html#config-filter-parameters) setting is treated as a regular expression instead of an exact match. By default, Rails suppresses logging of `ssn`, which makes sense, but this also prevents logging of `publicationISSN`, which is important for debugging API interactions.

I'm removing the `ssn` setting, since it is important for us to log anything related to an ISSN. I don't expect that we will ever receive social security numbers.


ahamelers:
Seems perfectly sensible 😁

PR #1171, opened 2023-04-12
Methods to get zip file contents in alternate ways for items with problems

sfisher:
I added a couple of fallbacks that read the entire file and try to get the contents from the full stream.  IDK if they're much more reliable but could maybe be used if needed.

The two methods:
- fallback_file_entries1
- fallback_file_entries2

Example:
```
$ rails c -e production

df = StashEngine::DataFile.find(412283)
zip_info = Stash::Compressed::ZipInfo.new(presigned_url: df.merritt_s3_presigned_url)
zip_info.file_entries
zip_info.fallback_file_entries1
zip_info.fallback_file_entries2
```

Output should be pretty much the same for all 3 on average files except fallbacks don't give compressed_size information.  Maybe there are some edge cases where one will give output that others can't.

### DIFF
--- a/lib/stash/compressed/zip_info.rb
+++ b/lib/stash/compressed/zip_info.rb
@@ -173,7 +173,7 @@ module Stash
         # response.body.each do |chunk|
         remote_file = Down::Http.open(@presigned_url)
         Zip::InputStream.open(remote_file) do |io|
-          while entry = io.get_next_entry
+          while (entry = io.get_next_entry)
             file_info << { file_name: entry.name, uncompressed_size: entry.size }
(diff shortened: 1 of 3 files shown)


ahamelers (commented):
Should the attempt to get zip file contents be set up to automatically try each of the 2 fallbacks in case of failure? Or are they only for use in particular circumstances?

sfisher:
I think it makes sense to have them run automatically but I haven't done that yet.  I'm trying to finish up the 1st run of all the old stuff and it probably makes sense to try another round with the fallbacks in that script and see how many more it can pick up and populate into our database.

sfisher:
I added the automatic fallback for zips.  I might look into similar for tars if it will take a stream from `curl` or something.

PR #1172, opened 2023-04-12
Document process for doing a targeted database restore

ryscher:
I had a typo in a command this morning and wiped out permissions on the staging server. Documenting the restore process so it is easier next time.


sfisher (approved):
Thanks for documenting that, @ryscher .  I've had to do it a couple of times over the years and I should've documented it then.

PR #1174, opened 2023-04-14
Fixes some incorrect paths and troublehooting counter upload

sfisher:
- Some of the `require_relative` paths were now incorrect after some refactor
- Adding a bit of output for errors.

It seems that our credentials at DataCite no longer work for uploading to their hub so trying to get updated ones from Kristian.


ahamelers:
Thanks Scott!

PR #1175, opened 2023-04-17
Do not dump stash_engine_container_files

sfisher:
since it causes the IO burst credits to go down drastically since it has lots of rows.

I am testing this out (see https://metrics.librato.com/s/public/azpakcmwd?duration=1800&source=%2Ards-uc3-dryad-prd%2A ).  On the first run, the burst credits went down from 99% to 95% after running.

The dump files are about 25% smaller.  I need to watch the credits for a day or so to see if they recover afterward or just keep going down more and more over time.  If the burst credits are depleting slowly still then we may need to look into other solutions (and may need to soon as our database grows for other reasons).

### DIFF
--- a/lib/tasks/dev_ops.rake
+++ b/lib/tasks/dev_ops.rake
@@ -95,7 +95,7 @@ namespace :dev_ops do
     db = YAML.load(ERB.new(File.read(File.join(Rails.root, 'config', 'database.yml'))).result)[Rails.env]
     # rubocop:enable Security/YAMLLoad
     file = File.join(directory, "#{Rails.env}_#{Time.now.strftime('%H_%M')}.sql")
-    p command = 'mysqldump --opt --skip-add-locks --single-transaction --no-create-db ' \
+    p command = 'mysqldump --opt --skip-add-locks --single-transaction --no-create-db --ignore-table=dryad.stash_engine_container_files ' \
                 "-h #{db['host']} -u #{db['username']} -p#{db['password']} #{db['database']} | gzip > #{file}.gz"
     exec command
   end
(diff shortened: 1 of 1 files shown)


ryscher:
Thanks Scott! I'm glad you were able to diagnose this so quickly.

Other tables we should be able to exclude:
- stash_engine_ror_orgs (easy to recreate from the ROR download)
- stash_engine_curation_stats (this table is just a cache of stats that take a while to calculate, though it's probably not a big space consumer)

I'm also wondering whether we can periodically clean stash_engine_frictionless_reports. It seems to have a lot of content, which we never use after the dataset is published.

ryscher:
For reference, here are the largest tables currently in the DB:

```
+---------------------------------------+-----------+
| Table                                 | Size (MB) |
+---------------------------------------+-----------+
| stash_engine_frictionless_reports     |      3908 |
| stash_engine_container_files          |      3725 |
| dcs_descriptions                      |       764 |
| stash_engine_generic_files            |       464 |
| stash_engine_manuscripts              |       390 |
| stash_engine_external_references      |       259 |
| stash_engine_identifiers              |       191 |
| stash_engine_curation_activities      |       137 |
| stash_engine_authors                  |       100 |

```

sfisher:
Thank you for further information, Ryan.  As Ashley pointed out, we can still retrieve these excluded tables from the RDS backups on AWS if we feel we really need them.

I added these things to ignore

```
                '--ignore-table=dryad.stash_engine_ror_orgs --ignore-table=dryad.stash_engine_curation_stats ' \
                '--ignore-table=dryad.stash_engine_frictionless_reports --ignore-table=dryad.stash_engine_download_tokens ' \
```

If we ignore the frictionless reports then it should generate any for new files uploaded and they don't matter much once something has been submitted once so I don't think they're very critical to keep.  Maybe a few people who had just uploaded files don't see a report.  🤷   IDK if you feel any differently.

The stash_engine_download_tokens are items given to us by Merritt for downloads that have been assembled as zip files and are just to cache the items while until they expire.  If the table is wiped out then some things might need to be recreated again by Merritt (but they expire after a day or a few days, anyway) so we don't need to keep the contents really.

ryscher:
Thanks Scott, this is great, and appears to be working well on prod.

PR #1176, opened 2023-04-18
Track the places downloads are initiated from the UI

sfisher:
- individual files
- pre-assembled zip
- wait to assemble and then download

You can open the javascript console in the web browser and see messages that it is being logged around these events.


ahamelers (approved):
This is a great addition! Thanks Scott

PR #1179, opened 2023-04-20
Add/update curation reports

ryscher:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2419 and https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2239

- Adding a rake report `curation_publication_report` that generates stats Jess wanted to present to the board. It is mostly a recombining of stats we had in other reports.
- On the curation stats page, replace the "To be curated" column with the similar but more useful "Queue size". The new value calculates the number of datasets that actually require curation on a given day, rather than the old stat that did a very complex calculation to determine the items that were "new" in the queue each day. Since this is a shift in meaning instead of a completely new stat, I kept the existing database column.

### DIFF
--- a/app/models/stash_engine/curation_stats.rb
+++ b/app/models/stash_engine/curation_stats.rb
@@ -40,6 +40,18 @@ module StashEngine
       populate_author_versioned
     end
 
+    def status_on_date(identifier)
+      return nil if identifier.created_at > date + 1.day
+
+      curr_status = 'in_progress'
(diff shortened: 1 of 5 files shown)


ahamelers (approved):
The new rake task looks good, and wow, that Queue Size query is much simpler! Seems more useful, too. Looks great as long as the curators are expecting this Queue Size change 😄

PR #1180, opened 2023-04-21
Update documentation for changing journal payment settings

ryscher:
Improve robustness of the sample code. Even when we have retroactive updates like https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2458, we don't want to go back and change all of the transactions that were individually billed through Stripe.


ahamelers:
Thanks Ryan! Again!

PR #1181, opened 2023-04-21
When transferring ownership, transfer editorial control as well

ryscher:
When a manuscript system transfers dataset ownership to a user, it should also transfer editorial control, in case the user changes their mind and decides not to submit the dataset, but wants to delete it. (Datasets created through the UI assign the initial editorial control to the submitter, so this is enforcing the same behavior.)

For reference, see [salesforce case 00022276](https://dryad.lightning.force.com/lightning/r/Case/5003h00000bTmQFAA0/view).


ahamelers:
Thanks Ryan

PR #1182, opened 2023-04-22
Improve robustness of geographic authors report

ryscher:
When I ran this report on prod, it stopped early. This should fix the problem.


ahamelers:
Thanks Ryan!

PR #1183, opened 2023-04-24
Use Pundit for stash_engine permissions

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2498

This currently applies the Pundit gem to stash_engine, and uses it in place of previous `before_action`s where the current_user's role drove the permission settings. It's also used in a couple spots for scoping (showing only particular records from a set based on user role).

There are other variables that come into play for permissions—session information, whether the resource "is modifiable", among others—We should also use Pundit for these settings but that is somewhat more complicated. I have added comments to files where we might do this in future.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -184,3 +184,5 @@ group :development, :test, :local_dev do
   # rspec command for spring (https://github.com/jonleighton/spring-commands-rspec)
   gem 'spring-commands-rspec'
 end
+
+gem 'pundit', '~> 2.3'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -528,6 +528,8 @@ GEM
     public_suffix (5.0.1)
     puma (6.1.0)
       nio4r (~> 2.0)
+    pundit (2.3.0)
+      activesupport (>= 3.0.0)
     racc (1.6.2)
     rack (2.2.6.3)
     rack-attack (6.6.1)
@@ -860,6 +862,7 @@ DEPENDENCIES
   pry-rails
   pry-remote
   puma
+  pundit (~> 2.3)
   rack-attack
   rails (~> 6.1.0)
   rb-readline
--- a/app/controllers/stash_engine/admin_dataset_funders_controller.rb
+++ b/app/controllers/stash_engine/admin_dataset_funders_controller.rb
@@ -5,12 +5,12 @@ module StashEngine
     def index
       setup_paging
 
-      @rep = authorize Report.new, policy_class: AdminDatasetsFunderPolicy
-      # the AdminFundersController::Report is where most of the complicated SQL is for this
+      @rep = authorize Report.new, policy_class: AdminDatasetFunderPolicy
+      # the AdminDatasetFundersController::Report is where most of the complicated SQL is for this
 
       # WHERE conditions
       # Limit to tenant by either role or selected limit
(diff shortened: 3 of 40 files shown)


ahamelers:
If this is merged in I will use Pundit policy scopes around the regular user dashboard and to complete https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2462

sfisher (changes requested):
Thanks for all this work, Audrey.  It looks like a lot of changes.

I ran into a couple problems when I tested this branch on my local machine against the dev database.

I get 
`Pundit::NotDefinedError in StashEngine::ZenodoQueueController#index
unable to find policy` when I try go go to manage the Zenodo queue as a superuser.

I got a similar error for curation stats, I think maybe from the curation page (I don't exactly remember):

```
Pundit::NotDefinedError in StashEngine::CurationStatsController#index
unable to find policy `StashEngine::AdminDatasetsController::StatsPolicy` for `#<StashEngine::AdminDatasetsController::Stats:0x00007fb062c8d880 @tenant_id=nil, @since="1000-01-01 00:00:00", @untouched_since=2023-04-27 10:42:43.235756 -0700>
```

I also tried to see what happened when I went to a page that wasn't authorized.  I change my role from `superuser` to `user` and tried to go to some curation pages.  It seemed to do a lot of redirects but then never really rendered any page or told me that I was unauthorized.

IDK, maybe there was something odd in my testing and I was pointing at the development database.

ahamelers:
Thank you Scott, great catches—the zenodo bug in particular only showed up if the ZenodoCopy results are Nil. When fixing these 2 things I also did a check to see if the same problems would occur from any of the other authorizations, luckily it didn't seem so.

ryscher:
Overall, looking good. There are a lot of cases to test here, so I hope we can catch them all.

Here are the issues I found:
  - As a superuser, on the Dataset funders page, any time I hit the Submit button, there are 0 results. (Possibly because a blank `tenant` is added to the params?)
  - As a curator, I have permissions to visit `/stash/status_dashboard`, even though it is not on the menu.
  - As a basic user, if I attempt to use a URL for the Datasets admin page with some parameters set, I get "too many redirects" instead of the expected permissions error. (e.g., `/stash/ds_admin?q=merritt&sort=relevance&commit=Search&curation_status=submitted&editor_id=23`)
  - As a basic user, I can download unpublished datasets through the API

ahamelers:
> As a basic user, I can use the API to view and download unpublished datasets

@ryscher I've only added pundit policies to stash_engine, not stash_api, and I cannot recreate this bug in my own environment

Screenshot 2023-05-05 at 09 47 47

ahamelers:
> As a basic user, if I attempt to use a URL for the Datasets admin page with some parameters set, I get "too many redirects" instead of the expected permissions error. (e.g., `/stash/ds_admin?q=merritt&sort=relevance&commit=Search&curation_status=submitted&editor_id=23`)

@ryscher I cannot recreate this either, on my local or using the local_dev environment, I get the not authorized message as expected:

Screenshot 2023-05-05 at 11 42 17

ahamelers:
The other two issues are fixed, but by the way, it looks like on the _current production site_, all visitors, including those not even logged in!!, can visit the status dashboard page 😱

ryscher:
Ok, I've reviewed these again, and I was unable to reproduce the two extra problems. It's quite likely that I made a mistake when testing. Merging this now...

PR #1185, opened 2023-04-26
Add social sharing buttons to dataset landing page

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2484

### DIFF
--- a/app/assets/stylesheets/stash_engine/ui.css
+++ b/app/assets/stylesheets/stash_engine/ui.css
@@ -7865,6 +7865,52 @@ img.c-upload__spinner {
   margin-bottom: 3em;
 }
 
+#share_buttons {
+  display: -webkit-box;
+  display: -ms-flexbox;
+  display: flex;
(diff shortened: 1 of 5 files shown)


sfisher (approved):
Works great.  Thanks for adding this, Audrey.

Though I find the prominence of this with its own heading and section a bit odd, but I think it's OK until/unless there is a redesign.

ahamelers:
> Though I find the prominence of this with its own heading and section a bit odd, but I think it's OK until/unless there is a redesign.

Yes, I think in a redesign I think it would be smaller, but possibly still on the right side. If I were redesigning the page, I'd really put a lot of the stuff on the right side into the larger main content on the left. In studies users tend not to look at the smaller "menu" side of a given page until they've already checked the main content for what they're looking for, so it's very odd to me that the _main content_ of the page, i.e. the dataset content itself, is minimized and over in the often ignored area.

ryscher:
Looks great, thanks!!!

PR #1188, opened 2023-04-26
Track tries analyzing compressed content and skip after 3

sfisher:
- Adds column `compressed_try` which is an integer
- Limit query to items with less than 3 tries
- Increment when we process an item

I think this may be good enough since the errors are logged and we can look up by the file_id


ahamelers (approved):
Perfect! Thanks Scott

PR #1190, opened 2023-04-27
Save and display resource total file upload size

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2126
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1695

### DIFF
--- a/app/assets/stylesheets/stash_engine/ui.css
+++ b/app/assets/stylesheets/stash_engine/ui.css
@@ -7432,10 +7432,28 @@ img.c-upload__spinner {
 }
 
 /* ********************************************** */
+.c-uploadtable-header {
+  display: -webkit-box;
+  display: -ms-flexbox;
+  display: flex;
+  -webkit-box-align: baseline;
+      -ms-flex-align: baseline;
+          align-items: baseline;
+  -webkit-box-pack: justify;
+      -ms-flex-pack: justify;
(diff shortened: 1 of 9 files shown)


ahamelers:
This check is in /stash_api/files_controller.rb and is also using the Merritt total storage size:

```ruby
def check_total_size_violations
      return if @resource.new_size <= APP_CONFIG.maximums.merritt_size && @resource.size <= APP_CONFIG.maximums.merritt_size

      (render json: { error:
                          'The files for this dataset are larger than the allowed version or total object size' }.to_json,
              status: 403) && yield
 end
```

Should it also be changed to use the total size of the current resource? Should anything else?

sfisher:
Looks pretty good to me.  I suppose total stored size is what we pay storage costs for and reflects certain information more accurately.

On the other hand if someone is just downloading a version then the size of that version may be a better reflection of what they want to do.  Also the zipped size of "full download" will also differ because of compression (which I think is why we stopped giving figures on the download page for that since we don't have an accurate size until the zip is created).

Even for charging people it seems like we've gone more toward "the uncompressed size of a latest version" and just figuring we might occasionally be eating some costs which might not be reflected in that size figure.

ryscher:
Great, thanks!

PR #1191, opened 2023-05-03
Basic file validation for JSON and XML files

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2404, https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2048

I have added basic validation for structured data files (JSON, XML) to the lambda—not as part of frictionless validation, but through python itself. This lambda code is currently deployed on AWS for testing—I took care that the code will keep working for existing file submissions to the lambda without disruption.

I've also added frictionless validation for TSV files. In my tests these get the same validation results as CSV versions of the same files. TSV and CSV results differ slightly from XLS/XLSX versions of the same files, but are comparable.


ryscher:
Looks great, works with the examples I tried. Thanks for adding this!

PR #1198, opened 2023-05-12
fixing permissions on zenodo queue (I hope)

sfisher:
I haven't looked through the permissions changes in depth, but some of these actions from the zenodo queue management page were no longer working.

I removed the "authorize" on specific queries and put in the `authorize %i[stash_engine zenodo_copy]` like in the index action.  I think this is ok, right?  The only people who need to access this controller/actions are the superusers.

@ahamelers IDK if this is the right thing to do since I haven't looked in depth, but I figured if this worked for the index it would be OK for the other actions, also.

### DIFF
--- a/app/controllers/stash_engine/zenodo_queue_controller.rb
+++ b/app/controllers/stash_engine/zenodo_queue_controller.rb
@@ -36,12 +36,14 @@ module StashEngine
     end
 
     def item_details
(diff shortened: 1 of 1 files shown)


ahamelers:
@sfisher Yes, that's authorizing against the policy without doing an actual database search and using the model, which seems to be best for this model since there are issues around Nilclass. This is a good solution.

PR #1201, opened 2023-05-15
Removes old datacite tracker for plausible.io

sfisher:
I believe the previous tracker was put in for plausible.io and it was still present in addition to the new one we put in and that was what Kristian was seeing.

I see the console log showing tracking is happening so I think it's working.  Too bad I missed the old one still there.


ryscher:
Thanks!

PR #1203, opened 2023-05-17
User dashboard redesign

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2462

Would also need to:

- ~Check and update any documentation referencing the user dashboard~
- Create a clean-up pull request, eventually, to delete unused templates and methods (there are a bunch floating around, not just around this user dashboard change)


sfisher (commented):
I liked this more than our current weird thing with datasets showing up in both sections.  I wasn't sure about the section grouping descriptions.  I was wondering if "Needs attention" would be better as "To complete" or "Incomplete."

I also wondered if there was a more descriptive heading for "Kept private" if these are only the PPR things.  I didn't try to see where "embargoed" show up since they have metadata released to the public but no files.

Overall it's probably better than what we have and which I always thought was confusing with multiple copies of the dataset showing in different places.  (It was something we received form someone else and I think it had even been tested and I was surprised other people didn't find it a bit confusing.  🤷 )

Screenshot 2023-05-22 at 5 06 47 PM

IDK if this is something that we wanted to move into a release right away or wait for more consideration and feedback?

sfisher:
Oh, @ahamelers , I think I found one problem with the changes.  Any link or button that creates a new version of a dataset needs to be locked after it's clicked once since creating a new version can be a bit slow.

Otherwise, CrazyClicking like a machine gun creates multiple new versions (which I think don't show up in this page well) but they are something we want to prevent.  It takes long enough for all the create operations for duplicating the dataset that people are able to do this.

PR #1205, opened 2023-05-22
The university of Rochester entity id changed for shibboleth

sfisher:
It looks like their shibboleth entity id changed, so updating it.


ryscher:
Thanks, Scott, this works for me.

PR #1208, opened 2023-05-24
Email links additional features to copy/paste

sfisher:
See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2540 about complaints.

I was unable to duplicate this as a real problem for most people.

- If you have an email program correctly associated on your computer then clicking the links correctly opened it and filled the address correctly (both Chrome/Firefox).
- Tried in normal window, incognito in both browsers.
- Never saw it fail.

Part of the obfuscation is that if a dumb bot reads the page source without rendering then it either gets `dev@null` or it gets the email address backward so they don't spam us to death with crap.

I added a copy button next to the obfuscated email addresses, so if someone needs to deal with getting the email addresses manually by copying and pasting (rather than just retyping it) then they get it correctly from that button.

IDK if this is the best solution, but we can let Audrey decide if we're not sure.


ryscher:
(I thought I had merged this already.... I was working in an area with poor connection, so maybe it didn't go through???)

PR #1209, opened 2023-05-25
Adds to cron to update compressed contents daily

sfisher:
Checks zip/tar.gz daily to add index of compressed contents.  Skips items tried at least 3 times.


ryscher:
Thanks, Scott. It's great to get this automated.

PR #1210, opened 2023-05-25
Further improve robustness for the social sharing links

ryscher:


sfisher (approved):
Just FYI, @ahamelers  I always try to avoid using the + sign for concatenation since it fails if a value is nil.  I try to use string interpolation instead (like `"some stuff: #{substitution}"`) since any nil values are automatically converted to an empty string.  (As long as I don't really care if the nil is replaced with nothing in the string then at least it doesn't error.)  Though in this case it was probably a nil in the method chain at least in that one place.

PR #1215, opened 2023-06-01
Make three keywords (subjects) required

sfisher:
- Adds a validation for at least 3 normal subjects (keywords)
- Add required css class
- Adds capybara action to fill keywords
- Finds places that required fields were failing in capybara tests and fix them
- Find additional failures because of keywords and fix tests

(Running all the slow tests and fixing the tests is what took a looong time with these changes.)

In order to test you can either change the code `subjects_require_date = '2023-06-07'` to a date in the past or else change the `@resource.identifier.created_at` to a date in the future past this cutoff.

I had the date set in the past to run all the tests and find and fix all the failures.

The ticket for this mentions adding to the search landing page, but they're already there. They are also already available as a search facet.

Screenshot 2023-06-01 at 4 33 54 PM

Screenshot 2023-06-01 at 4 34 19 PM

Possibly confusing since we call them `keywords` in the entry and landing page and the search calls them `Subjects`.  They are stored as DataCite subjects no mater what they're called.  We can revise terminology to be consistent if we wish.

### DIFF
--- a/app/javascript/react/components/MetadataEntry/Keywords.jsx
+++ b/app/javascript/react/components/MetadataEntry/Keywords.jsx
@@ -50,7 +50,7 @@ function Keywords({
 
   return (
     <div className="c-keywords">
-      <label className="c-input__label" id="label_keyword_ac" htmlFor="keyword_ac">Keywords:</label>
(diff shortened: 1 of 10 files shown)


ryscher:
Thanks, Scott. Sorry the tests took so much work to make compatible.

PR #1216, opened 2023-06-02
When old old bots try cached paths from v1 Dryad, don't error

ryscher:


sfisher (approved):
Good so they don't get a 404.

PR #1217, opened 2023-06-02
Make rake task for importing new versions of PLoS subjects

sfisher:
The PLoS keywords/subjects we had before seemed to be mangled and the old import was out of date and hard to use.

- Get PLoS keywords
- Save the Excel file to tab separated values
- Run the script with the environment variables noted like `rails keywords:update_plos RAILS_ENV=local_dev PLOS_PATH="/Users/sfisher/Downloads/plosthes.2020-1.full.tsv"`
- Wait for it to finish

It will add subjects that don't exist and also accurately fill a better subject_scheme and scheme_URI for keywords that already exist but that don't have them filled for a value that matches a PLoS subject.  It may also correct capitalization to match the way PLoS has done it officially.

This may mean we can only autocomplete on controlled vocabularies so people don't select dumb stuff in their autocomplete in the future.

### DIFF
--- a/lib/tasks/keywords.rake
+++ b/lib/tasks/keywords.rake
@@ -1,0 +1,31 @@
+namespace :keywords do
+  task update_plos: :environment do
+    $stdout.sync = true # keeps stdout from buffering which causes weird delays such as with tail -f
+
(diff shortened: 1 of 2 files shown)


ryscher:
Ahh, this is great! So much nicer to have a large vocab of mostly-uniform keywords.

PR #1220, opened 2023-06-05
Add a test-tenant for institutional testing

ryscher:


sfisher (commented):
I'm not sure if `partner_display` is really used, but if it is then it might be good to set that to `false` for production, also.

ryscher:
Good catch! With it set to true, the tenant would still be listed on the membership page. Fixed now.

sfisher (approved):
Looks good.  Thanks.

PR #1225, opened 2023-06-07
try to save capybara screenshots as artifacts

sfisher:
So that maybe we can see what those failures look like and why they're failing on github actions.

---

I think this probably works.  It's ridiculous since it was failing repeatedly and not creating artifacts from the run.  Then I found out I needed to set it so that it runs that step only on a failure condition and now it has succeeded repeatedly.  🤷‍♂️ 

This also adds the "fail fast" option so that after one failure it stops.  I figure this may be nicer for this long test so if it fails you can troubleshoot or else click the "rerun" button faster.


ryscher:
Thanks, this will be incredibly useful.

PR #1226, opened 2023-06-08
Adding additional metadata fields to DataCite XML & rework of dates

sfisher:
https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2403 & https://github.com/CDL-Dryad/dryad-product-roadmap/issues/1185

- DataCite dates for issued (first time metadata publicly viewable) and available (first time files viewable) -- pulls off the identifier and resource view flags from model.
- Add Subject schemes and schemeURIs.
- Author affiliation RORs & Funder ids (currently crossref funder ids) gives scheme & id
- Additional testing galore for the trip from our database into XML output for DataCite

Also added some changes to try to make the feature tests a little more reliable because of race conditions.

Usually things are fine if you set up everything from the user interface, but if you modify the database contents independently of the UI then you get race conditions since database setup is modified directly from the test and things taking place from the browser and puma are in a different thread.

🏁 🚗  (database changes in test)
🏁 🚚  (puma and browser automation)


ryscher:
I created an published a test submission on my machine. The DataCite metadata is viewable at: https://api.test.datacite.org/dois/application/vnd.datacite.datacite+json/10.7959/dryad.x95x69pcr

Issues I still see:
- The date "Issued" is 2023 (should be 2023-06-14)
- date "Available" is missing
- The ROR "University of Zulia" does not have a schemeUri
- The funder "Bill and Melinda Gates Foundation" does not have a schemeUri
- The "FOS" prefix is still present in the subject "FOS: Environmental biotechnology"
- I got a duplicate "Environmental biotechnology" from the PLOS keywords (though I think this wouldn't happen on production with the DB changes that Scott made earlier)

Sorry to be so picky about seemingly meaningless stuff, but it's what the GREI team agreed on for maximum cross-repository connections.

sfisher:
> Issues I still see:
> 
> * The date "Issued" is 2023 (should be 2023-06-14)
> * date "Available" is missing

The dates being unavailable was because of the use of side effects and callback filters to submit to DataCite when the CurationActivity changes to certain states. It makes it very confusing what is happening and what order anything happens.  In order for dates to be set correctly for DataCite then the date and view/download flags need to be set correctly in the database first before generating DataCite submission.

> * The ROR "University of Zulia" does not have a schemeUri
> * The funder "Bill and Melinda Gates Foundation" does not have a schemeUri

The schemeURIs are set now.

> * The "FOS" prefix is still present in the subject "FOS: Environmental biotechnology" (Where is this coming from? I see it removed in the code, and I verified that it's using the correct version on my machine.)
> * I got a duplicate "Environmental biotechnology" from the PLOS keywords (though I think this wouldn't happen on production with the DB changes that Scott made earlier)

See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2586 about this.  This was also confusing since I couldn't find anywhere in our code or the DataCite mapping library that added this.  Turns out that the extra heading is being added by DataCite and not by us and only in the JSON version (not the XML version).

When I get more feedback about how to handle this I'll update that piece of generation code according to their guidelines (leaving out FOS or not and other things they suggest to work with their code and not to create the extras on their end).

ryscher:
Looks great.... There are still some problems in https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2586 and https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2587, but those depend on DataCite, so this is as good as we can get for now.

PR #1228, opened 2023-06-09
Don't update Field of Science subjects

sfisher:
Our subjects are a mess and we want to assign subjects to match the PLoS vocabulary when possible.

But don't update FOS type of subjects which are used and treated differently.


ryscher:
Thanks for cleaning up this duplication.

PR #1230, opened 2023-06-12
2437 quarterly update gem dependencies

sfisher:
Both bundler and yarn upgraded libraries without much interference, though there were a few errors and deprecations I needed to fix.

I'm not sure how I tell if this fixes some of the security things that github/dependabot flags.  It seems to only run against the main branch, so I guess this needs to be merged before I can tell if it fixed most of the notifications.

Also cherry-picked some of my test fixes in until those are merged into main and keep running into extreme github actions testing problems for the feature tests so attempting to find ways to make them more reliable.  It will be good to get #1225 merged in if it will save screenshots of what is constantly going wrong on github actions.


ryscher:
I had to do a little more work on the tests, but everything is passing now. Thanks for getting the bulk of it done.

PR #1231, opened 2023-06-13
Update JWT for COUNTER reporting

sfisher:
This updates things so we can submit reports again.

Also cherry-picking lots of changes in to make github action tests more reliable (I hope) since those haven't made it into main branch yet and I want this PR to have passing tests.


ryscher:
Glad this is working now!

PR #1233, opened 2023-06-15
Fix bug in initial selection of no funding

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2562


ryscher:
Thanks!

PR #1235, opened 2023-06-15
Fix 'discover' links to landing pages and HTML search results

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2037

### DIFF
--- a/app/controllers/catalog_controller.rb
+++ b/app/controllers/catalog_controller.rb
@@ -267,11 +267,17 @@ class CatalogController < ApplicationController
   # based on the InternalDatum types defined below
   # GET stash/discover?query=[:internal_datum_value]
   def discover
-    internal_datum_types = %w[pubmedID publicationDOI manuscriptNumber]
+    internal_datum_types = %w[pubmedID manuscriptNumber]
     where_clause = 'stash_engine_internal_data.data_type IN (?) AND stash_engine_internal_data.value = ?'
-    identifiers = StashEngine::Identifier
+    internal_data = StashEngine::Identifier
       .publicly_viewable.joins(:internal_data)
       .where(where_clause, internal_datum_types, params[:query])
+    related_ids = StashEngine::Identifier
+      .publicly_viewable
+      .joins(resources: :related_identifiers)
+      .where("work_type = 6 AND related_identifier like '%#{params[:query]}'")
+
+    identifiers = internal_data + related_ids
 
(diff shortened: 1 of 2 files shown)


app/controllers/catalog_controller.rb
```diff
@@ -267,11 +267,17 @@ def has_search_parameters?
...
+    internal_datum_types = %w[pubmedID manuscriptNumber]
     where_clause = 'stash_engine_internal_data.data_type IN (?) AND stash_engine_internal_data.value = ?'
-    identifiers = StashEngine::Identifier
+    internal_data = StashEngine::Identifier
       .publicly_viewable.joins(:internal_data)
       .where(where_clause, internal_datum_types, params[:query])
+    related_ids = StashEngine::Identifier
+      .publicly_viewable
+      .joins(resources: :related_identifiers)
+      .where("work_type = 6 AND related_identifier like '%#{params[:query]}'")
```
ryscher, line 278: I think there is some Rails way to use a the enum object to get the 6 here, rather than hardcoding the value.

ryscher: This should work:
.where(work_type: :primary_article).where("related_identifier like '%#{params[:query]}'")

ahamelers: Initially I had tried this, and based on your feedback I did try it again:

```ruby
related_ids = StashEngine::Identifier
      .publicly_viewable.joins(resources: :related_identifiers)
      .where(related_identifiers: {work_type: :primary_article})
      .where("related_identifier like '%#{params[:query]}'")
```

But it does not work! According to the logs, it creates a query like this:

```sql
SELECT `stash_engine_identifiers`.* FROM `stash_engine_identifiers` INNER JOIN `stash_engine_resources` ON `stash_engine_resources`.`identifier_id` = `stash_engine_identifiers`.`id` INNER JOIN `dcs_related_identifiers` `related_identifiers` ON `related_identifiers`.`resource_id` = `stash_engine_resources`.`id` WHERE `stash_engine_identifiers`.`pub_state` IN ('published', 'embargoed') AND `related_identifiers`.`work_type` = 'primary_article' AND (related_identifier like '%10.1186/s13068-020-1649-7')
```

Which in my test does not bring back results, while explicitly querying for 6 does. I think it's because of the nested join, complicating matters—unlike the internal_data, which is tied to the identifier, the dcs_related_identifier is tied to the resource. I'll try some other methods.

PR #1237, opened 2023-06-16
Handle old URLs that include "/repo"

ryscher:
I discovered that there are some old articles (including at least one by me) that reference old Dryad URLs with "/repo/blah". This prevents those URLs from producing a 404.


ahamelers:
Thanks Ryan!

PR #1239, opened 2023-06-20
Add new tenant and fix tenant logo appearance

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2588

In the course of adding a new SVG logo, I found that it and a bunch of other SVG tenant logos weren't appearing  because they were being read as having 0 height and width. I've fixed all these files.


ryscher:
Thanks!

PR #1241, opened 2023-06-22
Reorderable funders, with new funding description field

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2479, https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2475

### DIFF
--- a/app/controllers/stash_api/datasets_controller.rb
+++ b/app/controllers/stash_api/datasets_controller.rb
@@ -127,8 +127,8 @@ module StashApi
       # Add funding info if it was not added by the submitter
       if @resource.contributors.blank? && art_params['funding_information'].present?
         art_params['funding_information'].each do |f|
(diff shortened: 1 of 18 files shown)


sfisher (approved):
Thanks for adding the extra field and adding the reordering.  This works really well for me when I test it out and also the refactoring is good so it can be reused.

ryscher:
Functionally, looks great. One minor edit: Instead of "Program/description", I would use "Program/division", since those are the most likely types of information that people will want to put in the box.

ahamelers:
I've updated the label in the [associated branch](https://github.com/CDL-Dryad/dryad-app/tree/funder-entry), but it's not appearing in this pull request for some reason 🤔

ahamelers:
I finally got it to appear by resetting the commit, committing it a second time, and force pushing. Pretty weird!

PR #1243, opened 2023-06-22
Fix deployment so it works across both our dev and stage servers

sfisher:
- Lock mini_racer to 0.6.3 since amazon linux isn't handling any other version (dev server problem)
- Reverting puma to 6.1.0 instead of 6.3.0 since it wouldn't start properly through systemd (dev server problem)
- Forcing all the babel components to same version number and moving one from dev dependency to regular dependency since otherwise it didn't seem to work.
- Also removing npm lockfile since we're using yarn and yarn.lock which I regenerated from scratch.

Seems to deploy on stage now.  All these things were preventing deployment in one way or another.


ryscher:
Working well for me. Thanks for tracking this down!

PR #1244, opened 2023-06-23
Break out primary articles from other articles

sfisher:
All articles on the landing page were shown together before.  This breaks out an extra heading for a primary article.

Screenshot 2023-06-22 at 5 01 24 PM


ahamelers:
So under 'Article', there can be a list? Should that heading be plural? Maybe "Primary article" and "Related articles"?

sfisher:
It pluralizes if there is more than one.  You'll see the  `.pluralize(<number>)` in there for the article headings.

ahamelers (approved):

PR #1245, opened 2023-06-23
Save typed keywords when the Enter key is pressed

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2468


sfisher (approved):
This will be a lot better for people not choosing a keyword from the drop-down list.

ryscher:
Thanks, this does make the behavior better!

PR #1250, opened 2023-06-23
API reports framework and datasets_without_primary_articles_report

ryscher:
Adds a new report `datasets_without_primary_articles_report`, which will be used to identify datasets that are likely to be improved by the addition of a relationship to a primary article.

In addition to this individual report, this starts development of a more unified reporting framework. Ideally, most Dryad reports should be available both via command line and via API. The `datasets_without_primary_articles_report` report takes several minutes to calculate on the production database, and the contents are not likely to change frequently. This makes it a poor candidate for calculating on-the-fly in response to an API request. Instead, the report is stored in a file (updated daily), and a new API endpoint is introduced to obtain any pre-calculated reports that are available.

To test:
- Run the report: `bundle exec rails identifiers:datasets_without_primary_articles_report` 
- Verify that it has the expected content and is stored in the `reports` directory
- Verify that the report is available through the API at `/api/v2/reports/datasets_without_primary_articles`


ahamelers (approved):

ahamelers:
All works for me!

PR #1251, opened 2023-06-23
Add more action buttons to user dashboard for datasets in PPR

sfisher:
See #2594 .

IDK if this is the very perfect fix, but it was formatting the larger buttons squished up side by side so, just for this section, styled it as`flex-direction:column`.  I tried some other stuff first, but this seemed the easiest way to put this extra edit button back in below.

Also called it "Edit new version" since it sounds like they are oriented toward editing their datasets and the new version is a side effect of editing.  We can make it "Create new version" to be the same if you want, though (it's the same action).

### DIFF
--- a/app/assets/stylesheets/stash_engine/ui.css
+++ b/app/assets/stylesheets/stash_engine/ui.css
@@ -8290,6 +8290,8 @@ dl.timeline div:nth-child(5n+5) dt, dl.timeline div:nth-child(5n+5) dd::before {
   -webkit-box-pack: justify;
       -ms-flex-pack: justify;
           justify-content: space-between;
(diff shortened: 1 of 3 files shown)


ahamelers:
> Also called it "Edit new version" since it sounds like they are oriented toward editing their datasets and the new version is a side effect of editing. We can make it "Create new version" to be the same if you want, though (it's the same action).

I think making it the same button makes sense; I think we should try to let the users know what is actually happening, and to direct them first to the actions we want them to take and then to any other actions we allow. Also, it's more 'DRY' that way.

ryscher:
Looks good, everything tests fine for me. Thanks for addressing this so quickly!

PR #1252, opened 2023-06-24
Removes duplicate subjects and leaves the Field of Science alone

sfisher:
- method to remove duplicate subjects (unassociate them)
- tests for removal
  - Ignore FOS since it's treated differently and not like a normal subject
  - Prefer to remove a subject without a controlled vocab than one with one

The ticket for this #2467 was vague on details, when or how this happens.  I put a call into the editing screen load (find or create) since I think that should catch duplicates if the import does them as was speculated.

It does something different now and maybe the problem is fixed, not really sure since there were no steps telling how to replicate the problem. 🤷‍♂️

### DIFF
--- a/app/controllers/stash_engine/metadata_entry_pages_controller.rb
+++ b/app/controllers/stash_engine/metadata_entry_pages_controller.rb
@@ -23,6 +23,8 @@ module StashEngine
 
     # GET/POST/PUT  /generals/find_or_create
     def find_or_create
+      @resource&.purge_duplicate_subjects!
+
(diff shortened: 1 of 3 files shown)


ahamelers (approved):
These changes look good to me, but I wonder if @ryscher can take a look as well, since from the comments on the original ticket he may have more of an idea of what it was about (it was one of those tickets created by copying and pasting info from the big spreadsheet of curator issues)

ryscher:
This looks great. I was unable to create a duplicate subject. I think it will solve any problem that could still have been in the system.

PR #1255, opened 2023-06-28
Document some process for cleaning institution names with asterisks

ryscher:
This is a first pass at a process for the regular cleaning of affiliations that are not tied to a ROR entry. We will probably improve this process every time we perform it.


sfisher (approved):
Thanks for documenting this, Ryan.

PR #1257, opened 2023-06-29
Fix functions of enter key on keyword entry

ahamelers:
I also fixed a console error I was seeing about the keyword PropTypes


sfisher (approved):
Thanks for the fix.  I tested and it works well.

PR #1258, opened 2023-06-29
Fix checks for required fields for primary article

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2585

Also fixed the import button to sentence casing

### DIFF
--- a/app/models/stash_datacite/resource/dataset_validations.rb
+++ b/app/models/stash_datacite/resource/dataset_validations.rb
@@ -99,8 +99,10 @@ module StashDatacite
 
         journal = " from #{@resource.identifier.publication_name}" if @resource.identifier.publication_name.present?
 
-        if @resource.identifier.import_info == 'published' &&
-          (@resource.identifier.publication_article_doi.blank? || @resource.related_identifiers.where(work_type: 'primary_article').count.positive?)
+        primary_article = @resource.related_identifiers.where(work_type: 'primary_article').first
+        if @resource.identifier.import_info == 'published' && (
+                      primary_article.nil? || primary_article.related_identifier.blank? || !primary_article.valid_doi_format?
+                    )
           err << ErrorItem.new(message: "Fill in {a correctly formatted DOI} for the article#{journal}",
                                page: metadata_page(@resource),
                                ids: %w[primary_article_doi])
--- a/app/models/stash_datacite/resource/dataset_validations.rb
+++ b/app/models/stash_datacite/resource/dataset_validations.rb
@@ -88,6 +88,30 @@ module StashDatacite
         err.flatten
       end
 
+      def article_id
+        err = []
+
+        if @resource.identifier.import_info != 'other' && @resource.identifier.publication_name.blank?
+          err << ErrorItem.new(message: 'Fill in the {journal of the related publication}',
+                               page: metadata_page(@resource),
+                               ids: ['publication'])
+        end
+
+        journal = " from #{@resource.identifier.publication_name}" if @resource.identifier.publication_name.present?
+
+        if @resource.identifier.import_info == 'published' &&
+          (@resource.identifier.publication_article_doi.blank? || @resource.related_identifiers.where(work_type: 'primary_article').count.positive?)
+          err << ErrorItem.new(message: "Fill in {a correctly formatted DOI} for the article#{journal}",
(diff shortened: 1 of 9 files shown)


sfisher (commented):
This works pretty well for me, though I notice some differences of what is acceptable for the DOI formatting from the past and now it only accepts one style in order to submit.  It now validates and requires a bare-DOI identifier style rather than also accepting the URL style (which I think is the DataCite official style).

It actually functions, no matter which style of DOI is entered and will autofill the page and also shows as an Article related work at the bottom of the page (with the full URL).

However the validation will not accept either style, even though it otherwise seems to function.  IDK if that's a big deal.

For example, try it with importing "Nature Energy" and "https://doi.org/10.1038/s41560-023-01282-z" vs "10.1038/s41560-023-01282-z" in different datasets.  Both will autofill and fill the related work, but only "10.1038/s41560-023-01282-z" is accepted as valid in the review page.  🤷‍♂️ 

Probably a dumb, minor thing but inconsistent with what we'll accept vs what the review page says.  (And if we insist on one or the other only, I think the full url is the preferred one by DataCite.)

ryscher:
For ease of copy/paste, we should accept all 3: bare style, URL style, and namespace style ("doi:")

ahamelers:
Here's all the information I can discover about why this check was added (almost exactly a year ago!):

The commit: https://github.com/CDL-Dryad/dryad-app/commit/5f9b5ac3b74bbb08a71a720adc428513f1e776d4

Pull request: https://github.com/CDL-Dryad/dryad-app/pull/884

Release notes: https://github.com/CDL-Dryad/dryad-app/releases/tag/v0.7.32a

app/models/stash_datacite/resource/dataset_validations.rb
```diff
@@ -88,6 +88,30 @@ def loose_errors
...
+
+        if @resource.identifier.import_info != 'other' && @resource.identifier.publication_name.blank?
+          err << ErrorItem.new(message: 'Fill in the {journal of the related publication}',
+                               page: metadata_page(@resource),
+                               ids: ['publication'])
+        end
+
+        journal = " from #{@resource.identifier.publication_name}" if @resource.identifier.publication_name.present?
+
+        if @resource.identifier.import_info == 'published' &&
```
sfisher, line 102: Hi Audrey, this is the part that wasn't working and it was changed in this pull request.  Before it had some conditions and returned at the first of the `article_id` method.  If it got past returning early then it simply checked that a primary article existed.

The logic in here lines 102-103 seem to be the following:

IF import_info == 'published' AND (the identifier has a publication DOI OR there is a primary article) THEN give an error.

I'm making another commit to fix this and another bug I found if someone puts in a bad identifier for the primary article first.

My new code is

```
        primary_article = @resource.related_identifiers.where(work_type: 'primary_article').first
        if @resource.identifier.import_info == 'published' && (
                      primary_article.nil? || primary_article.related_identifier.blank? || !primary_article.valid_doi_format?
                    )
```

It also just looks up the primary article from this resource.  The identifier version looks through the whole history for a primary article, but we probably don't need to do that and if their primary article is gone then they probably should change the options from those 3 choices.

sfisher:
BTW, I added more commits that should fix this problem. 

PS.  The check that was there only checked that a DOI was filled if they had selected "article" and a publisher was filled in as I recall.

ahamelers:
>  The check that was there a year ago only checked that a DOI was filled IF they had selected "article" AND a publisher was filled in as I recall. I don't think it required anything to be filled if nothing was filled in the form (even if publisher was selected from radio buttons).

Yes, that is what the pull request is fixing, it is making filling these fields required.

app/models/stash_datacite/resource/dataset_validations.rb
ahamelers: Okay, I think I can see in the old code, that previously the primary_article DOI check was only triggered if the publication_doi field was blank. I guess when I was recongfiguring these checks I thought that was a bug, because it meant the primary_article DOI check was almost never triggered. If it never being triggered is preferred, should I not just remove that check and check for the blank publication_doi field alone?

ryscher:
Looks good to me. This is good to roll out now, and we can improve the overall interaction with a future redesign.

PR #1259, opened 2023-06-29
Enable admin dataset institution filter for multi-institution consortia

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2556

### DIFF
--- a/app/helpers/stash_engine/application_helper.rb
+++ b/app/helpers/stash_engine/application_helper.rb
@@ -4,11 +4,6 @@ module StashEngine
   module ApplicationHelper
     # displays log in/out based on session state, temporary for now
     # :nocov:
-
-    def institution_select
-      StashEngine::Tenant.all.map { |item| [item.short_name, item.tenant_id] }
-    end
-
     def log_in_out
       if session[:user_id].blank?
         link_to 'log in', stash_url_helpers.tenants_path
@@ -17,6 +12,10 @@ module StashEngine
       end
     end
     # :nocov:
+
+    def institution_select
+      policy_scope(StashEngine::Tenant).map { |item| [item.short_name, item.tenant_id] }
+    end
 
     # no decimal removes the after decimal bits
     def filesize(bytes, decimal_points = 2)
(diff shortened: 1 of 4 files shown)


sfisher:
This seems to work OK with the tenant naming convention for SUNY ("suny") but it turns out not to work well for the way we set up The Claremont Colleges since their administrative tenant is Claremont College Services ("clare-cs").

It might be possible to get the relationships from the RORs since the parent might list all the RORs for the other campuses, but in a roundabout and maybe hard to figure out way.

I can also imagine that people from various campuses might be randomly given curator permission for the consortium, even if they're not at a "boss" institution.  They could probably be assigned to a synthetic "boss" tenant in an artificial way if they don't need submitting privileges also which I suppose is kind of what we've done up until now as a hack since we really haven't designed anything very special for a consortium.

This all makes me think that in the long run, the most explicit and unconfusing way to deal with consortium special admins is to make another role like "consortium_curator" and have a more explicit definition of what other tenants are encompassed as part of the consortium.  Then if they're one of these tenants they get permissions for all the consortium as a "consortium_curator."

For now what you have works fine for SUNY and could probably be made to work for Claremont Colleges by some renaming of the clare-cs tenant and updating all the references for clare-cs to just "clare" in the database or other places.

ahamelers:
> This all makes me think that in the long run, the most explicit and unconfusing way to deal with consortium special admins is to make another role like "consortium_curator" and have a more explicit definition of what other tenants are encompassed as part of the consortium. Then if they're one of these tenants they get permissions for all the consortium as a "consortium_curator."

I agree that we need a more long term plan for handling consortia and their members and permissions, and @ryscher has mentioned this need as well, but I think that needs actual planning and is also out of scope for this ticket!

In the meantime, I will change this to use ROR IDs, which is not as hard as I imagined and works for both consortia.

sfisher:
Looks good to me!

ryscher:
Thanks, I love the new solution! It's elegant and works well.

PR #1261, opened 2023-06-30
Update to configure our development environment to use InvenioRDM@zenodo

sfisher:
Set up the new sandbox invenio account to work with development.

Not all the uploads work right now, but it at least logs in and most API calls work.  Checking with Alex about what to do about file uploads and can create another PR and ticket to fix that, depending what he says.


sfisher:
Hi @ahamelers  thanks for looking at this.  I've now changed the credentials around in both files.  You can view some with `rails credentials:show`.  There is a default zenodo one in there for the old zenodo sandbox.  The dev environment in there has is overridden by the new invenio version.  Production is unchanged (overridden).

In the app\_config.yml, I went and structured it the same since it was confusing otherwise when it wasn't structured the same.

When we move fully to the Invenio version we can make it the default and just override in production.  Or if we want different application IDs for both dev/stage at zenodo then we can do that.

ahamelers (approved):
Thanks Scott, looks good!

PR #1262, opened 2023-06-30
Adding SUNY Fredonia and logo

sfisher:
Seems to be working at least on my dev server.

Screenshot 2023-06-30 at 4 18 30 PM


ahamelers:
@sfisher , when SVGs do not have a height and width set they are seen as 0 height and width and don't appear, it's weird. It can be fixed by adding a height and width to the SVG XML, but this works too!

ahamelers:
Thanks Scott!

PR #1263, opened 2023-07-01
Tenant for University of Maryland, College Park

sfisher:
Added this tenant and there was something weird about their SVG that made it not display in the browsers, so I converted it to a PNG.

Screenshot 2023-06-30 at 5 40 55 PM


ahamelers:
@sfisher , when SVGs do not have a height and width set they are seen as 0 height and width and don't appear, it's weird. It can be fixed by adding a height and width to the SVG XML, but this works too!

ahamelers (approved):

sfisher:
Huh, I never thought to look inside the SVG to set a height and width.

PR #1264, opened 2023-07-04
Fix null pointer issue

ryscher:


ahamelers:
Are there cases where users do not have IDs? All controllers using policies around the user id already have an additional before_action requiring a logged in user (see https://github.com/CDL-Dryad/dryad-app/blob/27373f36dbab5f08e674bf8f27c956ccb3654bef/app/controllers/stash_engine/dashboard_controller.rb#L3), do they all need this as well? Or, should we modify [the require_login and require_user_login actions](https://github.com/CDL-Dryad/dryad-app/blob/27373f36dbab5f08e674bf8f27c956ccb3654bef/app/controllers/stash_engine/shared_security_controller.rb#L12) to require user ids?

ryscher:
All users do have IDs. But there are cases where this code gets called and the `user` variable itself is `nil`. I didn't track them down source of these cases -- they may be due to bogus URLs. Mostly this change is just preventing an uncaught exception that will then trigger an error email to clog my inbox.

sfisher:
I'm not sure how this happens.  I know sometimes after a deploy to dev/stage my cookie or session is expired but I'm accessing a stale page or something and the ID is nil(?).  This is probably good to prevent errors.

PR #1265, opened 2023-07-04
Add a "base rate" calculation to tiered billing

ryscher:
Adds the final major piece of the tiered billing model, the calculation of a "base rate". Essentially, at the end of each quarter, we calculate the total cumulative submissions the sponsor has sent for the year, and this sets the per-dataset price for the current quarter. 

Currently, I'm testing this by running against actual data from https://github.com/datadryad/journal-payments/tree/main/shoppingcart and comparing the results against the calculator at https://docs.google.com/spreadsheets/d/1BqKpkuf1-IYgqqSscOuPHwMpWlbKnMvjOlZHm92ezk8/edit?usp=sharing, but long term it would be better to refactor this so the majority of the code is in a regular Ruby class, so it is easier to write tests in rspec.


sfisher (approved):
Thank you for adding this, Ryan.

I was able to run it and it produced reports as indicated in the directions.  I wasn't able to verify the costs against the Google sheet since I didn't have access (at least with whatever user stuff is going on with Google and how I'm logged in?).

It seems to work and isn't directly user-facing so seems OK to merge in to me.

PR #1266, opened 2023-07-06
Fixing problems with the issn not saving to the database

sfisher:
Didn't save when when autocompleting.  Some other process seemed to fill it in if going to another tab (from textual match?).

Before when I looked at my log output after autocompleting it only saved the form without an ID.  Now saves the form with the ID and you can see the query in logs like:

```
  StashEngine::InternalDatum Update (90.6ms)  UPDATE `stash_engine_internal_data` SET `stash_engine_internal_data`.`value` = '0363-6445', `stash_engine_internal_data`.`updated_at` = '2023-07-06 19:37:41' WHERE `stash_engine_internal_data`.`id` = 6175
```

It looks like the identifier field info for selecting it from the autocomplete list got changed to 'single_issn' (which appears to be a ruby function) rather than `issn` which is what the Name/ID json results list returns.  There were also some javascript key errors in React without it acting as a key.

### DIFF
--- a/app/javascript/react/components/MetadataEntry/PrelimAutocomplete.jsx
+++ b/app/javascript/react/components/MetadataEntry/PrelimAutocomplete.jsx
@@ -69,7 +69,7 @@ export default function PrelimAutocomplete({
   const nameFunc = (item) => (item?.title || '');
 
   // Given a js object from list (supplyLookupList above) it returns the unique identifier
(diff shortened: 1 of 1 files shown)


ryscher:
Thanks Scott. I must have introduced that bug by over-aggressively replacing the `issn` calls when I added the multi-ISSN functionality.

PR #1272, opened 2023-07-11
changes to pub updater as in 2259

sfisher:
closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2259 .

There is a lot in that ticket that was not unpacked fully or designed but I think this does the things asked for there.  We may want to modify the text at the top of the journal updater screen to explain what is happening better.

This also ends up modifying the import from CrossRef in the first step in the data entry since they use the same or entangled code.

## To test import of metadata from that page:

1. First find some good DOIs on CrossRef that you'd like to import like `https://api.crossref.org/works?query.bibliographic="pet snakes"&filter=type:journal-article&rows=100` (Change `query.bibliographic` value to whatever query you want to find some good articles to attempt importing metadata for).
2. Press CTRL/CMD-F to find `subject` in the json and find a record with a few subjects if you can.  Try to find the journal name and DOI in the record though it's sometimes hard to read. (JSON formatting plugins for your browser might help.)
3. Start a new dataset, choose `a published article` and put the journal name and doi in the form.
4. Click `Import article metadata` and you should see most of the metadata appear in the page.  You should see the subject items appear under the subject keywords section.  This is the new functionality and it didn't import keywords before.

## To test pub updater

1.  Do step 1 & 2 and find a good candidate article to pretend-enter a new dataset for.
2. Create a new dataset and enter the title and all the authors that the dataset has so it will be very similar. Use fake email addresses or whatever you need to do.
3. Continue filling stuff into the dataset and submit it.
4. After submission, go to the console and type `bundle exec rails publication_updater:crossref RAILS_ENV=development` (or replace environment to something else here).  It should run and try to find similar matches for datasets in Dryad on CrossRef and put the info into the database.
5. Go to the publication updater link in the admin area and you should be able to find a match between an article and the partial information you entered for the dataset.
6. If you click `accept` it should create a primary article link for that in the dataset, add publication name/issn in internal data (shown in first step of data entry) and populate new subjects that didn't exist. It also adds some stuff to the curation history.

## Testing other new functionality
- You can filter the list by search words.  This is an AND query (each word is required).
- You can click the export link and it exports what is essentially the same as what is shown on the screen (only all pages).

I find the DOI column a bit confusing here since in both cases (dataset and article) it's referring to the DOI recorded for the primary article.  You might think the DOI for the dataset row is the dataset DOI, but it's the DOI the dataset has recorded for the primary article.

I can look at instruction improvements and other stuff if you want.

The publication already sorts, I believe.  The way this is shown is wonky since there are compound rows to show changes and you can really only sort on one of them since the paired rows should be kept together.  In most cases they are the same or similar, though.

### DIFF
--- a/app/controllers/stash_engine/publication_updater_controller.rb
+++ b/app/controllers/stash_engine/publication_updater_controller.rb
@@ -4,35 +4,36 @@ module StashEngine
     before_action :require_user_login
     before_action :setup_paging, only: [:index]
 
-    CONCAT_FOR_SEARCH = <<~SQL
-        JOIN
-         (SELECT sepc2.id,
-          CONCAT_WS(' ',
-            sepc2.publication_doi,
-            sepc2.publication_issn,
-            sepc2.publication_name,
-            sepc2.title,
-            sei2.identifier,
-            ser2.title) AS big_text
-        FROM `stash_engine_proposed_changes` sepc2
-          INNER JOIN `stash_engine_identifiers` sei2
-            ON sei2.id = sepc2.identifier_id
-          INNER JOIN `stash_engine_resources` ser2
-            ON ser2.`id` = sei2.`latest_resource_id`
-       ) search_table
-        ON search_table.id = stash_engine_proposed_changes.id
+    CONCAT_FOR_SEARCH = <<~SQL.freeze
+       JOIN
+        (SELECT sepc2.id,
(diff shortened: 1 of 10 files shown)


ryscher:
Thanks, Scott, this is great!!! Everything I tested imports well, and the new search functionality is very helpful.

PR #1273, opened 2023-07-12
Ensure PPR end date is set when dataset is set to PPR

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2543


sfisher:
Thanks for tracking this down.  It looks ok to me, but I'll let Ryan comment first before merging anything.

ryscher:
Thanks for tracking this down. Should solve the problems.

PR #1278, opened 2023-07-18
remove automated DataCite counter runs until verification complete

sfisher:
I would like to disable this on Sunday.  I hope before I leave on the 31st that I can un-disable this for the next release.  I'll start the counter stats generation for July 1-15th today.


ryscher:
Sounds good.

PR #1279, opened 2023-07-18
Add user testing signup form and system email

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2604

### DIFF
--- a/app/views/stash_engine/sessions/feedback.html.erb
+++ b/app/views/stash_engine/sessions/feedback.html.erb
@@ -1,0 +1,67 @@
+<% @page_title = 'Give feedback' %>
+<%= javascript_pack_tag 'application' %>
+<script type="text/javascript">
+  ReactRailsUJS.mountComponents();
+  console.log("loaded rails javascript for data entry and react components");
+</script>
+<div id="page-columns">
+  <div id="page-content">
+    <h1>Direct the future of Dryad</h1>
+    <p>A few times a year, we ask volunteer user testers to provide feedback on sneak peeks at changes and new features to Dryad. Sign up here!</p>
+    <% if !current_user %>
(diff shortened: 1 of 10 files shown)


sfisher (approved):
Thanks!  This looks good to me.

It seems a bit odd that it's under the sessions controller which controls user sessions and logins, but it doesn't add a lot to the controller, so maybe ok if that makes sense for some other reason.

ahamelers:
@sfisher I put it there because it's getting and storing parameters in the session to keep track of them through the login process (to keep the url query values that will be sent into it for how the user reached the form). So this is where it made the most sense to put it, to me. Is there somewhere else you think it would fit better?

app/views/stash_engine/sessions/feedback.html.erb
```diff
@@ -0,0 +1,67 @@
...
+  console.log("loaded rails javascript for data entry and react components");
+</script>
+<div id="page-columns">
+  <div id="page-content">
+    <h1>Direct the future of Dryad</h1>
+    <p>A few times a year, we ask volunteer user testers to provide feedback on sneak peeks at changes and new features to Dryad. Sign up here!</p>
+    <% if !current_user %>
+      <p style="margin: 2em auto"><a class="t-login__buttonlink" href=<%="/stash/sessions/choose_login?origin=feedback#{"&m=#{params[:m]}" if params[:m].present?}#{"&l=#{params[:l]}" if params[:l].present?}"%>>Log in to Dryad</a> to import your information.</p>
+    <% end %>
+    <%= form_with(url: stash_url_helpers.feedback_signup_url, method: :post) do |form| %>
```
ryscher, line 14: Is there a reason to use `feedback_signup_url` here? When the url form is used, it generates an absolute URL, which doesn't work properly in testing environments (e.g., using the "local" environment, the form redirects to localhost, so the submission fails). This works correctly when using the relative path generated by `feedback_signup_path`.

ryscher (approved):
Other than that one comment, all looks good.

app/views/stash_engine/sessions/feedback.html.erb
ahamelers: @ryscher , in all honesty, it's very likely my text editor provided an autocomplete list with feedback_signup_url at the top and I just went with it 🤦

PR #1283, opened 2023-07-19
2668 cleanup non submitted uc ezid datasets over a year old

sfisher:
Can someone have a look at this task to be sure it seems reasonable?  It removes unsubmitted datasets over a year old.

I just want to double-check it before running I really run it.


ryscher:
Yes, looks good as far as I can tell. I did some spot-checking of the identifiers it would destroy, and they're fine to clean up.

PR #1284, opened 2023-07-21
2669 script to create lists of ezid registeredreserved items

sfisher:
This rake task script creates a csv file of all the non-dryad DOIs with status and EZID status (reserved/public) for further work.

It sounds like Maria maybe would like us to register all the reserved ones with minimal placeholder metadata for ease of transferring later (meeting next week).


ryscher:
Thanks for working through all of the details of this transition issue!

PR #1287, opened 2023-07-26
given list of reserved EZIDs turn them into registered ids with placeholder data

sfisher:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2682 .

We will create a list of IDs in EZID that are only reserved but not registered and run this on them to make them transferrable to DataCite.

I think the instructions in the rake task explains it pretty well.

Beware in development we are using test IDs and any IDs older than a couple of weeks make fail randomly since they purge test identifiers from their database occasionally and it causes various things to fail.

Easiest way to test is to create a couple datasets on the current development under a UC tenant.  These will be reserved-only before submitted.

Then run this script on them and it should create placeholder EZID records that are fully registered.

You can check at EZID with URLs like https://ezid.cdlib.org/id/doi:10.5072/FK2HT2SM2K .

I'm open to changing the placeholder data if people have better suggestions.

I'm also including the previous pull request code about [Turn off minting of new EZID/UC DOIs](https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2664) in here since I needed part of that configuration to do a check in the code.


ryscher:
Thanks, this should be very useful.

PR #1289, opened 2023-07-28
code that removes EZID entirely and only updates from DataCite

sfisher:
Don't merge this into the `main` branch until we are ready to move all our identifiers to DataCite and coordinate with Maria G.

- Removes EZID from code
- Removes indirection that could use EZID or DataCite
- Configuration is now in APP_CONFIG for all tenants (no per-tenant DOIs providers)
- Fixes tests that broke

Test by creating a new dataset (the ID should be created as a DataCite one) and when it's published, it should send it to DataCite successfully at a URL such as https://doi.test.datacite.org/dois/10.7959%2fdryad.3n5tb2s0 on their stage system.


ryscher:
Ok, looks good to me!

PR #1294, opened 2023-07-31
Additional Pundit scopes and variable fixes

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2502


ryscher:
This seems fine. I'm not certain whether I hit every context where these policies are used, but I'm unable to break it.

PR #1295, opened 2023-07-31
Activity notes from PubUpdater are credited to the system, not the author

ryscher:
When PubUpdater makes a change, the log entry should indicate the System user.


ahamelers:
Is it also possible to make sure the status isn't changed, or is changed back, and an email isn't sent? Assuming this ticket is related: https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2701

sfisher (approved):
Looks ok to me.

PR #1300, opened 2023-08-03
Correct name for the EEB CEDAR template

ryscher:
This name should have been changed with #1256.


ahamelers:
The last time I looked at a cedar template I noticed it was a bit buggy; the styles weren't quite right and the close button didn't work. Is that a concern for this, or can it wait?

ryscher:
It can wait. I will make a new ticket to fix the one biggest issue, which is the close button.

PR #1301, opened 2023-08-03
Add google captcha for user testing form

ahamelers:
Added for users not logged in.

Currently, if a (not logged in) user does not click the recaptcha, to show an error it will reload the page, losing their form entry, which is not ideal. We can fix this non-urgently in the future when editing the form for existing tickets.


ryscher:
Thanks for putting in the extra work on this!! Sorry you couldn't just copy the other one... I realize now that Salesforce does the validation piece for the other form.

PR #1305, opened 2023-08-12
Initialize Zenodo error classes

ahamelers:
Fix issue with Zenodo uninitialized constants

(cherry picked from commit 1c5d9da6678079ce1194a46ef8e27ca225af1371)

### DIFF
--- a/lib/stash/zenodo_replicate/retry_error.rb
+++ b/lib/stash/zenodo_replicate/retry_error.rb
@@ -1,0 +1,5 @@
+module Stash
+  module ZenodoReplicate
+    class RetryError < StandardError; end
+  end
+end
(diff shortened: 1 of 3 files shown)


sfisher (approved):
This looks good to me and seems close to equivalent to the previous version.

What problem was happening that made it not work?  I guess if something wasn't loading or requiring the ZenodoConnection file then it didn't work?  So maybe in the Rake task it didn't work?

ahamelers:
@sfisher it's a very minor issue, I got the big red error screen in local testing when sending supp files saying that ZenodoError was an uninitialized constant—I think due to [the move to zeitwerk in rails 6.0](https://guides.rubyonrails.org/6_0_release_notes.html)

PR #1306, opened 2023-08-12
Create, edit, and save required README files in markdown editor

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2410


sfisher:
Thank you for doing all this, @ahamelers .

It looks really beautiful and is so much better than our old process.

Filling in the extra README made me wonder how annoyed I would be if I was an author and I needed to re-fill the *title*  (maybe we could pre-fill if no readme exists yet) and also the related works.

I also found myself questioning inclusion of all the related works. I found myself thinking that maybe we should put a link to the DOI or Dryad URL in the readme so someone could get all the rich and up-to-date metadata in case they lost track of where they got the data files from.

I also wasn't sure how much some of the stuff overlaps with what they've already filled.  Like I can see people just putting in the abstract for "give a brief summary" section (whether that is meant to be that way or not).

I think these are all metadata questions and maybe I'm not the right person to answer them so I might be completely off base.  However, as a user if I interpret the readme as filling in information I've already entered and I have to retype it all again, then I'd be annoyed.

ahamelers:
@sfisher The curators are currently working on revisions to the template, and we've talked briefly about the possibility of auto-generating parts of it (like the title and DOI). I'll post your comments in https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2726 for consideration!

sfisher (commented):
This looked good to me before (aside from the comments I made), but now it seems to be not loading for me.  IDK if anything has changed or it's just my system.

Screenshot 2023-08-21 at 5 08 00 PM

Is this working for you still?  I also tried yarn installing libraries again but it seems to think some some things are missing.

ryscher:
@sfisher I had to do an extra `npm install`, but its working for me with the current code.

ryscher:
Looks great, and works well in every scenario I can anticipate.

PR #1307, opened 2023-08-14
Correct curation stats counting

ryscher:
Only count a dataset as submitted on a particular date if the submitted_date matches. This avoids erroneous counting for datasets like doi:10.5061/dryad.m63xsj47f, where the dataset was submitted on one date, but had some non-versioning activity that added new `submitted` statuses on another date.

### DIFF
--- a/app/models/stash_engine/curation_stats.rb
+++ b/app/models/stash_engine/curation_stats.rb
@@ -91,11 +91,14 @@ module StashEngine
       # for each dataset that received the target status on the given day
       cas = CurationActivity.where(created_at: date..(date + 1.day), status: %w[submitted])
       cas.each do |ca|
-        # include this dataset unless it has a previous resource that had been submitted
         this_resource = ca.resource
         found_dataset = this_resource&.identifier
         next unless found_dataset
 
+        # skip if the dataset was not first submitted on this date
+        next unless found_dataset.first_submitted_resource.submitted_date.to_date == date
+
+        # include this dataset unless it has a previous resource that had been submitted
         prev_resources = this_resource.identifier.resources.where(id: 0..this_resource.id - 1)
         prev_resources.each do |pr|
           found_dataset = nil if pr.submitted_date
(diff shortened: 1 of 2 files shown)


ryscher:
(partial fix for https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2705)

ahamelers (approved):
Thank you for finding this, Ryan! I spent a lot of time looking at it, but was difficult for me to pin down.

ahamelers:
@ryscher I've added another commit adding the same check to the peer review lookup (rather than the overly complicated check about other kinds of log entries that might have peer review status), and another check to populate_author_versioned meant to verify that the date is the first 'submitted' date for the resource. Please take a look

app/models/stash_engine/curation_stats.rb
```diff
@@ -232,7 +232,7 @@ def populate_author_versioned
       cas = CurationActivity.where(created_at: date..(date + 1.day), status: 'submitted')
       cas.each do |ca|
         # check if this was the actual date of submission for this resource
-        next unless ca.resource.curation_activities.where(status: 'submitted')&.first&.created_at&.between?(date, date + 1.day)
+        next unless ca.resource.curation_activities.where(status: 'submitted')&.first&.created_at&.to_date&.between?(date, date + 1.day)
```
ryscher, line 235: I think this needs to check for status to be 'submitted' or 'peer_review'

ryscher:
Thanks for starting work on this. I was going to suggest we review the other stats for the next sprint, and only change the "New to submitted" for now, since that is the one that shows the greatest variation. Note that by adding the new check, it now relies on the ResourceState and not just the CurationActivities, so the rspecs must explicitly set the value for ResourceState to 'submitted'.

app/models/stash_engine/curation_stats.rb
ahamelers: This is a check for tracking where some version was published or embargoed, and a later version is then re 'submitted'. I believe peer-review after publication is actually impossible... isn't it?

PR #1309, opened 2023-08-15
Log API user resubmitting resources, keep current_editor

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2701


ahamelers:
@ryscher I made this a draft so there wouldn't be a rush to get it in the current release, but it is in fact ready for review

ryscher:
Great, this is a minor change that should fix the problem.

PR #1311, opened 2023-08-15
Remove unneeded images

ryscher:
Extra cleanup for https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2728


ahamelers:
@ryscher I think there are at least a few more headshots no longer needed. Also there are some tenant logos in the public/images/ directory, which is not where they're used from. If you want to turn this into a general unused image removal pull request... 😄

sfisher (approved):
Looks good to me.  I tried the branch and didn't see any missing images since it should be serving them from the assets.

PR #1312, opened 2023-08-16
Upgrade to self-managing selenium-webdriver gem

ahamelers:
Needed so tests will run; otherwise there are Chrome driver errors

```
Selenium::WebDriver::Error::SessionNotCreatedError:
              session not created: This version of ChromeDriver only supports Chrome version 114
              Current browser version is 116.0.5845.96 with binary path /Applications/Google Chrome.app/Contents/MacOS/Google Chrome
```

Looked at managing driver versions with webdrivers gem again and got this message:

```
Post-install message from webdrivers:
Webdrivers gem update options

Selenium itself now manages drivers by default: https://www.selenium.dev/documentation/selenium_manager
* If you are using Ruby 3+ — please update to Selenium 4.11+ and stop requiring this gem
```

selenium-webdriver 4.11+ seems to do the trick (see https://www.selenium.dev/documentation/selenium_manager)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -95,8 +95,10 @@ gem 'omniauth-orcid', '~> 2.1', '>= 2.1.1'
 gem 'omniauth-rails_csrf_protection', '~> 1.0.1'
 gem 'omniauth-shibboleth', '~> 1.3.0'
 gem 'posix-spawn', '~> 0.3.15'
+gem 'pundit', '~> 2.3'
 gem 'rack-attack'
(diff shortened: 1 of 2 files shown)


ahamelers:
@ryscher or @sfisher : please check and merge into main as soon as possible!

ryscher:
Looks good, basic selenium tests work.

sfisher:
Thanks, Audrey.

PR #1314, opened 2023-08-18
Fix bug in DOI linking invitation check

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2739


sfisher (approved):
Thanks, Audrey.  This gets it anywhere in any of the strings of any curation activity notes.

PR #1316, opened 2023-08-18
Following up with users 3 times if an action is required & auto withdrawal

sfisher:
closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2542 .

To understand how to test, see the rspec tests.

This finds dates that an action_required started and each reminder.  It also sends out the emails as indicated by jess.


ryscher:
I was able to trigger all 3 emails by constructing a dataset and manipulating the timestamps in the CurationActivities. However, the final setting to withdrawn status did not happen. When I tried to use the same code in the console to trigger the withdrawn status, I received:

`app/mailers/stash_engine/user_mailer.rb:31:in `user_journal_withdrawn': undefined local variable or method `withdrawn_by_journal' for #<StashEngine::UserMailer:0x00000000029630>`

ryscher:
Thanks, these updates have fixed the problems.

PR #1318, opened 2023-08-21
Prevent resubmissions without changes

ahamelers:
Does not allow submission after curation with no changes

- If the previous status was 'action_required', shows an error.
- Otherwise, encourages the user to delete the unchanged submission and revert to the previous version.

Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2742


sfisher (approved):
This looks like good functionality to me.

I'm sure the curators will appreciate this.

ryscher:
Thanks! Looks good, and works well with my test objects.

PR #1320, opened 2023-08-22
Give update choices for pub updater (preprint, primary article, related article)

sfisher:
See comments on https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2667 .

This is easier to test against our development environment since there are already some crossref items populated into `stash_engine_proposed_changes`.  You can also reset items in that table by setting `approved` and `rejected` both to 0 and it will show up in the pub updater again.

- Add `xref_type` to the database and populates it from xref for newly processed items.
- This may be a hint at the type of work, for instance a preprint shows `posted-content` sometimes for a preprint (?)
- Now gives 3 options for acceptance type: primary (article), related (article) and preprint.  All three have the choice of updating with or without the metadata.
  - With the metadata it does as before (updates the issn, publication name, subjects and related identifier)
  - without metadata it only updates the related identifier relationship

I'll add another PR for the other related ticket about other, more minor things to change about the publication updater.

### DIFF
--- a/app/controllers/stash_engine/publication_updater_controller.rb
+++ b/app/controllers/stash_engine/publication_updater_controller.rb
@@ -63,11 +63,12 @@ module StashEngine
     # rubocop:enable Metrics/AbcSize
 
     def update
+      byebug
       respond_to do |format|
         @proposed_change = authorize StashEngine::ProposedChange.find(params[:id])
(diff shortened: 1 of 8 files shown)


ahamelers (approved):
Looks great to me! Thanks, Scott

PR #1321, opened 2023-08-23
Disallow overwrite of curated metadata with import

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2745


ryscher:
Thanks for adding this! I'm sure it will save time for the curators.

PR #1323, opened 2023-08-24
Reinstates weekly counter processing in August

sfisher:
August processing is running manually right now (maybe be done by sometime next week).

This reinstates the cron since we don't need to have a strict partial month for August (like we thought we did in July for validating the tracker with DataCite).


ryscher:
Good to get this back to normal.

PR #1324, opened 2023-08-25
Addresses publication updater issues curators raised

sfisher:
See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2720 .

Addresses the items raised in this PR and the previous one.

- System matching to Withdrawn dataset
  - Excluded from the query
- System makes match with no apparent changes
  - Skips matches without apparent changes.  I believe these may have originally been different when the proposed change was imported and the dataset was updated sometime after the import from CrossRef.
- System wants to replace one ISSN with another (same journal)
  - This seems a curation call about which ISSN is appropriate.  It can be rejected or now there is the option to just update the related DOI if desired without updating the rest.
- System wants to remove ISSN
  - Items with blank ISSNs and no other differences are skipped.
- System wants to replace correctly primary article with unrelated article
  - From previous pull request, curators can now choose to update as either primary (article), related (article), preprint (article) if it is one of these instead or reject.
- Missing (newly added to pub updater) metadata
  - Added additional script to reprocess items that haven't been accepted or rejected yet and it will remove the existing item and re-check crossref to populate the keywords and the CrossRef `type` if they are available and put the record back in again with the newer items (If it's still available to lookup from CrossRef)

### DIFF
--- a/app/controllers/stash_engine/publication_updater_controller.rb
+++ b/app/controllers/stash_engine/publication_updater_controller.rb
@@ -27,7 +27,9 @@ module StashEngine
     # the admin datasets main page showing users and stats, but slightly different in scope for curators vs tenant admins
     def index
       proposed_changes = authorize StashEngine::ProposedChange # .includes(identifier: :latest_resource)
-        .joins(identifier: :latest_resource).where(approved: false, rejected: false).select('stash_engine_proposed_changes.*')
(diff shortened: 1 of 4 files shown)


ryscher:
Thanks Scott! This should update the concerns. I do wonder whether we should add a cron entry for the rescan task.

PR #1325, opened 2023-08-28
Collapse landing page README section

ahamelers:
Also update submission review screen to look more like landing page


ryscher:
Looks great, works well for me. Thanks for this last-minute change!

PR #1327, opened 2023-08-29
Author submitted dataset with no data files -- disallow

sfisher:
As far as I can tell, the author must've had two different tabs or windows open.  https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2717

To reproduce.

- Open one window with all data correctly filled.
- Open the same dataset in another window, but then remove the data files after the 1st window says it's fine.
- Submit in the first window.

it should now prevent submission if all requirements aren't filled.  It checks again before submission and stops them at that point, also.

I also started having lots of zenodo submission issues on my local machine with uninitialized constant errors when trying to require dependencies.  I spent hours trying to figure out and did updates.  I finally got it to work by requiring the module above the one that had errors.  But then all tests failed here.

So . . . reverting some things and 🤞 .


sfisher:
I think the uninitialized constant errors are fixed in a separate pull request now.

ryscher:
This tested well in my personal testing. However, it took four tries for it to pass the rspec tests. Something somewhere is dependent on the order of the tests.

PR #1329, opened 2023-08-29
Fix for uninitialized constant/zeitwork

sfisher:
We have some classes with a pattern from log ago where there is a module that requires all submodules in a subdirectory.

This seems to interfere with Zeitwork somehow or sometimes.  After a bunch of trial and error, I found removing this require fixes the problem.


sfisher:
I'm going to merge this and then deploy on dev to be sure it works there, also.

PR #1331, opened 2023-08-30
Add new tenant

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2724


config/tenants/neomed.yml
```diff
@@ -0,0 +1,49 @@
...
+  tenant_id: neomed
+  identifier_service:
+    provider: datacite
+    prefix: "10.7959"
+    account: <%= Rails.application.credentials[Rails.env.to_sym][:datacite_username] %>
+    password: <%= Rails.application.credentials[Rails.env.to_sym][:datacite_password] %>
+    sandbox: true
+  authentication:
+    strategy: ip_address
+    ranges: [ "140.220.0.0/16", "199.18.154.122/123", "199.18.157.82", "132.174.254.221" ]
```
sfisher, line 18: Hi Audrey,

Is the one with `199.18.154.122/123` really correct?  I could be wrong, but that doesn't seem like the typical CIDR notation I usually see and is an unusual number (though maybe it's correct).

sfisher (commented):
If that CIDR is right, it looks good as far as I can see.

config/tenants/neomed.yml
ahamelers: No, I think you're right! I don't know why I did that, excellent catch! Will fix.

PR #1334, opened 2023-08-30
Handle non UTF-8 README files

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2757


ryscher:
This works, and tests well with all of the problematic files that triggered alerts today.

PR #1338, opened 2023-09-05
Allow curators to create and publish dataset collections

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2731

- Curators (limited curators and above) can create dataset collections with URL /stash/resources/new?collection
- Collected Dryad datasets are added as related works (relation_type: haspart, work_type: dataset)
- No other workflow changes—currently these still need to go through curation, but this step can be entirely fake
- No payment required for these (no new data)


app/views/stash_datacite/related_identifiers/_collection.html.erb
```diff
@@ -0,0 +1,12 @@
+<% if review.collected_datasets.count.positive? %>
```
ryscher, line 1: I'm confused by the `review` -- not sure what this is meant to hold or where it is set.

app/views/stash_datacite/resources/_show_collection.html.erb
```diff
@@ -0,0 +1,23 @@
+  <%= render partial: "stash_datacite/titles/show", locals: { title: review.title_str, highlight: false } %>
```
ryscher, line 1: Same with the `review` here -- is this only applicable on the review page?

ryscher:
1. It would be good to make a documentation page to outline how the collections work, and give the URL for creating them.
2. When I'm made a "regular" dataset, the submission form's section for Related works shows the version for collections, with "Please list all the datasets in the collection" (note this happened directly after creating a collection; not sure whether there was something being cached.)
3. The landing page for my collection did not show the datasets.

app/views/stash_datacite/resources/_show_collection.html.erb
ahamelers: @ryscher the _show_collection.html.erb file was copied and edited from the _show.html.erb file, which uses review in exactly the same way.

app/views/stash_datacite/related_identifiers/_collection.html.erb
ahamelers: This is a local set in `app/views/stash_datacite/resources/_show_collection.html.erb`
```
 <% if review.collected_datasets.count.positive? %>
    <%= render partial: "stash_datacite/related_identifiers/collection", locals: {review: review} %>
  <% end %>
```

and `app/views/stash_datacite/resources/_review.html.erb`
```
<% if @resource&.resource_type&.resource_type == 'collection' %>
      <%= render partial: "stash_datacite/related_identifiers/collection", locals: {review: @review} %>
    <% end %>
```

app/views/stash_datacite/resources/_show_collection.html.erb
ahamelers: I believe it's initially set in the landing mixin (`app/controllers/stash_datacite/landing_mixin.rb`)

```
    def review
      @review ||= StashDatacite::Resource::Review.new(resource)
    end
```

As usual, I fit my changes to existing structures. I don't know why a lot of this stuff is in review, but it's where the related_identifiers are set, and since the collected datasets are a subset of those, it's where I set them up as well.

ahamelers:
> The landing page for my collection did not show the contained datasets.

@ryscher On what server? Did they appear on the review page?

ryscher:
I was testing on my personal server. I redeployed this code on dev. It resolved the issue of not displaying collection contents. However, the issue of displaying the "wrong" section for related works is still occurring.

ahamelers:
There were too few equal signs in an if statement, so instead of checking if the type was a collection, I set the type to collection 🤦 

It's a shame rubocop doesn't check the erb files!

PR #1339, opened 2023-09-05
Correct dashboard ordering by last update

ahamelers:
Order of items on user dashboard was by identifier updated_at. What is displayed on the dashboard is resource/last curation activity update. Corrected the ordering to match.


ryscher:
Thanks, it's always great to improve consistency.

PR #1340, opened 2023-09-05
Adds config for mrt bucket and refactors S3 wrapper library

sfisher:
Adds an additional bucket config to the yml files.

Refactors all the calls to the S3 wrapper library we use to be an instance rather than a static object. Default of `new` is to create based on the temp storage bucket.  Otherwise pass a parameter for the bucket and it will use a different one.

A bunch of tests broke and had to add different ways to update the mocks.

I believe I got all the places where this library is called and it's a lot more places than I thought.


ryscher:
Thanks, Scott. This looks good, and has no noticeable effect on current functionality.

PR #1341, opened 2023-09-06
Ask users to describe file edits between versions

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2299


sfisher (approved):
I tested this out and looks good to me.  I guess you have to submit the dataset first before it shows up, right?

looks good.

PR #1344, opened 2023-09-08
Set current_editor to submitter when API request makes new version

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2701


sfisher:
Thanks for the PR.  This looks good and is straightforward and will allow users to edit things that an external API does.

PR #1351, opened 2023-09-14
Dryad owned bucket for dev deposited to by Merritt

sfisher:
Updating to use the cross-account bucket and merritt collection so we can really test the direct S3 functionality in our development environment since we'll need this.


ryscher:
Thank for getting this new collection set up!

PR #1352, opened 2023-09-15
Instructions for reconfiguring for cross-org access

sfisher:
Accsss without ACLs so both organizations can access content in the bucket.


ryscher:
Thanks, Scott. This makes the process clearer.

PR #1353, opened 2023-09-15
Change the zenodo sandbox on dev

sfisher:
They said we should use this different URL instead.

I'll deploy to dev and restart the delayed_job and try some new submissions to be sure they go through to zenodo.


sfisher:
It seems like this now has this following error at the new server address:https://zenodo-rdm-qa.web.cern.ch

```
{"message"=>"Referer checking failed - no Referer.", "status"=>400} for 
http.post https://zenodo-rdm-qa.web.cern.ch/api/deposit/depositions
{"message"=>"Referer checking failed - no Referer.", "status"=>400}
```

IDK if this is a temporary problem or if we need to change something.

ryscher:
Thanks for following up with Alex. I'm going to merge this despite the issue, so we can keep testing.

PR #1354, opened 2023-09-19
Expand journals on the journal page

ryscher:
Curators were having difficulty reconciling journal information for journals that send metadata (integrated), but do not have an active sponsor. 

This allows journals to show on the `/journals` page even if there is no sponsor present.


sfisher (approved):
Makes sense why they were not showing up.  Thank you, Ryan.

PR #1357, opened 2023-09-20
Makes items download with S3 presigned URLs instead of Merritt ones

sfisher:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2637 .

I believe it all works and all tests are changed and passed.  I'll test the Zenodo submission/copying stuff a bit more on the server and I hope the new InvenioRDM server is working correctly today.


sfisher:
Merritt is giving some unexpected behavior.  Rather than setting to `created`, a newly uploaded file is treated as a pointer to a previous version for the following case (from david):

```
checks against all files with the same name, matching length and digest.
```

This means, even when we purposefully upload a file again, it may not be stored in the version we think it is stored in for merritt.  :-/

sfisher:
I'm seeing problems with this for README.md files where it is not changed between versions, but uploaded again every time.  Also if I delete and upload the same filename and contents again.

This is a real annoyance since we don't calculate digests in upload, though maybe we can one from amazon.  The indication of "created" therefore cannot be trusted as he marker for the first in a chain of files if the contents are the same as some other previous file.

We also don't calculate the sha256 since we don't insert ourselves in the file upload and it goes into S3 in chunks that we sign for upload from the user's browser.

IDK how we're going to get around this issue.

ryscher:
This looks great, except for the issue Scott notes above. I have tested this PR, and everything else works fine.

I have some ideas about how we can address the issue, which we can discuss...

ryscher:
Thanks, Scott! This works well for me now. Glad you were able to find the correct files without doing anything too complex.

PR #1359, opened 2023-09-22
Fix error with rendering resources in the API

ryscher:
We were getting a lot of emails about this as bots hit API pages. The rendering would fail for authors who did not have an affilation.


ryscher:
(Closed because I accidentally pulled in a lot of other things....)

PR #1360, opened 2023-09-22
Fix error with rendering authors in the API

ryscher:
We were getting a lot of emails about this as bots hit API pages. The rendering would fail for authors who did not have an affiliation.

(Fix has been applied on production.)


sfisher (approved):
Thanks for the null value updates.

PR #1363, opened 2023-09-27
Use ISO dates in admin CSVs

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2827


spec/features/stash_engine/curation_activity_spec.rb
```diff
@@ -107,9 +107,9 @@
         # but it doesn't matter as long as it is synced with the index below
         csv_parts = csv_line.split(',')
```
ahamelers, line 109: This change is because the dates no longer contains commas

sfisher (approved):
Thanks Audrey,

I tested this out and both the pages and the CSVs look fine and probably smart just to leave everything in UTC.

PR #1364, opened 2023-09-27
Edge case for items manipulated in Merritt outside of what's recorded in our database

sfisher:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2843 .

I tested from the production server in the console based on new code.  I tested the failure examples from the ticket and this seemed to fix retrieval for all of them.

I think most of the datasets like this will have been migrated from one collection to another, had ingest problems or some other special handling to make them so that there are versions in Merritt that are not accurately represented in our database.  Probably the usual giveaway would be that the version and merritt_version numbers are different in the versions table.

It only really uses this method after the other ways to walk back through the history from our database have failed.

### DIFF
--- a/app/models/stash_engine/data_file.rb
+++ b/app/models/stash_engine/data_file.rb
@@ -9,7 +9,9 @@ module StashEngine
       "#{resource.s3_dir_name(type: 'data')}/#{upload_file_name}"
     end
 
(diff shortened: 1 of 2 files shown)


ryscher:
Thanks, Scott. This looks fine. I didn't try to construct broken objects to test, but it sounds like you have tested enough that it will work in practice.

PR #1366, opened 2023-09-29
Adds name to mapping table since it makes it easier

sfisher:
to update things later.


ryscher:
Thanks, this update makes sense.

PR #1368, opened 2023-09-29
Charles R. Drew university tenant and changes to author_match strategy

sfisher:
It seems like payment was not calculating correctly for the author match strategy.  This corrects that problem and adds tests.  The login for author_match was already working.

The strategy means that institution payment is allowed if 1) the depositor chooses the institution with that strategy on first login, 2) The dataset is created under that same tenant, 3) the sets of author and tenant RORs have an intersection.  (At least one author affiliation is in the set of tenant rors).

You can test by doing something like 1) setting C.R. Drew tenant for user, 2) create new dataset with that tenant, 3) Select the ROR affiliation for Charles R. Drew -- then the last page should show deposit covered.   If you change number 3 to non-CRD affiliation it should become non-covered.


ryscher:
Thanks, Scott! This looks good.

PR #1369, opened 2023-10-02
Give better response and email to "not found" file errors from initial deposit mismatches

sfisher:
Also documents queries for troubleshooting these.

See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2848 .

These issues were caused by deposits that happened in a later version in Merritt/S3 than where we requested the deposit.  Often you can tell that some hacking went on to correct something because the dryad and merritt versions do not match in the versions table.

This returns a 404 instead of a server error and emails us the information instead of sending us the generic report.

You can test with the dataset https://dryad-dev.cdlib.org/stash/dataset/doi:10.7959/dryad.8931zct2 and the NIOO.svg file.  It is made to be bad like these so that the deposit took place into S3 in V2, but our database indicates V1 deposit.

I think it's somewhat important to have our "created" states line up with when files were actually deposited in Merritt or at least not point to an S3 version before something was actually deposited.

### DIFF
--- a/app/mailers/stash_engine/user_mailer.rb
+++ b/app/mailers/stash_engine/user_mailer.rb
@@ -89,6 +89,7 @@ module StashEngine
       logger.warn("Unable to report update error #{error}; nil resource") unless resource.present?
       @zenodo_error_emails = APP_CONFIG['zenodo_error_email']
       return unless resource.present? && @zenodo_error_emails.present?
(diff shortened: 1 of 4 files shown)


ryscher:
Thanks, Scott! It's very useful to have these queries handy for pulling out file information.

PR #1370, opened 2023-10-03
Remove dataset splitting instructions

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2847


sfisher (approved):
Thanks for this update.

PR #1372, opened 2023-10-04
Tweaks to EZID placeholder data submission

sfisher:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2856 .

I also ran this against production so now all reserved items have some placeholder metadata. This should allow the transition to proceed whenever it is ready since no new EZID identifiers are being created by our system.

I couldn't replace the publication year with `:unas` or any of the other special datacite values and that was causing failures, I discovered after some trial and error.


ryscher:
Thanks, Scott. I think we ran into the issue with `:unas` long ago regarding embargoed datasets. Martin Fenner advised us to use a placeholder like 9999, but it should be fine to leave it as 2023 for these datasets.

PR #1374, opened 2023-10-05
Fix bug in locating tenants for shopping cart reports

ryscher:
And update documentation to be clearer about which files to use for base rate.


sfisher (approved):
That change makes sense (Tenant instead of Tenants) though I didn't test everything fully.  I believe this is good.

PR #1376, opened 2023-10-09
Add automatic refresh of token for Zenodo

sfisher:
From id and secret.

See the ongoing ticket https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2829 .

I've tested this against zenodo and it does seem to allow us to create a new blank deposit, but adding metadata seems to break with them giving us a 500 error.

This is more like how many other people do OAuth2, but it does mean we need extra code to take care of renewals and to cache the information about our token for the hour.

I also added a global state table for things like this where there is probably going to be one instance (of the token) that is used across the application and can be read and written to by the application which is likely running across different servers.  So a way to share temporary state besides the configuration files which are more permanent.

Once we see this working on dev we will want to add our id and secrets to the other environments and need to deploy around the same time they do for the change.


sfisher:
I suppose we do not want to merge this until Zenodo is on the new system and we're sure it works against Invenio, so will switch to draft once tests pass.

PR #1378, opened 2023-10-10
Static content adjustments

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2863
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2871


ryscher:
Thanks, Audrey. This fixes the issue of finding the codebase. 

Note that the information for Stanford is "wrong", because it doesn't match the ROR name, but it was already "wrong" before, so this is no worse. I'll make a new ticket to fix the actual problem.

PR #1382, opened 2023-10-11
Intermittently failing tests fix

ahamelers:
Using the method by which keywords were created in the tests, very occasionally there were unintentional duplicates, which were then purged, which then blocked submission tests because of insufficient # of keywords (this is conveniently demonstrated below in the test of the first commit)


ryscher:
Glad you found this!

PR #1383, opened 2023-10-11
Remove merritt presigned URL attempt for 'created' README

ahamelers:
There's a new bug when the README editor attempts to load content from files. It seems like this is being caused by an attempt at downloading from a merritt bucket that (I think?) the files are no longer actually going to (please correct me if I've got the wrong end of the stick there!).

If this corrects the bug and we could get this merged and released that would be great, as it is preventing some users from proceeding.


sfisher (approved):
Looks fine to me.  Let's try it on stage.

PR #1384, opened 2023-10-12
Intermittently failing tests fix

ahamelers:
The keyword thing (#1382) is still happening occasionally. I think my previous fix created a unique list of words without ensuring each item in the list was unique within the list 🤦 

I've tested this new method over and over and it hasn't failed yet!


ryscher:
Ok, looks good. Hope this finally fixes it!

PR #1386, opened 2023-10-17
Adds grouping for ROR instead of CrossRef funders for NIH

sfisher:
We needed to add our special sauce for the dialog that forces people to choose a directorate.

I tested it out and it seems to work in the UI still.


ahamelers (approved):
Thanks Scott!

PR #1391, opened 2023-10-19
XRef Funder ID to ROR transition steps and changes

sfisher:
This documents exact tasks, queries and procedures that need to take place to do this ROR transition.

I've tested all these on our development system and they seemed to work ok.  I also made a table copy of our current data and tried the transform query on it in our production database.

I needed to make some changes to the way we're calling to generate DataCite XML by calling it directly rather than going through the Merritt packaging infrastructure since it had some embedded validation that didn't work correctly with ROR.


ryscher:
Thanks Scott, this looks great.

PR #1392, opened 2023-10-20
Initial updates to shibboleth docs

ryscher:
For now, renaming the directory so it makes sense, and removing a lot of outdated text.

Text will be updated as we complete more portions of the shibboleth transition.


sfisher (approved):
Thanks for doing this cleanup, Ryan.

PR #1395, opened 2023-10-23
Do not allow PPR after any version has been curated

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2899

Changes PPR display check to be based on submission status of submission and not just individual version


sfisher (approved):
This looks good to me.  That sounds like a good workflow fix.

PR #1396, opened 2023-10-23
Fixes problems with the Zenodo re-submission queue dashboard

sfisher:
See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2900 .

This does the following:

- Makes all links in the table open in new windows
- Removes the error info as a column since it didn't display usefully and was making the page not load with the long text and delays getting it from the database.
- Makes the *resend* button not navigate away from the page but instead resubmit by AJAX/JS.
  - The results of the resubmission show in the *state* column instead of on a new page (prerequisite, in runner, ...submitting)
  - All this makes it much easier to go through individual items to resubmit without any big delays and without losing your place in the table.

This should make it possible to get the page to load quickly and also go through and resubmit things without confusion and delay as we hassle trying to get Zenodo stuff to go back through.

BTW, if you need to test there are still things in our development database stuck from our zenodo issues.

### DIFF
--- a/app/controllers/stash_engine/zenodo_queue_controller.rb
+++ b/app/controllers/stash_engine/zenodo_queue_controller.rb
@@ -4,12 +4,11 @@ module StashEngine
     helper SortableTableHelper
     before_action :require_user_login
 
-    ALLOWED_SORT = %w[id identifier_id resource_id state updated_at copy_type size error_info].freeze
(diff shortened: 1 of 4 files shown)


ryscher:
Thanks, Scott, this looks good. I tested the very basics, and all seems fine.

PR #1397, opened 2023-10-24
Resets Zenodo presigned to use old tokens

sfisher:
The old tokens do not work on any of the test environments that they've told us to use, but maybe it works in production.  Zenodo says to use the old tokens for presigned RAT urls.

This gets just the presigned URLs to use the old tokens and maybe it works on production.  Your guess is as good as mine since none of the staging/testing servers had old tokens working and they had information inside the UI about how to get tokens through OAuth2 which is different.

In any case, if they don't work then the presigned link at a URL like http://localhost:3000/stash/share/mFs3BVKuYSKL_CUAP3W9wVtiqMuDi7XISZXkgV8KvsA just returns an "unavailable" message when it fails, so I guess no worse off then it is currently.

Let me see if lots of tests break.


ryscher:
This looks good to me....we might as well try it.

PR #1400, opened 2023-10-27
Initial configuration for v3 servers

ryscher:
- Add rails environments for v3 servers
- Force webpack to use the version that we're running on the CDL servers (if npm loads a default version, it fails with a cryptic error)


ryscher:
@ahamelers Can you take a look at this? It is stuck on the github issue we discussed. Is the best approach to make a generic Github user? Or is there another solution to allow anonymous access?

sfisher:
Interesting that PR 1401 passes tests and this one doesn't.  (Though I didn't diff the code to see the differences).

I would've thought it would work ok as a http url (rather than a `git@` one).  I haven't diffed to see what is different, though.

sfisher:
I merged main back into this branch and tried to resolve yarn.lock problems.  I hope this makes it so that tests run and pass since only 1/5 were working.

sfisher:
I also notice the encrypted file changed which is hard to merge if there are any conflicts.  I extracted the text and did a diff against the main branch and it all looks good.  I know it should raise conflicts, but I'm always nervous when two people make changes to this file in the same sprint.

PR #1403, opened 2023-10-31
Set single ISSN from journal autocomplete

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2911


ryscher:

PR #1406, opened 2023-11-02
Create an option for assembly of zip files within the browser

ahamelers:
From https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2787

See [comment in the ticket](https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2787#issuecomment-1791241092)

I've also changed the order of the dataset downloads to prioritize the dataset version individual file downloads, and updated the UI and text to make it more clear that 1) those are expandable sections 2) that contain file download links


ryscher:
I'm worried that this will be confusing for users:
Screen Shot 2022-09-13 at 1 11 30 PM

They will say "what happened to the other versions?"

ryscher:
I'm worried that this will be confusing for users:
Screenshot 2023-11-06 at 12 05 42 PM

They will say "what happened to the other versions?"

ahamelers:
> I'm worried that this will be confusing for users: Screenshot 2023-11-06 at 12 05 42 PM
> 
> They will say "what happened to the other versions?"

Options:

- Use index of published versions as version numbers in UI and zip file name
- Instead of referring to a version number at all, refer to the date, maybe like:
    - Oct 20, 2023 version files
    - doi_10_etc__v20231020.zip
- Keep as is

ryscher:
> * Instead of referring to a version number at all, refer to the date, maybe like:
>   * Oct 20, 2023 version files
>   * doi_10_etc__v20231020.zip

Ideally, we should do user testing to determine what the users would really like to see. In the meantime, I like referring to the date. This allows us to clarify the UI a bit, without yet committing to a "new" version number that we would need to support forever.

ryscher (commented):
Overall looks good, and functionality works well.

sfisher (approved):
Wow, this is really cool that this works so well and saves a lot of server resources.

I tested on Chrome, Safari and Firefox on MacOS.  They all worked beautifully.  I believe these are all the latest release versions since we need to keep them up to date for security.

I also tested on my Android phone using Kiwi browser (a fork of mobile Chrome) and it worked great.  I think mobile Chrome would also be fine then.

The only things I thought were a bit weird were:

- The download dialog is a bit different.  On Chrome I didn't really notice it downloading right away, but if i look specifically at the downloads then I see it is downloading.
- In Safari, it seemed to download to a folder instead of to a zip file!  IDK if there is some auto-extraction magic going on with MacOS, but I saw a folder with all the files, but there just wasn't a zip file on my system.  I think that's ok since the files all arrived.
- Firefox made the download more obvious.
- Safari prompted me with a message first in the browser "Do you want to allow downloads on "\<domain name\>"? You can change which websites can download files in the Websites section of Safari Settings.
- Mobile chrome made it quite obvious it was downloading.

The download does seem a bit different than a normal download.  I wonder if it would make sense to do something in the dialog that comes up to make the download more obvious.  Like maybe it could say something like "Adding file \<name\> to your zip download" for each file.  Or maybe that makes no sense and would just confuse people.  The fine tuning is more of a minor thing. This is really up to whatever you and Ryan think or we could ask users eventually.

It is super impressive how well this works.  I know I was thinking of this for selfish reasons so that we don't have to pay so much to assemble zip files and set up all the extra infrastructure, but it seems to work really well on all the modern browsers I tried.

I suppose we could also warn that zip downloads may not work on older browsers.  (or just add a footnote that you must use an up-to-date browser for download and security purposes).

I didn't notice changes to COUNTER, but I think maybe it's the same?  Probably when the download starts it should count a hit, just like the other downloads do.  I think right now we're counting in both Ruby logs and also by hitting an AJAX endpoint at datacite to make a count.  I'm hoping soon it will all just be the AJAX endpoint.

ahamelers:
> * The download dialog is a bit different.  On Chrome I didn't really notice it downloading right away, but if i look specifically at the downloads then I see it is downloading.
> 
> * In Safari, it seemed to download to a folder instead of to a zip file!  IDK if there is some auto-extraction magic going on with MacOS, but I saw a folder with all the files, but there just wasn't a zip file on my system.  I think that's ok since the files all arrived.
> 
> * Firefox made the download more obvious.
> 
> * Safari prompted me with a message first in the browser "Do you want to allow downloads on "<domain name>"? You can change which websites can download files in the Websites section of Safari Settings.
> 
> * Mobile chrome made it quite obvious it was downloading.

@sfisher amazingly all of this should be because of individual browser settings and preferences! In my own testing I found the safari one especially weird (and managed to break it during archive expansion a couple times before I turned that setting off 🙄). Whether or not a file download permission or location selection prompt is shown will depend on a preference set in each browser (usually, quite granularly, an option will be set for zip file type downloads specifically).

Have you tested downloading live site zip files (either from _within_ datasets, or pre-assembled) on the same browsers? You _should_ 🤞 have the same experiences in each with the existing and this new method!

ahamelers:
> I suppose we could also warn that zip downloads may not work on older browsers. (or just add a footnote that you must use an up-to-date browser for download and security purposes).

@sfisher I did add a message like this on the share page (in addition to offering the individual files), but on the landing page I just remove the zip download button (the same as when a dataset is too large). Mainly because the sidebar of the landing page is so small and I didn't want to clutter it—something we could reconsider if we move the data files into the main content!

You can test what it's like, on either screen, on a browser without this service enabled easily in Firefox—in addition to requiring a pretty modern browser, service workers require https, and aren't enabled in Firefox private browsing.

ryscher:
Thanks for the updates.

PR #1410, opened 2023-11-07
Improve element focus indicators

ahamelers:
Make focus look less like an error marker and set different focus for links for mouse and keyboard users


ryscher:
Ahhh, that's much nicer!

PR #1415, opened 2023-11-08
Revert "Create an option for assembly of zip files within the browser"

sfisher:
Reverts CDL-Dryad/dryad-app#1406

I think that this is best to revert until we can fix some cases we didn't anticipate.


sfisher:
I'm going to merge this for now and we can revisit under less time pressure.

PR #1417, opened 2023-11-09
Handle multiple file presigning with one request for file zip

sfisher:
OK.  I think this now fixes the problems with too many requests for zips with lots of files.

Changes from what Audrey originally had.

- The form for the zip download button now just contains the resource_id that is for download.
- The worker queries `/stash/downloads/zip_assembly_info/${resource_id}` to get all the information it needs (names, sizes and presigned URLs).
  - To prevent people from hitting the URL to get presigned downloads for anything, the session stores recent resource ids of things that the UI has shown that user for download (and has determined they're able to download) and the download is based on this list.  Also works for the PPR-preview stuff.
  - We also don't want to generate pre-signed downloads for every possible file just because someone looks at the landing page (and embed it into the hidden fields in the form).  These only get generated from the controller when the service worker actually tries to do the download. 
- I tested on dev with the dataset https://dryad-dev.cdlib.org/stash/dataset/doi:10.7959/dryad.34tmpg50 which has around 200 files and it semeed to work ok.

I'll do the Counter updates in another PR to work with this new download.

---

PS.  I notice things with many files take a bit longer to get started since it has to create so many presigned URLs.  I also noticed submitting these datasets with so many files have a number of other problems which I added a backlog ticket about (the review page, the submission button, crazy numbers of loads of curation history or other stuff).  

PPS. This was tricky to get the code back because of the revert commit, so I had to follow the instructions at https://stackoverflow.com/questions/27852143/how-to-pr-and-merge-again-after-reverting-pr-using-github-revert-button#:~:text=To%20solve%20this%20problem%20you,and%20create%20a%20new%20PR. to unrevert (that's the new-zip-updates2 branch that got merged).

### DIFF
--- a/app/assets/stylesheets/stash_engine/ui.css
+++ b/app/assets/stylesheets/stash_engine/ui.css
@@ -1009,7 +1009,7 @@ datalist {
   color: #3c763d;
 }
 
-*[class^=o-download__] {
+*[class*=o-download__] {
   display: -webkit-box;
   display: -ms-flexbox;
   display: flex;
@@ -1021,7 +1021,7 @@ datalist {
   border: none;
   background-color: #98cae2;
 }
-*[class^=o-download__] #download_icon {
+*[class*=o-download__] #download_icon {
   position: relative;
   display: block;
   width: 20px;
@@ -1030,7 +1030,7 @@ datalist {
   font-family: FontAwesome;
   font-size: 20px;
 }
-*[class^=o-download__]:hover {
+*[class*=o-download__]:hover {
   opacity: 0.85;
 }
 
--- a/app/controllers/stash_engine/downloads_controller.rb
+++ b/app/controllers/stash_engine/downloads_controller.rb
@@ -49,7 +49,7 @@ module StashEngine
           @resource = Resource.find(params[:resource_id])
           info = @resource.data_files.present_files.map do |f|
             {
-              size: "#{f.upload_file_size}",
+              size: f.upload_file_size,
               filename: f.upload_file_name,
               url: f.s3_permanent_presigned_url
             }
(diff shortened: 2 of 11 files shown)


ahamelers:
Hi @sfisher , thinking about it, I think it's still preferable to keep the fetching of the urls, size, and name info independent of the service worker. That way, the same service worker can be used in the future to download other collections of files sent to it, not just full datasets based on the resource number. What do you think?

ahamelers:
I made the change. In the end it was easier to add the links to the page, submit the form, and immediately remove them again than mess around with the javascript FormData, which can seemingly only be sent over javascript requests and not by normal form submission. Getting all the links at once does ultimately make the download much faster!

sfisher:
I began looking at this.  Thanks for the updates, Audrey.  I was getting a network error message on some stuff that worked before.  IDK why.  Maybe we can delay until we have a chance to do more testing before rolling this in.

ahamelers:
> I began looking at this. Thanks for the updates, Audrey. I was getting a network error message on some stuff that worked before. IDK why. Maybe we can delay until we have a chance to do more testing before rolling this in.

Is this on dev (or local dev)? Can you provide some links or IDs that aren't working, for me to test?

ahamelers:
I found some that were giving the network error and fixed the issue. It looks like I was clearing the urls etc. from the page too quickly! Works in all instances I can find now.

sfisher:
Thanks, Audrey.  This works great.  I was running on my localhost (local_dev) which was probably slower and tried with http://localhost:3000/stash/dataset/doi:10.7959/dryad.34tmpg50 (which had 200 files) and also on another.  It seems you found the problem since these work great now.

Sorry, I should've given you better information and I didn't realize that some worked and some didn't since both of the things I tried at first failed.

PR #1421, opened 2023-11-14
Allows specifying the exact filename for presigned urls

sfisher:
In case s3 doesn't set it correctly from what the bucket has or if the browser wants to mangle it somehow.

Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/969 for the new downloading outside of Merritt.

See http://localhost:3000/stash/dataset/doi:10.7959/dryad.0gb5mkm0 (dryad-dev.cdlib.org) which has an example filename that liked to get mangled by S3 or the browser.


ryscher:
Thanks for cleaning up this old issue!

PR #1426, opened 2023-11-20
Modify tinyMCE editor functionality to include less garbage HTML

ahamelers:
Removes and restricts menu buttons according to the ticket, and strips almost all style (and attributes) out of pasted text.

Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2916

Also add save-on-change functionality (so exiting the editor is no longer required to save, as mentioned in curator meeting)


ryscher:
Thanks, the paste is much better now. The re-ordering of buttons makes sense, since most-used features are now at the front and less likely to get buried under the "more" button.

PR #1431, opened 2023-11-29
Create environment-specific credentials files

ahamelers:
Remove stage & production credentials from main credentials file and create env specific credentials

Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2941

A couple of things to note with this solution:
- All credentials files must list all environments to work with our initializers (they can just copy the defaults if not needed for the actual credentialed environment, but they must be listed)
- Whenever the defaults are changed or added to, this must be done in all 3 credentials files
- Add `-e env_name` to the end of credentials commands to show/edit the new files (for example `rails credentials:show -e stage`)


ahamelers:
You can find the locations for the new keys in .gitignore

Let me know how you'd like me to deliver the actual keys 😄

ryscher:
Thanks, this is a nice clean solution!

PR #1436, opened 2023-12-02
Dataset zip download using AWS Lambda

ahamelers:
Closes (?) https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2939

I have created an AWS lambda function that creates and streams a zip file to the client, and edited the version download in our API to use it. There are several remaining things to be done.

- For streamed lambda responses, "There is an initial maximum response size of 20 MB, which is a soft limit you can increase." To meet the target of up to 10GB downloads, the size of downloads (streamed response size) allowed by the lambda will need to be increased, which looks like it will require contacting Amazon ("The payload size for streamed responses can be increased from default values. Contact AWS Support to inquire further.")
- When we're satisfied with the functionality, we can also set up the browser zip download to use this for those users without serviceworkers enabled in their browser (JS limitations).

### DIFF
--- a/lib/stash/download/zip_version_presigned.rb
+++ b/lib/stash/download/zip_version_presigned.rb
@@ -13,8 +13,17 @@ module Stash
 
       # passing the controller context allows us to do actions the controller would normally do such as redirecting
       # or rendering within the rails context
-      def initialize(controller_context:)
+      def initialize(controller_context:, resource:)
         @cc = controller_context
+        @resource = resource
+        return if @resource.blank?
+
+        @tenant = @resource&.tenant
+        @version = @resource&.stash_version
+      end
+
+      def valid_resource?
+        !(@resource.blank? || @tenant.blank? || @version.blank?)
       end
 
       def download(resource:)
--- a/lib/stash/download/zip_version_presigned.rb
+++ b/lib/stash/download/zip_version_presigned.rb
@@ -26,24 +26,35 @@ module Stash
         !(@resource.blank? || @tenant.blank? || @version.blank?)
       end
 
+      def generate_token
+        token = @resource.download_token
+        token.token = SecureRandom.uuid if token.token.nil?
+        token.available = Time.now.utc + (1 * 60)
+        token.save
+        token.token
+      end
+
       def download(resource:)
-        if resource&.total_file_size&. < APP_CONFIG.maximums.zip_size
+        @resource ||= resource
+        if @resource&.total_file_size&. < APP_CONFIG.maximums.zip_size
           credentials = ::Aws::Credentials.new(APP_CONFIG[:s3][:key], APP_CONFIG[:s3][:secret])
           signer = ::Aws::Sigv4::Signer.new(service: 'lambda', region: APP_CONFIG[:s3][:region], credentials_provider: credentials)
 
-          time = resource.publication_date.present? && resource.publication_date < Time.now.utc ? resource.publication_date : resource.updated_at
+          time = @resource.publication_date.present? && @resource.publication_date < Time.now.utc ? @resource.publication_date : @resource.updated_at
 
           zip_name = "#{"doi_#{resource.identifier_value}__v#{time.strftime('%Y%m%d')}".gsub(
             %r{\.|:|/}, '_'
           )}.zip"
 
           h = Rails.application.routes.url_helpers
-          download_url = h.zip_assembly_version_url(resource.id).gsub('http://localhost:3000', 'https://dryad-dev.cdlib.org').gsub(/^http:/, 'https:')
+          download_url = h.version_zip_assembly_url(@resource.id).gsub('http://localhost:3000', 'https://dryad-dev.cdlib.org').gsub(/^http:/, 'https:')
+
+          p generate_token
 
           zip_url = signer.presign_url(
             http_method: 'GET',
             expires_in: 3600,
-            url: "https://w27i4vubmmwkssdf6xpszafmhq0rffog.lambda-url.us-west-2.on.aws/?filename=#{zip_name}&download_url=#{CGI.escape(download_url)}"
+            url: "https://w27i4vubmmwkssdf6xpszafmhq0rffog.lambda-url.us-west-2.on.aws/?filename=#{zip_name}&download_url=#{CGI.escape("#{download_url}/#{generate_token}")}"
           )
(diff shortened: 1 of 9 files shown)


lib/stash/download/zip_version_presigned.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      def valid_resource?
+        !(@resource.blank? || @tenant.blank? || @version.blank?)
+      end
+
+      def generate_token
+        token = @resource.download_token
+        token.token = SecureRandom.uuid if token.token.nil?
+        token.available = Time.now.utc + (1 * 60)
+        token.save
```
ryscher, line 33: Where does this save? I could be completely missing something, but it looks like `token` is a local variable that should not respond to `save`. Do you mean something like this?

`@resource.update(download_token: token)`

lib/stash/download/zip_version_presigned.rb
```diff
@@ -0,0 +1,65 @@
...
+          signer = ::Aws::Sigv4::Signer.new(service: 'lambda', region: APP_CONFIG[:s3][:region], credentials_provider: credentials)
+
+          time = @resource.publication_date.present? && @resource.publication_date < Time.now.utc ? @resource.publication_date : @resource.updated_at
+
+          zip_name = "#{"doi_#{resource.identifier_value}__v#{time.strftime('%Y%m%d')}".gsub(
+            %r{\.|:|/}, '_'
+          )}.zip"
+
+          h = Rails.application.routes.url_helpers
+          download_url = h.version_zip_assembly_url(@resource.id).gsub('http://localhost:3000', 'https://dryad-dev.cdlib.org').gsub(/^http:/, 'https:')
```
ryscher, line 50: Hmmmm. I hate the assumption that `localhost == dryad-dev`, but I understand the necessity. We really need to fix https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2267

lib/stash/download/zip_version_presigned.rb
```diff
@@ -0,0 +1,65 @@
...
+        token = @resource.download_token
+        token.token = SecureRandom.uuid if token.token.nil?
+        token.available = Time.now.utc + (1 * 60)
+        token.save
+        token.token
+      end
+
+      def download(resource:)
+        @resource ||= resource
+        if @resource&.total_file_size&. < APP_CONFIG.maximums.zip_size
```
ryscher, line 39: Is the extra `&.` a typo? Or is it actually looking for a method to be produced from the code below? I'm confused.

lib/stash/download/zip_version_presigned.rb
```diff
@@ -0,0 +1,65 @@
...
+            %r{\.|:|/}, '_'
+          )}.zip"
+
+          h = Rails.application.routes.url_helpers
+          download_url = h.version_zip_assembly_url(@resource.id).gsub('http://localhost:3000', 'https://dryad-dev.cdlib.org').gsub(/^http:/, 'https:')
+
+          zip_url = signer.presign_url(
+            http_method: 'GET',
+            expires_in: 3600,
+            url: "https://w27i4vubmmwkssdf6xpszafmhq0rffog.lambda-url.us-west-2.on.aws/?filename=#{zip_name}&download_url=#{CGI.escape("#{download_url}/#{generate_token}")}"
```
ryscher, line 55: Crazy domain name should be in a config file. Isn't this already defined somewhere for the other lambdas?

lib/stash/download/zip_version_presigned.rb
```diff
@@ -0,0 +1,65 @@
...
+        @cc = controller_context
+        @resource = resource
+        return if @resource.blank?
+
+        @tenant = @resource&.tenant
+        @version = @resource&.stash_version
+      end
+
+      def valid_resource?
+        !(@resource.blank? || @tenant.blank? || @version.blank?)
```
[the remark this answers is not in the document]
ahamelers: I copied this from the existing version_presigned when doing the rspec tests. I've got no problem with changing it.
https://github.com/CDL-Dryad/dryad-app/blob/ac0ef0a8a7d34be5200d778b11b14b0d42e24235/lib/stash/download/version_presigned.rb#L28C15-L28C15

lib/stash/download/zip_version_presigned.rb
ahamelers: I copied this from here, and tested it thoroughly, primarily to make sure it would fail in the later check if the available time has elapsed. It definitely saves to the database. I never invent anything, Ryan! Great artists steal, etc.

https://github.com/CDL-Dryad/dryad-app/blob/ac0ef0a8a7d34be5200d778b11b14b0d42e24235/lib/stash/download/version_presigned.rb#L84

lib/stash/download/zip_version_presigned.rb
ahamelers: Totally a typo, thanks!

lib/stash/download/zip_version_presigned.rb
ahamelers: This is a specific URL for the specific lambda function, none of our other lambdas are invoked via URL. I will move this to the config. Thanks! https://docs.aws.amazon.com/lambda/latest/dg/urls-invocation.html

lib/stash/download/zip_version_presigned.rb
ryscher: Wow, that is an unexpected way for Ruby to parse the construction. I guess it's something to do with the `<` being secretly a method call in the lower levels. So it's fine.

lib/stash/download/zip_version_presigned.rb
ryscher: Agreed. I'm not expecting this to be fixed now. More just a note for both of us that the `localhost` assumption is problematic and needs to be fixed soon.

lib/stash/download/zip_version_presigned.rb
ryscher: Oh, got it. I thought the token was just a string, but it's a whole object. My bad.

ryscher:
This is fully tested and working well, now that Amazon has increased our limit to 200MB for Lambda Response Streaming. 

In practice, downloads through Lambda Response Streaming take about 3 times as long as downloading the files individually  from S3. So although we may increase the limit whenever Amazon allows it to be higher, we don't want to steer API users to use this option for larger files.

PR #1438, opened 2023-12-04
Stop error messages for invalid presigned URLs

sfisher:
Invalid because they are DataFiles which we don't give RAT urls for Zenodo for.

Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2729


ahamelers (approved):
Thanks Scott!

PR #1439, opened 2023-12-06
Capistrano optiional shared files

sfisher:
Will copy over files from the shared directory, but only the ones in the list that exist.

i had something from online but couldn't get it work right.  Sorry about all the commits, but I kept changing stuff and testing the deploy on our dev server until it worked. :-/


ryscher:
Thanks, Scott! This is a more comprehensive solution than I was looking at, and it's nice to remove the dependence on the SSM gem. Since the only way I know to really test this is to deploy on a server, I can deploy it to stage....

ryscher:
Yay, tests passed, and it deploys correctly on stage. Thanks!!!!

PR #1440, opened 2023-12-07
Sentence case more things

ryscher:
These items fall in a gray area between proper names and "regular" headings, but Jen asked us to err more on the side of sentence case, so I think these all fall in line with what she wanted.


app/views/layouts/_our_governance.html.md
```diff
@@ -2,7 +2,7 @@
 
 <p>Our <a href="https://blog.datadryad.org/code-of-conduct/">Code of Conduct</a> applies for all Dryad community spaces.</p>
```
ahamelers, line 3: How about 'Code of Conduct' ?

ryscher: fixed

PR #1444, opened 2023-12-12
Adding script for comparison of our COUNTER logs vs DataCite tracking

sfisher:
This script was just to analyze our log events vs the ones that DataCite gets.

Please see the readme.MD here for full information.  I can share the files I used non-publicly (because of privacy) and also the SQLite database it creates.

I was able to get a best guess match rate of  over 90% for DataCite events being matched in our logs.  Matching is pretty speculative.  We have a lot more events than the DataCite tracker in general.

I didn't do any in-depth analysis, but just looked over the database which shows our logs and where it believes matches occur.  I looked over the first 1,000-2,000 rows.  The pattern I see is that there are a whole lot of bots (as seen by the user agent) crawling our site constantly.  The items that are matched generally look like real browsers and users and the other traffic not as much.

We have had a very static approach to bot and machine identification with an initial list we got from the Counter project early on from a github repository and it was rarely updated (or it was moved on to someplace else without us being aware).

It seems like this difference is a major reason that the items are so different and the differences make some sense.

The tracker doesn't work with non-JS agents (such as many bots and also the API).  For tracking API usage, we got a suggestion to log and combine somehow.  But I think an simpler solution may be to make a request from our server to the tracker at DataCite to log these events with an HTTP request if that is allowed. Or perhaps we can get those events and batch them to DataCite at some frequency instead.

I'm not sure what the stance on tracking API usage is for COUNTER, really, but their tracker doesn't do it out of the box.

### DIFF
--- a/script/counter_log_comparison/Gemfile
+++ b/script/counter_log_comparison/Gemfile
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+source "https://rubygems.org"
+
+gem "activerecord", "~> 6.0.0"
+gem "sqlite3", "~> 1.4.0"
+gem "byebug"
--- a/script/counter_log_comparison/Gemfile.lock
+++ b/script/counter_log_comparison/Gemfile.lock
@@ -1,0 +1,35 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
+    activemodel (6.0.6.1)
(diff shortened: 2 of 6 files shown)


ryscher:
Thanks for archiving this script!

PR #1445, opened 2023-12-12
Clarify token documentation

henrypinkard:
@ryscher thanks for your help in getting me set up with API! I think this change to the docs would help others not be confused in the same way as I was in the future


ryscher:
Thanks Henry, you're correct.

PR #1446, opened 2023-12-15
Do not unlink zenodo if it's been published before

sfisher:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3014

Doesn't unlink if it's been published before since it seems like republishing with only metadata changes doesn't always set the view tag for the files.

This makes it work when it should and it still prevents people who haven't published from being worried that their unpublished items don't work on Zenodo yet.


ryscher:
Thanks, Scott. Glad to see the fix for this problem was relatively simple.

PR #1447, opened 2023-12-18
Add a rake task which generates a CSV for datasets w/ possible articles associated

DragosIorgulescu:
As per spec -- see notes in the description

Todo:
- [x] fix specs

ref https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2124

### DIFF
--- a/lib/tasks/stash_engine_tasks.rake
+++ b/lib/tasks/stash_engine_tasks.rake
@@ -432,7 +432,10 @@ namespace :identifiers do
         .where.not(
           id: StashEngine::Resource
             .joins(:related_identifiers)
-            .where({ related_identifier_type: 'doi', work_type: 'primary_article' })
+            .where({
+                     "#{StashDatacite::RelatedIdentifier.table_name}.related_identifier_type": 'doi',
+                     "#{StashDatacite::RelatedIdentifier.table_name}.work_type": 'primary_article'
+                   })
             .pluck(:identifier_id)
         ).each do |i|
 
--- a/lib/tasks/stash_engine_tasks.rake
+++ b/lib/tasks/stash_engine_tasks.rake
@@ -420,6 +420,24 @@ namespace :identifiers do
     end
   end
 
+  desc 'Generate a report of datasets with possible articles'
+  task datasets_with_possible_articles_report: :environment do
+    FileUtils.mkdir_p(REPORTS_DIR)
+    outfile = File.join(REPORTS_DIR, 'datasets_with_possible_articles.csv')
+    p "Writing #{outfile}..."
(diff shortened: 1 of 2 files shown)


lib/tasks/stash_engine_tasks.rake
```diff
@@ -420,6 +420,24 @@ namespace :identifiers do
...
+    FileUtils.mkdir_p(REPORTS_DIR)
+    outfile = File.join(REPORTS_DIR, 'datasets_with_possible_articles.csv')
+    p "Writing #{outfile}..."
+    CSV.open(outfile, 'w') do |csv|
+      csv << %w[ID Identifier ISSN]
+      StashEngine::Identifier.publicly_viewable.joins(:internal_data)
+        .where(internal_data: { data_type: 'publicationISSN' })
+        .where.not(
+          id: StashEngine::InternalDatum.where({ data_type: 'publicationDOI' }).pluck(:identifier_id)
+        ).each do |i|
```
DragosIorgulescu, line 434: I am not entirely sure this query is correct. The ticket describes that identifiers without a `publicationDOI` internal data record associated should be queried, however it does not look like `publicationDOI` is an accepted type on the `InternalDatum` model validations:
https://github.com/CDL-Dryad/dryad-app/blob/d567ffd22ba09245c457ed468aa23eee5ebcc269/app/models/stash_engine/internal_datum.rb#L5-L9

And looking at a similar rake task, it seems that the publication DOI is located elsewhere?
https://github.com/CDL-Dryad/dryad-app/blob/bbe7a36a8b444fef6c28aa8f85adfcde4c0cdb8b/app/models/stash_engine/identifier.rb#L318-L326

Curious to know your thoughts on this.

spec/tasks/stash_engine_tasks_spec.rb
```diff
@@ -0,0 +1,28 @@
+describe 'identifiers:datasets_with_possible_articles_report', type: :task do
+  let(:report_path) { File.join(REPORTS_DIR, 'datasets_with_possible_articles.csv') }
+
+  before do
+    # create(:identifier, internal_data: [build(:internal_datum, data_type: 'publicationISSN')])
+    # create(:identifier, internal_data: [build(:internal_datum, data_type: 'publicationDOI')])
```
DragosIorgulescu, line 6: These 2 fail to be created due to validation errors, however I will be waiting to resolve the previous comment on the existence of `publicationDOI` type before fixing the test setup here.

ryscher:
Thanks, this report looks good. It is returning more datasets than I expected, so we will need to discuss next steps with the curation team.

PR #1448, opened 2023-12-18
feat: set hold for peer review on the resource if review is not allowed

DragosIorgulescu:
Looking into this, here are my current findings/questions:
- it seems that there is a check already preventing the radio buttons from being displayed unless the dataset is [ready for review](https://github.com/CDL-Dryad/dryad-app/blob/a02744b44f5e6bce4c82248fd6570311da7170cf/app/views/stash_datacite/peer_review/_review.html.erb#L1)
- the `allow_review?` method already checks [for `has_accepted_manuscript?`](https://github.com/CDL-Dryad/dryad-app/blob/bf7a774eec7eb1101cecc38dfc357fc21984c3d9/app/models/stash_engine/identifier.rb#L280), which in turn already checks that the manuscript [is not rejected](https://github.com/CDL-Dryad/dryad-app/blob/bf7a774eec7eb1101cecc38dfc357fc21984c3d9/app/models/stash_engine/identifier.rb#L314)
- I went ahead and set `hold_for_peer_review` in the controller action, but the condition might change based on the questions above

ref [GREI 1.3.6 - Automatically PPR datasets#2996](https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2996)


app/controllers/stash_datacite/resources_controller.rb
```diff
@@ -29,6 +29,7 @@ def review
           check_required_fields(@resource)
           @review = Resource::Review.new(@resource)
           @resource.has_geolocation = @review.geolocation_data?
+          @resource.hold_for_peer_review = true if !@resource.identifier.has_rejected_manuscript? && !@resource.identifier.has_accepted_manuscript?
```
DragosIorgulescu, line 32: The ticket requirements specifies these 2 conditions, but shouldn't we also check that this dataset is related to a manuscript and not a `published article` / `other`?

DragosIorgulescu: Add a check here making sure that the dataset is not published as well

app/models/stash_engine/identifier.rb
```diff
@@ -306,7 +310,7 @@ def allow_review?
       # do not allow to go into peer review after already published
       return false if pub_state == 'published'
       # do not allow to go into peer review if associated manuscript already accepted or published
-      return false if has_accepted_manuscript? || publication_article_doi
+      return false if has_accepted_manuscript? || has_rejected_manuscript? || publication_article_doi
```
ahamelers, line 313: I don't think we want to prevent people from doing PPR if they have a rejected manuscript since they often resubmit to other journals in such cases

DragosIorgulescu:
closing this as it will be handled in the related PR

PR #1450, opened 2023-12-18
Add checksum generation method and bimonthly validation

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2792

Checksum utility modified from https://github.com/CDLUC3/mrt-core2/blob/main/core/src/main/java/org/cdlib/mrt/utility/Checksums.java

### DIFF
--- a/app/mailers/stash_engine/user_mailer.rb
+++ b/app/mailers/stash_engine/user_mailer.rb
@@ -97,6 +97,16 @@ module StashEngine
            subject: "#{rails_env}General error \"#{@resource.title}\" (doi:#{@resource.identifier_value})")
     end
 
+    def file_validation_error(file, error_text)
+      logger.warn("Unable to report update error #{error}; nil file") unless file.present?
+      @zenodo_error_emails = APP_CONFIG['zenodo_error_email']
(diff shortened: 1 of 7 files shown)


ryscher:
When I run this on a test file, I get:
```
Error: Error getting checksums: Error processing input: Unsupported input type
/home/ec2-user/dryad-app/lib/stash/checksums.rb:49:in `rescue in get_checksums'
/home/ec2-user/dryad-app/lib/stash/checksums.rb:44:in `get_checksums'
```
And the `validated_at` date is set, despite the validation failing.

ryscher:
Oh, it looks like the "example" block in checksums.rb is running when it is initialized, so it expects the example file to exist, but I don't have that file on my system.

ryscher:
A few thoughts:
- Validation failures are important enough that we should receive email, in addition to any error in the log
- We should do load testing on production before adding this to the monthly cron, in case we want to move this function to a dedicated server.
- We will want to stagger the initial validated_at dates... it would be useful to have a process for running the existing files in batches.

ryscher:
Updates look good, thanks for the quick fixes!

PR #1451, opened 2023-12-18
Adding DataCite Counter credentials

sfisher:
See https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2793

We obtained these credentials from DataCite.  We probably don't want to start using them until the January reporting for COUNTER, but I put them into the encrypted files as entries that are commented out so they are stored someplace more permanent and are easy to change out when the time comes.

I also disabled automatic running of reports in the cron-related files.  In January, I plan to run the rest of the December reporting manually and then we can switch out the credentials after that.

The Dryad credentials will still be able to get stats but they cannot update the reports CDL submitted (and they were clear that we shouldn't submit the same data under both accounts). If they need to be changed, CDL can re-submit old reports or else DataCite said they could intervene if there was a problem.

I'm still waiting for news from DataCite about when/if they can transition our use of the web tracker into real usage rather than test usage. That might or might not correspond with the first of the year, but we will stop log processing and reporting from logs when the web tracker goes into effect and is considered real data by them.


ryscher:
Converted this to draft status. I will need to manually merge the changes into the credential files.

ryscher:
Closing in favor of #1459.

PR #1460, opened 2024-01-03
Instructions about dealing with zenodo errors

sfisher:
I updated detailed instruction at the top of the page and added a couple of recent problem cases since their transition.


ryscher:
Thanks for the updates!

PR #1465, opened 2024-01-09
Fix loading order of javascripts, remove unused files

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3030

In future Dryad could be cleaned up to no longer use jquery.


ryscher:
Great, thanks!!!!

PR #1467, opened 2024-01-10
fix: don't hijack the default rails logger

DragosIorgulescu:
- causes non-api logs to sometimes show up in the api logs

ref https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3042


ryscher:
Thanks, this works well in my testing.

PR #1468, opened 2024-01-10
Add checks before resetting peer review hold for default_to_ppr journals

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3015

### DIFF
--- a/app/controllers/stash_datacite/publications_controller.rb
+++ b/app/controllers/stash_datacite/publications_controller.rb
@@ -77,8 +77,10 @@ module StashDatacite
         @msid = manage_internal_datum(identifier: @se_id, data_type: 'manuscriptNumber', value: parsed_msid)
       end
 
-      # if the newly-set journal wants PPR by default, set the PPR value for this resource
-      @resource.update(hold_for_peer_review: @se_id.journal&.default_to_ppr)
+      if @resource.identifier.allow_review? && @resource.previous_curated_resource.nil? && @resource.curation_start_date.nil?
+        # if the newly-set journal wants PPR by default, set the PPR value for this resource
+        @resource.update(hold_for_peer_review: @se_id.journal&.default_to_ppr)
+      end
 
       save_doi
     end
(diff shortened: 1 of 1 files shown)


app/controllers/stash_datacite/publications_controller.rb
```diff
@@ -77,8 +77,10 @@ def save_form_to_internal_data
         @msid = manage_internal_datum(identifier: @se_id, data_type: 'manuscriptNumber', value: parsed_msid)
       end
 
-      # if the newly-set journal wants PPR by default, set the PPR value for this resource
-      @resource.update(hold_for_peer_review: @se_id.journal&.default_to_ppr)
+      if @resource.identifier.allow_review? && @resource.previous_curated_resource.nil? && @resource.curation_start_date.nil?
```
DragosIorgulescu, line 80: do we know these can only be `nil`? Maybe `blank?` would be a safer option here
```suggestion
      if @resource.identifier.allow_review? && @resource.previous_curated_resource.blank? && @resource.curation_start_date.blank?
```

app/controllers/stash_datacite/publications_controller.rb
```diff
@@ -77,8 +77,10 @@ def save_form_to_internal_data
...
+        # if the newly-set journal wants PPR by default, set the PPR value for this resource
+        @resource.update(hold_for_peer_review: @se_id.journal&.default_to_ppr)
```
DragosIorgulescu, line 82: What do we do in case of validation errors here? Do these DB changes need to run in a transaction? 

I guess this is a more general question across the whole project, I've seen multiple controller actions that make multiple DB changes in a non-atomic way, just wondering if this is intentional?

app/controllers/stash_datacite/publications_controller.rb
ahamelers: This is the same check done on the screen for the user (seen in the ticket), where it tests correctly. It seems safest for these to match.

https://github.com/CDL-Dryad/dryad-app/blob/3ad044f3eea08b14214d77a3410fccdf20d7165e/app/views/stash_datacite/peer_review/_review.html.erb#L1

DragosIorgulescu (approved):

app/controllers/stash_datacite/publications_controller.rb
ryscher: Will discuss in next dev standup

PR #1469, opened 2024-01-10
chore: update `publicationISSN` usage to be load via `Identifier#publication_issn` where possible

DragosIorgulescu:
This is part of the preparation to move `publicationISSN` out of `InternalData` and into `RelatedIdentifier`

ref https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3019


ryscher:
publicationISSN appears in many other places throughout the code -- we should discuss this in more detail.

PR #1480, opened 2024-01-16
chore: add annotations to all models & fixtures

DragosIorgulescu:
These are automagically updated on every `db:migrate` command (run in dev -- `local`), so no need to touch it once set up.


ryscher:
Thanks,  this looks great!

PR #1481, opened 2024-01-17
feat: allow passing `holdForPeerReview` on `POST /api/v2/datasets`

DragosIorgulescu:
- updates the `Resource#hold_for_peer_review` based on this value
- update the associated curation activity created to be of `peer_review` status

ref https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2945

### DIFF
--- a/app/models/stash_api/dataset_parser.rb
+++ b/app/models/stash_api/dataset_parser.rb
@@ -27,7 +27,7 @@ module StashApi
       user_note = "Created by API user, assigned ownership to #{owning_user&.name} (#{owning_user_id})"
       hold_for_peer_review = @hash['holdForPeerReview']
       @resource.curation_activities << StashEngine::CurationActivity.create(
-        status: hold_for_peer_review ? 'peer_review' : @resource.current_curation_status || 'in_progress',
+        status: @resource.current_curation_status || 'in_progress',
         user_id: @user.id,
         resource_id: @resource.id,
         note: user_note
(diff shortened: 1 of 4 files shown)


app/models/stash_api/dataset_parser.rb
```diff
@@ -25,10 +25,13 @@ def parse
       owning_user_id = establish_owning_user_id
       owning_user = StashEngine::User.find(owning_user_id)
       user_note = "Created by API user, assigned ownership to #{owning_user&.name} (#{owning_user_id})"
-      @resource.curation_activities << StashEngine::CurationActivity.create(status: @resource.current_curation_status || 'in_progress',
-                                                                            user_id: @user.id,
-                                                                            resource_id: @resource.id,
-                                                                            note: user_note)
+      hold_for_peer_review = @hash['holdForPeerReview']
+      @resource.curation_activities << StashEngine::CurationActivity.create(
+        status: hold_for_peer_review ? 'peer_review' : @resource.current_curation_status || 'in_progress',
```
DragosIorgulescu, line 30: because the `status` is defined as a `string_enum` on `StashEngine::CurationActivity`, we cannot use `StashEngine::CurationActivity.statuses[:peer_review]` to check the validity of the string. 

I would propose we create a ticket to go through all enums in the system and clean this up a bit:
- make sure any column that could work as an enum in the db is treated as such in ruby
- check existing enums and see if we can improve anything (like the one mentioned here)

ryscher, line 30: After reviewing the behavior, we need to remove the changes above for `CurationActivity`. The change below in `@resource.update` is sufficient.

ryscher:
Please also add this field to the documentation:
- `documentation/apis/submission.md`, section "Retaining a dataset in a private status for peer review purposes"
- `documentation/apis/sample_dataset.json`

ryscher:
Great, thanks for updating this! It takes time to learn all of the little quirks of the system.

PR #1482, opened 2024-01-17
Load single CEDAR template in specific context, and save CEDAR json file

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2866


ryscher:
This is looking good. A few notes:
- Based on recent emails, I think we could go live after one more update of the CEE code.
- Do we want to move the config for the CEDAR templates into the default section of the config file? There are currently separate entries for the dev and stage environments.
- Should we add comments to the config to help us remember that configured templates will not display unless the UI code is changed?

ryscher:
Looks good. We're waiting on confirmation about the latest neuroscience template. Is this ready to be enabled for production? It looks like the production config is currently set to disallow loading.

ryscher:
Great! Let's see if people start to use it.

PR #1485, opened 2024-01-18
Retry zip download & improve error messages

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3044


ahamelers:
Talked to Scott about the original reason for the problem-causing session check; will put in additional checks for permissions for unpublished resources

ryscher:
Looks good. Hope that takes care of the problems.

PR #1486, opened 2024-01-19
Add a storage_version_id to stash_engine_generic_files

ryscher:
As described in https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2940, this provides a place to indicate the storage location of a file, so the file can be easily located, regardless of which version/resource the file was originally created for.

Initially, we can leave the old "walk through the versions" code in place, and use it whenever `storage_version_id` is null, until we have completely transitioned to the new process.


ahamelers (commented):
Thanks Ryan!

PR #1487, opened 2024-01-23
Add new tenant config and logo

ahamelers:
For https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3056


ryscher:
Merging, as this fixes the initial issue for #3056. But we will also need to address the health system as described in #2998

PR #1490, opened 2024-01-24
Removing scott from most email notifications within Dryad

sfisher:
I went through and removed myself from most email notifications within Dryad.  There were quite a few places, including some environments that I don't think are used anymore.  Where there was no one else's email, I substituted ryan instead for now.

I left myself on the production submission errors since it will still be useful for me to see if things are failing to go through Merritt in production and I can check in with them as needed.

This is sort of sad, 😞 but it will be nice to have a cleaner inbox. 

mailbox


ahamelers (approved):
Thanks Scott—we'll miss you!

PR #1493, opened 2024-01-31
Improve neuro keywords

ryscher:
An attempt to improve the keywords that trigger the neuroscience CEDAR template. `cortex` can describe many non-neuro structures, including plants. It's also useful to include words like `cerebral`, `cerebellum`, and `cognitive`.


ahamelers:
Thanks Ryan!

PR #1496, opened 2024-01-31
Fix rspec tests requiring abstract

ahamelers:
I believe that the abstract is being entered in the tinyMCE editor but not saved due to the new need for an API key. Rather than store a valid API key in our example config file, we can create the abstract directly and ignore the editor (something already being done for several other page elements) which allows these tests to pass.


ryscher:
Thanks!!!!

PR #1498, opened 2024-01-31
Non-Merritt storage: phase 1

ryscher:
This PR is the first phase of the shift to storing content without using Merritt, performing the needed tasks directly in S3.

This feature still has some known problems, but due to restructuring of the old gem layout, it is getting quite large and complex. Therefore, I want to get more eyes on it and hopefully get it back in sync with the main branch. I think this will help prevent merge conflicts and make it easier to keep the rspecs working.

If you think it's not ready to merge yet, that's fine. Let me know which issues are the most critical to fix before we merge. I will keep working in this branch until it is merged, working through the known problems in the order they are listed below.

Features included:
- Major restructuring of the storage code: remove the `stash-merritt` gem and `stash-deposit` gem. The relevant classes have migrated to `Stash::Repo`
- Simplification of the abstraction hierarchy that we never really used, merging classes and removing unneeded methods
- Newly uploaded files are stored in a directory/key structure that starts with `v3`, instead of the `ark` hierarchy that Merritt used. Note that many datasets will contain a mix of files that are in the `ark` hierarchy and the `v3` hierarcy; the retrieval code has been updated to check in both places as needed.
- Removed reliance on the `merritt-manifest` package that creates Merritt-specific files (though I expect to add back a simpler manifest to aid with understanding the files)
- Although this PR still contains the "builder" classes that create summaries for internal Merritt processing, they are not currently used.
- Simplified the process of moving files around. Previously, files were uploaded to the temporary S3 bucket, then copied to the Merritt ingest server for procesing, and copied to the permanent S3 bucket. Now, they are copied directly from the temporary bucket to the permanent bucket using the S3 API, which is much faster.

Known problems:
- There is some issue with converting a dataset to `published` status, even though the file downloads work correctly.
- Files that are "uploaded" via a URL are not handled yet
- Files over 5GB are not supported by the S3 copy method -- they need to be switched to a streaming copy
- When a filename contains a 4-byte Unicode character (🙃), the individual file does not download properly.


ryscher:
@ahamelers I fixed the issue with being able to publish a dataset. The DataCite XML building process had been dependent on some of the other Merritt builder classes that are no longer used. I move the relevant pieces directly into `DataciteGen`

ahamelers:
Thanks Ryan! I've made separate tickets for the remaining issues.

PR #1506, opened 2024-02-07
Change annotate gem settings

ahamelers:


ryscher:
Thanks, this will be much less annoying.

PR #1508, opened 2024-02-07
Update config instructions for v3 servers

ryscher:
- Adds some steps to avoid a rogue version of Ruby running on the server, which can cause the dreaded `ffi` error. 
- Adds instructions for installing and configuring Apache, so Rails can respond to port 80 while allowing the Shibboleth library to run as well.


ahamelers:
Thanks, Ryan! And thanks for your detective work on this issue

PR #1509, opened 2024-02-08
Permanent storage for files over 5GB

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3127

According to [the docs](https://www.rubydoc.info/gems/aws-sdk-resources/2.9.15/Aws/S3/Object), multipart copy "can provide performance improvements on large objects", so we may want to do it for more than just files over 5GB. We could start somewhere in the middle (1GB?), or go all the way to the other extreme, and do it for everything over 5MB

>  *:multipart_copy* (Boolean) — default: false — When ‘true`, the object will be copied using the multipart APIs. This is necessary for objects larger than 5GB and can provide performance improvements on large objects. Amazon S3 does not accept multipart copies for objects smaller than 5MB.


ryscher:
Great job. This works well.

ahamelers:
@ryscher What do you think about lowering the size threshold for when the multipart upload kicks in, to take advantage of potential "perfomance improvements" Amazon mentions? Any thoughts on what it should be? See commentary in the ticket. I can put in another PR adjusting the value and moving it to the config.

ryscher:
In my limited testing, the time to copy between S3 buckets is miniscule when compared to the time it takes to upload the file into the "staging" bucket, so the performance improvements won't have much impact. But we can certainly lower the threshhold. 1GB seems like a natural place to start. (Weird that S3 doesn't just make its own decisions about this...)

PR #1515, opened 2024-02-12
Correct file size calculations for new storage

ryscher:
After running some tests, I saw that file size calculations were improperly including files that had been deleted from the version. This ignores them.


ahamelers:
Good catch!

PR #1518, opened 2024-02-13
Fix persistent loading icon error

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3148


ryscher:
Thanks for the quick cleanup!

PR #1519, opened 2024-02-15
Fix README loading and saving settings

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3156
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3170


ryscher:
I believe this is checking for a 10% change from the previously saved version. This does not work for the case where a submitted version of the README was acceptable or near-accecptable, and the user is making one more version to add finishing touches. 

With the current approach of disabling saves, we also have the problem where a user can start typing, switch to another page of the submission system to check on something, and when they switch back, their edits have disappeared.

I suggest:
- Always save changes that are made in the editor
- Move the similarity check into the validation process
- Perform the similarity check against the original template, instead of against the last saved version
- Increase the requirement for % of change (or, in the current code, reduce the similarity score). If we're comparing against the original template, I would expect the changes of a "real" edit to be more than 50%, perhaps 75%.

ahamelers:
Thanks @ryscher. The issue I was trying to account for is that the editor itself is making changes to the loaded template content, which the system is counting as a change, and then saving. I will go back to the drawing board.

ahamelers:
@ryscher I've fixed the bug of the first loading of the template being saved as README content, which I think was added in the fix for https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2975. No additional checks for now.

ryscher:
Thanks, this looks good.

PR #1520, opened 2024-02-15
Add more neuroscience keywords

ryscher:
Followup to #1493 -- it was surprising that the CEDAR template didn't appear when editing [Zefan's dataset](https://doi.org/10.5061/dryad.b2rbnzsh9), so these are keywords that would have triggered it.


ahamelers:
Thanks Ryan!

ahamelers:
I'm not sure about 'memory', I think there are a lot of non-human-neuroscience fields that could crop up in, but we can try it for now!

PR #1526, opened 2024-02-21
Update ROR data download script

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2893

Might it be worthwhile to have different FILE_DIR variables for different servers and store them in the config?


ryscher:
Thanks for the updates!

PR #1527, opened 2024-02-21
Fix typo that could prevent submissions from processing

ryscher:
We never noticed that this typo was causing exceptions, because there is an exception handler in the method that treats the exception as an external error.


ahamelers:
Thank you, Ryan

PR #1540, opened 2024-03-06
Fix checksum memory issue and other bugs

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2430
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3208
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3181

The checksum issue is a very simple fix (it was getting the object from AWS that was overloading the memory and not working intermittently, switching to using a URL fixed the issue) but took me forever to test because of a simple error I made in testing. Sorry about that!

### DIFF
--- a/lib/stash/repo/submission_job.rb
+++ b/lib/stash/repo/submission_job.rb
@@ -103,7 +103,7 @@ module Stash
             logger.info("    #{data_file.url} ==> #{permanent_bucket}/#{permanent_key}")
             s3_perm.put_stream(s3_key: permanent_key, stream: f)
           end
-          @digest_input = data_file.url
+          digest_input = data_file.url
         else
           logger.info("    #{staged_bucket}/#{staged_key} ==> #{permanent_bucket}/#{permanent_key}")
           s3.copy(from_bucket_name: staged_bucket, from_s3_key: staged_key,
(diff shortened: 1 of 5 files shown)


lib/stash/repo/submission_job.rb
```diff
@@ -114,7 +114,7 @@ def copy_to_permanent_store(data_file)
         update = { storage_version_id: resource.id }
         if data_file.digest.nil?
           digest_type = 'sha-256'
```
ahamelers, line 116: This is the change that fixes the checksum generation, if you want to make this change on the live server, @ryscher

ryscher:
I made this change on the production server, BUT I don't think the current configuration works for files that are uploaded via URL. The current checksum statement relies on `staged_key`, using the default "staged" bucket. When files are uploaded via URL, they don't exist in this location. I think we can swap this out for the permanent location, but I didn't verify it.

lib/stash/repo/submission_job.rb
```diff
@@ -103,7 +103,7 @@ def copy_to_permanent_store(data_file)
             logger.info("    #{data_file.url} ==> #{permanent_bucket}/#{permanent_key}")
             s3_perm.put_stream(s3_key: permanent_key, stream: f)
           end
-          @digest_input = data_file.url
```
ahamelers, line 106: @ryscher The input for URL uploads is set here, and should not be an issue. However there is another issue which was causing checksum duplication—`digest_input` should be a local variable in both instances 🤦

ryscher:
Great, thanks for tracking this down!

PR #1541, opened 2024-03-07
Increase efficiency for URL uploads and checksums

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3212

I got the down gem to work to process URLs for checksums, and then to load files via external URL to S3, without overloading memory or using disk space. I then restructured the implementation so that the S3 copy and S3 URL upload files are handled more separately, and the checksum generation and upload to S3 are handled in a single stream of the URL uploads.

### DIFF
--- a/lib/stash/aws/s3.rb
+++ b/lib/stash/aws/s3.rb
@@ -75,6 +75,12 @@ module Stash
         s3_bucket.objects(prefix: "#{s3_key}/").batch_delete!
       end
 
+      def object(s3_key:)
+        return unless s3_key
+
+        s3_bucket.object(s3_key)
+      end
+
       def objects(starts_with:)
         return unless starts_with
 
--- a/lib/stash/checksums.rb
+++ b/lib/stash/checksums.rb
@@ -1,6 +1,6 @@
 require 'digest'
 require 'logger'
-require 'open-uri'
+require 'down'
 
 module Stash
   class Checksums
@@ -88,8 +88,14 @@ module Stash
     private
 
     def process_url(url)
-      URI.parse(url).open do |f|
-        process_stream(f)
+      Down.open(url).each_chunk do |chunk|
+        @input_size += chunk.length
+        @digest_list.each { |digest| digest.algorithm.update(chunk) }
+      end
+
+      @digest_list.each do |digest|
+        finish_digest(digest)
+        digest.input_size = @input_size
       end
     end
 
--- a/lib/stash/repo/submission_job.rb
+++ b/lib/stash/repo/submission_job.rb
@@ -1,4 +1,5 @@
 require 'rails'
+require 'down'
 require 'active_record'
 require 'concurrent/promise'
 require 'byebug'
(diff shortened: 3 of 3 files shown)


ryscher:
When testing on the old dev server, I recevied this error:
```
/apps/dryad/apps/ui/shared/bundle/ruby/3.0.0/gems/aws-sdk-s3-1.134.0/lib/aws-sdk-s3/multipart_stream_uploader.rb:105:in `abort_upload&#39;: multipart upload failed: Part n\
umber must be an integer between 1 and 10000, inclusive; Part number must be an integer between 1 and 10000, inclusive; Part number must be an integer between 1 and 10000,\
 inclusive; Part number must be an integer between 1 and 10000, inclusive; Broken pipe (Aws::S3::MultipartUploadError)^M
        from /apps/dryad/apps/ui/shared/bundle/ruby/3.0.0/gems/aws-sdk-s3-1.134.0/lib/aws-sdk-s3/multipart_stream_uploader.rb:94:in `upload_parts&#39;^M
        from /apps/dryad/apps/ui/shared/bundle/ruby/3.0.0/gems/aws-sdk-s3-1.134.0/lib/aws-sdk-s3/multipart_stream_uploader.rb:50:in `block in upload&#39;^M
        from /apps/dryad/apps/ui/shared/bundle/ruby/3.0.0/gems/aws-sdk-core-3.181.0/lib/aws-sdk-core/plugins/user_agent.rb:28:in `feature&#39;^M
        from /apps/dryad/apps/ui/shared/bundle/ruby/3.0.0/gems/aws-sdk-s3-1.134.0/lib/aws-sdk-s3/multipart_stream_uploader.rb:48:in `upload&#39;^M
        from /apps/dryad/apps/ui/shared/bundle/ruby/3.0.0/gems/aws-sdk-s3-1.134.0/lib/aws-sdk-s3/customizations/object.rb:394:in `block in upload_stream&#39;^M
        from /apps/dryad/apps/ui/shared/bundle/ruby/3.0.0/gems/aws-sdk-core-3.181.0/lib/aws-sdk-core/plugins/user_agent.rb:28:in `feature&#39;^M
        from /apps/dryad/apps/ui/shared/bundle/ruby/3.0.0/gems/aws-sdk-s3-1.134.0/lib/aws-sdk-s3/customizations/object.rb:393:in `upload_stream&#39;^M
        from /apps/dryad/apps/ui/releases/20240307202419/lib/stash/repo/submission_job.rb:137:in `copy_external_to_permanent_store&#39;^M
```

ryscher:
The S3 multipart upload only allows 10,000 parts, but each part can be up to 5GB. The default chunk size of Down (at least for my sample file) is 16384, which limits us to files of 163MB. I'm testing a way to increase the chunk size...

ahamelers:
Ah sorry, it worked with the huge example URL in my testing, even without explicitly setting the chunk size. With that added does it now work for you @ryscher ?

ryscher:
Not quite there yet. I found that even if you're chunking the downloads, the `upload_stream` does separate chunking, which can still use up the 10,000 limit. So I had to set a chunk size on the upload as well. But with a very large file (100GB), the process isn't finishing... I suspect there is some time of timeout. So we may need to still support writing a temp file to disk. Or limit the file size?

ryscher:
Ok, I did more testing from a variety of sources and it looks like the current code works about as well as possible. The remaining problems are caused by sites that do not deliver the files well. The 100GB file I was testing with is on Dropbox, and that file is erratic even when I try to download it with a browser or curl.

So I think this is done. I will let @ahamelers look over it before merging, but I may try it out on the prod servers in the meantime.

PR #1542, opened 2024-03-07
Add checksum cleanup tasks

ahamelers:
Once checksum generation is working correctly (#1541), and whenever we find a historic API digest that was incorrect, we can clean up bad checksums in the database and propagate newly added checksums to 'copied' files with these tasks.

I think this can close https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3155 and https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3031

### DIFF
--- a/lib/tasks/file_validate.rake
+++ b/lib/tasks/file_validate.rake
@@ -22,4 +22,50 @@ namespace :checksums do
 
     end
   end
+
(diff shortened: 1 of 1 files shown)


DragosIorgulescu (approved):

ryscher:
Looks good, and should be helpful.

PR #1544, opened 2024-03-12
Add a standardised set of API error classes and handling at the controller level

DragosIorgulescu:
Closes: https://github.com/datadryad/dryad-product-roadmap/issues/3237
Closes: https://github.com/datadryad/dryad-product-roadmap/issues/3668

Since we already use an simpler error message response in multiple places, here is an example of how the error response envelope looks like for API requests:
```
{
    "error": "The userId orcid is not known to Dryad. Please supply a matching orcid in the dataset author list."
}
```
ref https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3012

Initial implementation was (NOT USED)
```json
{
    "errors": [
        {
            "code": "bad_request",
            "message": "The userId orcid is not known to Dryad. Please supply a matching orcid in the dataset author list.",
            "metadata": {}
        }
    ]
}
```


ahamelers (approved):

DragosIorgulescu:
Since this will change how all error responses, we need to hold off on merging until we:
- update API docs
- go through all API endpoints and validate what errors are being generated

In the meantime:
- create a separate PR with just the status code fix

alinvetian:
1. As differences, all errors that were just raising an error will be returning new errors json, but same HTTP code.
    - This should not generate any issue for the API users that only check the status code and not the message
    - But will generate issues for the users that check for raised message error

2. There new error response is not consistent with the other app that handles JSON error response, which looks like
https://github.com/datadryad/dryad-app/blob/main/app/controllers/stash_api/datasets_controller.rb#L39
```
{
    "error": "A dataset with same information already exists."
}
```

PR #1548, opened 2024-03-18
feat: auto-publish (approve) resources based on different criteria

DragosIorgulescu:
Items left TO DO:
- [ ] do some testing in sandbox and check how the changed fields look like when making certain changes to a dataset
- [ ] add spec coverage
- [ ] record activity entry

Questions:
- how do you set a primary article on a dataset? I've only been able to add `article` type relationships, not `primary_article`
- what activity log should we generate for auto-approvals?

### DIFF
--- a/app/controllers/stash_datacite/subjects_controller.rb
+++ b/app/controllers/stash_datacite/subjects_controller.rb
@@ -54,18 +54,11 @@ module StashDatacite
     private
 
     def ensure_subject(subject_str)
-      subject = find_or_create_subject(subject_str)
+      subject = StashDatacite::Subject.find_or_create_subject(subject_str)
       subjects = @resource.subjects
       return if subjects.exists?(subject.id)
 
       subjects << subject
-    end
-
-    def find_or_create_subject(subject)
-      existing = Subject.where('subject LIKE ?', subject).non_fos.first
-      return existing if existing
-
-      Subject.create(subject: subject)
     end
 
     def set_subject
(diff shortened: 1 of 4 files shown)


DragosIorgulescu:
> how do you set a primary article on a dataset? I've only been able to add article type relationships, not primary_article
This is happening when choosing the `a published article` option when adding / editing a dataset

> what activity log should we generate for auto-approvals?
Look at `CurationActivity` 
- set the `user_id` to 0 (so it's treated as a system activity)
- examples:
```
   Status change notification sent to author                                                                                                                                  |
| Status change notification sent to author                                                                                                                                  |
| Private for peer review notification sent to journal                                                                                                                       |
| Status change notification sent to author                                                                                                                                  |
| Status change notification sent to author                                                                                                                                  |
| received notification from journal module that the associated manuscript is submitted, but the dataset is in_progress, so it will retain that status                       |
| System auto-assigned curator Kailasha Purushothaman                                                                                                                        |
| Status change notification sent to author                                                                                                                                  |
| Status change notification sent to author                                                                                                                                  |
| Status change notification sent to author                                                                                                                                  |
| Private for peer review notification sent to journal                                                                                                                       |
| Status change notification sent to author                                                                                                                                  |
| System noticed possible duplicate dataset 10.5061/dryad.280gb5mvw                                                                                                          Status change notification sent to author                 
```

Maybe something like: 
`automatically approved based on ... changes`

ahamelers:
> how do you set a primary article on a dataset? I've only been able to add article type relationships, not primary_article

The `primary_article` is an article linked by filling out a journal & DOI, or Manuscript number, at the top of the page. We'll probably revise this eventually as it does cause confusion (it's not just you!). It also causes curators too much effort when a primary article must be linked after the dataset is published.

ahamelers:
> what activity log should we generate for auto-approvals?

I think `Automatically published minimal revision` is succinct and descriptive for a `user_id: 0` activity message

PR #1553, opened 2024-03-20
fix: escape certain strings when passing queries to solr

DragosIorgulescu:
These are pretty edge cases, but I did test the string that was presented in the ticket and in the following format it works correctly: https://datadryad.org/search?q=Global+Positioning+System+%5C%28GPS%5C%29


ahamelers (approved):

ryscher:
These changes are fine, but do not address the requirements in the ticket. I'll merge this and put details in the ticket itself.

PR #1554, opened 2024-03-20
fix: correct automatic peer review settings

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3005
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3015
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2899

When the review page is accessed, if the submitter is unable to set PPR (because a previous version of the submission has already been curated, published, etc.) the system will also save that false PPR setting in the database, so it will no longer go into PPR unexpectedly.


ahamelers:
Just had a look and this is definitely going to merge conflict with #1448, @DragosIorgulescu 🫣

PR #1555, opened 2024-03-20
feat: set datasets associated w/ manuscripts under review to PPR automatically

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2996

@DragosIorgulescu @ryscher I stole this out from under you because my PR #1554, which this is following on, will create loads of merge conflicts with #1448 .

### DIFF
--- a/app/controllers/stash_datacite/resources_controller.rb
+++ b/app/controllers/stash_datacite/resources_controller.rb
@@ -29,7 +29,10 @@ module StashDatacite
           check_required_fields(@resource)
           @review = Resource::Review.new(@resource)
           @resource.has_geolocation = @review.geolocation_data?
-          unless @resource.identifier.allow_review? && @resource.previous_curated_resource.blank? && @resource.curation_start_date.blank?
+          if @resource.identifier.manuscript_number? && !@resource.identifier.has_accepted_manuscript? &&
+            !@resource.identifier.has_rejected_manuscript
+            @resource.hold_for_peer_review = true
+          elsif !@resource.identifier.allow_review? || @resource.previous_curated_resource.present? || @resource.curation_start_date.present?
             @resource.hold_for_peer_review = false
             @resource.peer_review_end_date = nil
           end
--- a/app/controllers/stash_datacite/resources_controller.rb
+++ b/app/controllers/stash_datacite/resources_controller.rb
@@ -29,6 +29,10 @@ module StashDatacite
           check_required_fields(@resource)
           @review = Resource::Review.new(@resource)
           @resource.has_geolocation = @review.geolocation_data?
+          unless @resource.identifier.allow_review? && @resource.previous_curated_resource.blank? && @resource.curation_start_date.blank?
+            @resource.hold_for_peer_review = false
+            @resource.peer_review_end_date = nil
+          end
           @resource.save!
         end
       end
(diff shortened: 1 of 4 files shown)


app/controllers/stash_datacite/resources_controller.rb
```diff
@@ -29,6 +29,13 @@ def review
           check_required_fields(@resource)
           @review = Resource::Review.new(@resource)
           @resource.has_geolocation = @review.geolocation_data?
+          if @resource.identifier.manuscript_number.present? &&
+            !@resource.identifier.has_accepted_manuscript? && !@resource.identifier.has_rejected_manuscript?
```
DragosIorgulescu, line 33: I'd propose consolidating this into a single method: https://github.com/CDL-Dryad/dryad-app/pull/1448/files#diff-c80483d1bdd4f952bcc678f28761d2eb1eacf1a67d184344072f5e4d13510f82R342-R354

app/controllers/stash_datacite/resources_controller.rb
```diff
@@ -29,6 +29,13 @@ def review
...
+            @resource.hold_for_peer_review = true
+          elsif !@resource.identifier.allow_review? || @resource.previous_curated_resource.present? || @resource.curation_start_date.present?
```
DragosIorgulescu, line 35: Same for these checks, it looks like we have them in 2 places, maybe add a new method on the `Resource` model?

DragosIorgulescu (commented):
Other than pulling the slight refactoring done on the [other PR](https://github.com/CDL-Dryad/dryad-app/pull/1448), this looks good to me 👌

app/views/stash_datacite/peer_review/_review.html.erb
```diff
@@ -1,4 +1,8 @@
-<% if @resource.identifier.allow_review? && @resource.previous_curated_resource.nil? && @resource.curation_start_date.nil? %>
+<% if @resource.identifier.automatic_ppr? %>
+  <div>
+    <p>This dataset will remain private for peer review until we receive a notification from the journal confirming formal acceptance of the associated manuscript.</p>
+  </div>
+<% elsif @resource.identifier.allow_review? && @resource.previous_curated_resource.blank? && @resource.curation_start_date.blank? %>
```
DragosIorgulescu, line 5: just curious if this should not be `!@resource.identifier.allow_review?` instead, as I see the condition in the controller is with a `!` in front and here it's not

ahamelers: No, this is for the opposite—if it's allowed and the submission hasn't already been curated, it allows the user to select whether the submission should be held for peer review. The controller is enforcing the opposite (setting the peer review hold to false if the user is NOT allowed to select the setting). This was clearer before when it was an "unless" in the controller 😄 https://github.com/CDL-Dryad/dryad-app/pull/1554/files#diff-734af58a5dee7c319f046e6d357e3fd166c3d9e992d5bd893660142b7fcdc2ad

DragosIorgulescu: gotcha, thank you for the clarification 🙏  this is a bit hard to read, not in terms of code, but in terms of what the expected business logic is. We should probably take a look at this at a higher level at some point

DragosIorgulescu (approved):

ryscher:
Ok, this looks fine. I hope it does everything the curators want.

PR #1559, opened 2024-03-22
feat: add option for curators to hide the landing page README display

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2874

### DIFF
--- a/app/views/stash_datacite/descriptions/_readme.html.erb
+++ b/app/views/stash_datacite/descriptions/_readme.html.erb
@@ -1,11 +1,11 @@
 <% if !review.readme_content.blank? %>
 <% content = review.readme_content %>
-<% # if content.match(/^\\#/) %>
-  <% # reformat = markdown_render content %>
-  <% # content = ActionView::Base.full_sanitizer.sanitize(reformat) %>
-<% # end %>
-<div id="landing_readme">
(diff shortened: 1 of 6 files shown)


app/views/stash_datacite/descriptions/_readme.html.erb
```diff
@@ -1,18 +1,19 @@
...
+<div id="landing_readme" class="<%='expanded' if expand_readme %>">
+    <% unless expand_readme %><button id="expand-readme" class="o-button__plain-text4" aria-expanded="false" aria-controls="landing_readme">Expand README <i class="fa fa-caret-down" aria-hidden="true"></i></button><% end %>
     <div class="t-landing__text-wall <%= 'highlight' if highlight_fields.include?('technical_info') %>">
     <% if !content.match(/^# /) %>
       <h2>README</h2>
     <% end %>
     <%= markdown_render content.sub(/^# /, '# README: ') %>
     </div>
 </div>
+<% unless expand_readme %>
```
ryscher, line 16: I think this line should be above the </div>, so the divs are closed properly

ryscher (commented):
Everything else looks great.

app/views/stash_datacite/descriptions/_readme.html.erb
ahamelers: The div is closed properly, as it is always present. The unless and if checks here control a class on the div, and the presence of both the expand/collapse button and the javascript that makes it work.

PR #1563, opened 2024-03-25
feat: improve ROR search results with exact alias & acronym matches

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3004
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3244


ryscher:
Thanks, this seems fine. I think the next time we adjust it, we will need to create a suite of examples to test, since we're at the point where fixing one example breaks another.

PR #1564, opened 2024-03-26
fix: load/process large sql queries in batches to improve memory use

ahamelers:
`find_each` and `find_in_batches` load and process ActiveRecord results in batches, reducing memory load. The default is 1000 items per batch.


DragosIorgulescu (approved):
Yep, this was part of where I was going as well, great work 👌

PR #1566, opened 2024-03-26
feat: convert email handling from sendmail to Amazon SES

ryscher:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/2809 and possibly others...

Handling email on the local machines was cumbersome, and we frequently received reports that emails were caught by spam filters. This switches the email handling to go through Amazon SES, which should trigger spam filters less, and will often deliver messages faster.


ahamelers:
Thanks Ryan!

PR #1575, opened 2024-03-28
fix: delayed job logs should rotate

DragosIorgulescu:


ryscher:
Thanks for finding this.

PR #1577, opened 2024-03-28
fix: update hard coded server paths

DragosIorgulescu:
TODO:
- these should be dynamically set


ryscher:
Thanks for fixing this!

PR #1579, opened 2024-03-29
fix: batch & adjust queries to optimize curation stats calculation

ahamelers:
Also condensed a previously used series of methods for easier readability


ryscher:
Thanks for the improvements!

PR #1582, opened 2024-04-01
feat: add rake task generating a report for the last month for GREI

DragosIorgulescu:
- adds a template in `/public` based on which the report will be generated
- generates a PDF report which is stored in the `/reports` dir
- todo: optimize stats calculation
- todo: add average stats metrics

---

Example of a generated report: [GREI_monthly_report_2024_04_01.pdf](https://github.com/CDL-Dryad/dryad-app/files/14823565/GREI_monthly_report_2024_04_01.pdf)

Items / questions left:
- [x] clean up the PDF UI
- [ ] update how stats are calculated
- [ ] Is this going to be a report generated manually or do we need to add it to some cron job?

### DIFF
--- a/config/initializers/inflections.rb
+++ b/config/initializers/inflections.rb
@@ -21,4 +21,5 @@ end
 ActiveSupport::Inflector.inflections(:en) do |inflect|
   inflect.acronym "NIH"
   inflect.acronym "XML"
+  inflect.acronym "GREI"
 end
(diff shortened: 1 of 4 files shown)


ahamelers (approved):

ryscher:
This is a great start. I'm going to merge it, even though there is some cleanup work to do, and the PDF will be adjusted, because it's a single report that doesn't affect anything else, so it can live in the codebase while it is being adjusted.

PR #1594, opened 2024-04-04
feat: migrate tenant configuration details into the database.

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3274

This is a major reworking, and (mainly due to rspec tests) touches a great number of files. I've categorized the work into commits to make it easier to parse—I recommend reading through the code changes commit-by-commit instead of all at once.

I will do a separate PR at a later date (probably after the new table is set up correctly and working on staging) to update our documentation and delete the `config/tenants` directory, which is no longer needed.


ryscher:
I'm still testing, but it looks good so far. One minor note (and this is actually a bug in the current prod system) in the `tenants.rake`, we want the environment checks to use `v3_production` instead of `production`.

ahamelers:
> I'm still testing, but it looks good so far. One minor note (and this is actually a bug in the current prod system) in the `tenants.rake`, we want the environment checks to use `v3_production` instead of `production`.

That's why the checks are `includes('production')`—should cover both!

ryscher:
> That's why the checks are `include('production')`—should cover both!

And that's why I should stop reading code when I know by brain has shut off for the night. Thanks!

PR #1597, opened 2024-04-05
fix: bugs stopping us from making DOIs live

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3278


DragosIorgulescu:
Ideally we should not have so many production env checks within the codebase... we should probably log these in a separate ticket and explore if we can avoid them

PR #1598, opened 2024-04-05
feat: add date range task for sending DOI metadata

ahamelers:
Closes https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3265


DragosIorgulescu:
do we want to add this to a cron as well?

ahamelers:
> do we want to add this to a cron as well?

No, I've added a new ticket for revamping the publication process into a queue with automatic retries etc. as the submission queue is, but it's lower priority (unless we can't get this DOI issue fixed!) https://github.com/CDL-Dryad/dryad-product-roadmap/issues/3285

PR #1619, opened 2024-04-24
Prevent nil pointer error in datasets controller

ryscher:
Prevent some nil pointer errors I saw in the logs


ryscher:
Yes, I saw them in the logs when I was testing another PR, after #1617 had already been merged.

PR #1621, opened 2024-04-24
chore: update tenant documentation

ahamelers:
And remove unused files


ryscher:
Thanks, this looks very thorough.

PR #1622, opened 2024-04-24
fix: github workflows

ahamelers:
Workaround from https://github.com/orgs/community/discussions/120966#discussioncomment-9211925


ahamelers:
Looks like the workflow issue workaround is no longer needed 🎉

PR #1627, opened 2024-04-29
chore: remove stash wrapper engine

DragosIorgulescu:
- no longer being used -- would be useful to have more context regarding where these files were migrated, or what they were used for

Related to the ruby & rails upgrade


ahamelers:
Close in favor of #1664

PR #1636, opened 2024-05-01
feat: store tenant logos in the database and modify in the UI

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3340

I wrote a test for this but it's mysteriously failing on github only. I think this may be because the current extra bits of the migration, which add all the initial logo files to the db so no one has to do it manually, are too complex and must be run each time the test db starts up. I have xited the test for now and will try to get it working again along with removing a lot of complications and extraneous stuff in a next-sprint cleanup PR.


ahamelers:
Sorry about all the force pushing, I had a very mysterious rspec error only on github

ryscher:
Looks good. I ran into a minor JS issue, but that might just be my server, so I'm going to to merge this while I do further testing.

PR #1646, opened 2024-05-09
fix: change proxy user to only exclude system roles

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3358

This should be deployed as soon as possible to fix the errors experienced by API users


ryscher:
Looks good to me.

PR #1647, opened 2024-05-10
fix: additional checks of readme editor files

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3359
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3345

### DIFF
--- a/app/javascript/react/components/MarkdownEditor/milkdown_editor.css
+++ b/app/javascript/react/components/MarkdownEditor/milkdown_editor.css
@@ -312,6 +312,7 @@
 
 .ProseMirror p {
   white-space: normal;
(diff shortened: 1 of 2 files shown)


ryscher:
This PR is fine, though I'm not certain it solve the full problem. I can rename a PDF file with a .txt extension, and I'm allowed to import it, though the results are less than stellar.

ahamelers:
It's hard to solve any "problem" of requiring or verifying specific mimetypes for anything entirely in the front end like this. We are trying to tell people how to do things corectly, but if there are people who really want their READMEs to look awful it's very difficult to stop them!

PR #1654, opened 2024-05-14
feat: aws js captcha for zip download

ahamelers:
Implement AWS challenge & captcha checking and tokens for the javascript zip download function.

Closes https://github.com/datadryad/dryad-product-roadmap/issues/3361


ryscher:
Let's discuss this. I want to understand it a bit more.

I wasn't able to make the CAPTCHA appear when testing, and this change breaks downloads in a private/incognito browser.

ahamelers:
> I wasn't able to make the CAPTCHA appear when testing

This is actually because of changes related to this! The CAPTCHA no longer appears, because the included [AWS browser 'challenge' test](https://docs.aws.amazon.com/waf/latest/developerguide/waf-captcha-and-challenge.html) is successful for normal browser users.

> This change breaks downloads in a private/incognito browser

Service workers (which is what generates our zip downloads) do not work in firefox in incognitio mode, and never have.

PR #1657, opened 2024-05-21
feat: add links from admin page to stripe invoice page

ryscher:
Ok, I'm not certain whether this is a good idea. This updates rendering in the "Payment info" box, so Stripe invoice IDs are linked to the relevant invoice in Stripe. 

Only a handful of Dryad staff actually have permissions within Stripe, so the link won't work for a lot of the people who see it. Should we wrap this link in some permissions logic?


ahamelers:
As discussed, I think this is fine and doesn't require permissions logic.

PR #1663, opened 2024-05-28
feat: upgrade Node & webpacker (shakapacker)

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3314
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3152

v20.13.1 of Node will need to be installed on the servers & used in the .bashrc file to deploy


ahamelers:
This is deployed on dev. Once it's merged, I will begin the Ruby and Rails upgrade as well.

ryscher:
Looks good, and works fine for me. One note, I see the following warning:

 " > shakapacker@6.6.0" has incorrect peer dependency "webpack-dev-server@^4.9.0".

Should we delete this dependency?

ahamelers:
> Looks good, and works fine for me. One note, I see the following warning:
> 
> " > shakapacker@6.6.0" has incorrect peer dependency "webpack-dev-server@^4.9.0".
> 
> Should we delete this dependency?

This is because shakapacker is asking for webpack-dev-server v4.9.0, but we're actually using webpack-dev-server version 5(something). After testing showed no problems I decided to keep using the newer version against shakapacker's wishes.

PR #1689, opened 2024-06-06
Only create ORCID invites for authors with emails

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3374

May also improve https://github.com/datadryad/dryad-product-roadmap/issues/2595 —it looks like this was only sending emails if there was more than 1 author needing an invitation previously! Can't explain this.


ryscher:
Oh, wow. This was some old logic, written before it was possible to reorder the authors. So it was assuming that the first author always had an ORCID. Thanks for catching it!

PR #1690, opened 2024-06-06
Improve design of author & affiliation lists

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3362

Also took this opportunity to update our fontawesome version to use the pretty ORCID icon in their new set


app/views/stash_datacite/authors/_show.html.erb
```diff
@@ -1,48 +1,58 @@
 <% unless authors.nil? %>
+  <% highlights = highlight_fields.detect {|h| h[:index] == i} %>
```
ryscher, line 2: This `i` is undefined? Seems this code was copied from other places in the PR where there was an index.

app/views/layouts/_our_governance.html.md
```diff
@@ -11,7 +11,7 @@
...
+  <h5>Jake Carlson (2023-2026)</h5><p> <a href="https://orcid.org/0000-0003-2733-0969" target="_blank" aria-label="ORCID profile (opens in new window)"><i class="fab fa-orcid" aria-hidden="true"></i></a> is the Associate University Librarian for Research, Collections and Outreach at the University at Buffalo (UB) Libraries. His responsibilities include increasing the capabilities of the UB Libraries to respond to researchers’ needs with managing, sharing and preserving their research data. Dryad will be a critical resource in the Libraries’ research data services program. Previously, he has been the Director of Deep Blue Repository and Research Data Services at the University of Michigan and a Data Services Specialist at Purdue University. Carlson has authored or co-authored more than 30 articles on research data services in libraries. His work has been supported by funding from the National Science Foundation (NSF), the Alfred P. Sloan Foundation, and the Institute for Museum and Library Services (IMLS).</p>
```
ryscher, line 14: The linebreak makes this icon a little confusing. In many screen sizes, it looks like the icon is the start of the sentence, not the "Jake Carlson". What if we put the icon inside the heading tags?

ahamelers: I think I was considering more how the text read for screen readers than how it looked. Will see if I can make it work for both.

app/views/stash_datacite/authors/_show.html.erb
```diff
@@ -1,48 +1,58 @@
...
-					end
-				}.reject(&:blank?).join(", ").html_safe  %>
-    </div>
-    <% deleted = highlight_fields.detect {|h| h.key?(:deleted) } %>
-    <% if deleted %>
-    	<div class="highlight"><%= deleted[:deleted] %> authors deleted</div>
-    <% end %>
+  	<b>Affiliations</b>
+  	<ol>
+  	<% affs.map do |aff| %>
```
ryscher, line 53: If all authors have non-ROR affiliations, `affs` is null, which causes an error here.

ahamelers: I can't recreate this error! I've put in a check for the variable around this section just in case, but no combination of ror and non-ror affiliations (even 0 affiliations) causes an error for me.

ryscher: Well, it's working now, so I'm assuming the changes helped.

PR #1691, opened 2024-06-06
fix: no longer download automatically from share links

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3411


ahamelers:
I've deployed this branch on production

PR #1694, opened 2024-06-07
Allow "Create Salesforce case" to properly redirect users to Salesforce

ryscher:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3413

(Already deployed on prod and appears to be working)


ahamelers:
If you want to keep the full word 'salesforce' in the button, we could alternatively change the button class to one that doesn't have a preset maximum width

PR #1695, opened 2024-06-07
Change API logs from rolling daily to a limited size

ryscher:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3414

Make these files roll at a set size, similar to the other logs. This probably isn't the best rolling method, since it doesn't guarantee us any particular number of days of history. The default Ruby logging doesn't support a number of days by default; it would be something we need to custom code or set up with an external tool.


alinvetian (approved):
Tested locally and it works as expected.

PR #1711, opened 2024-06-18
Increase size limit of files to use with Frictionless to 50MB

ryscher:
Based on testing in https://github.com/datadryad/dryad-product-roadmap/issues/2880, we are updating the size of files to use with the Frictionless testing. New limits are 50MB for production and 1MB for test servers.

### DIFF
--- a/config/app_config.yml
+++ b/config/app_config.yml
@@ -87,7 +87,7 @@ test:
     api_zip_size: 200
     files: 100
   frictionless:
-    size_limit: 6291456
+    size_limit: 1_000_000
   rate_limit:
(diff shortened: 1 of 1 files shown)


ahamelers:
@ryscher Looking at the data, do you not think we could go up to 100MB?

ahamelers:
> I'm worried about whether submitters will actually wait for 4 minutes or more. But I'll defer to your judgement.

If we allow it to run on even larger files, submitters can go ahead while it's still loading, but then eventually the curators could still see the errors. We could merge this for now and double check whether the curators think that is a good idea

PR #1715, opened 2024-06-18
Prefer gem over node package load for Fontawesome

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3442


ryscher:
Thanks!!!!! This worked.

PR #1719, opened 2024-06-19
Improve speed of SQL queries & admin dashboard load

ahamelers:
Through careful changing of how and when (and what kind of) joins are loaded, I've been able to increase the speed of queries for this page dramatically (on dev).

In addition to the changes directly to the page, I've created migrations & changes for new tables which should expand our ability to use associations instead of complicated query calculations or methods generally—the dates table should make it much easier to calculate curation stats, though I am waiting until the reporting and stats decisions are made to make those changes.

### DIFF
--- a/app/models/stash_engine/resource.rb
+++ b/app/models/stash_engine/resource.rb
@@ -499,13 +499,13 @@ module StashEngine
 
     # Date on which the user first submitted this resource
     def submitted_date
-      curation_activities.order(:id).where("status = 'submitted' OR status = 'peer_review'")&.first&.created_at
+      process_date.submitted || process_date.peer_review
     end
 
     # Date on which the curators first received this resource
     # (for peer_review datasets, the date at which it came out of peer_review)
     def curation_start_date
-      curation_activities.order(:id).where("status = 'submitted' OR status = 'curation'")&.first&.created_at
+      process_date.curation_start
     end
 
     # ------------------------------------------------------------
--- a/app/models/stash_engine/resource.rb
+++ b/app/models/stash_engine/resource.rb
@@ -50,7 +50,7 @@ module StashEngine
     self.table_name = 'stash_engine_resources'
     # ------------------------------------------------------------
     # Relations
-
+    has_one :process_date, as: :processable, dependent: :destroy
     has_many :authors, class_name: 'StashEngine::Author', dependent: :destroy
     has_many :generic_files, class_name: 'StashEngine::GenericFile', dependent: :destroy
     has_many :data_files, class_name: 'StashEngine::DataFile', dependent: :destroy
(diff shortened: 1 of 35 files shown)


app/models/stash_engine/resource.rb
```diff
@@ -497,13 +499,13 @@ def init_curation_status
 
     # Date on which the user first submitted this resource
     def submitted_date
-      curation_activities.order(:id).where("status = 'submitted' OR status = 'peer_review'")&.first&.created_at
+      process_date.submitted || process_date.peer_review
```
ryscher, line 502: Is this changing the meaning of the date? It might make sense to reverse the order, since peer_review dates will typically be before submitted dates.

app/views/stash_engine/admin_dashboard/_table_row.html.erb
```diff
@@ -72,7 +72,7 @@
       <td class="c-lined-table__digits"><%= formatted_datetime(dataset.last_curation_activity.updated_at) %></span></td>
     <% end %>
     <% if @fields.include?('submit_date') %>
-      <td class="c-lined-table__digits"><%= formatted_datetime(dataset.submit_date) %></td>
+      <td class="c-lined-table__digits"><%= formatted_datetime(dataset.process_date.submitted || dataset.process_date.peer_review) %></td>
```
ryscher, line 75: Again, it might make sense to switch the order here, so any peer_review date is preferred. (Depending on how the submitted date is calculated)

app/views/stash_engine/admin_dashboard/index.csv.erb
```diff
@@ -39,29 +39,29 @@ head += ',Published' if @fields.include?('publication_date')
...
     dpc = ''
     dpc = dataset.tenant&.short_name if dataset.identifier.payment_id = dataset.tenant_id
-    dpc = dataset.journal_title if dataset&.issn&.include?(dataset.identifier.payment_id)
+    dpc = dataset.identifier&.journal&.journal_title if dataset.identifier&.journal&.issn_array&.include?(dataset.identifier.payment_id)
     dpc = dataset.identifier.payment_id.split("funder:").last.split("|").first if dataset.identifier.payment_id.starts_with?('funder')
     row << dpc
   end
   row << dataset.last_curation_activity.updated_at if @fields.include?('updated_at') 
-  row << dataset.submit_date if @fields.include?('submit_date')
+  row << (dataset.process_date.submitted || dataset.process_date.peer_review) if @fields.include?('submit_date')
```
ryscher, line 64: Here too

app/models/stash_engine/resource.rb
ahamelers: I guess this is changing it a bit. On the admin dashboard the submission date helps curators decide what order to do things in based on how long they have been waiting in the queue, in which case using an earlier peer review date over the submitted date makes it seem like things have been there for months and something has gone wrong. I think it makes sense to prefer the date the status actually goes to 'submitted'. Should we discuss more?

ryscher: Can you check with Molly? She has the best understanding of who is using each type of date.

ryscher:
This is ready to merge after the above comments are resolved.

PR #1720, opened 2024-06-20
updated Github raw URLs regex and original_url handling

alinvetian:
closes https://github.com/datadryad/dryad-product-roadmap/issues/3260

- [ ] Any URL upload errors ("Crawling of HTML files isn't supported.") that do not allow URL uploads in the front end should also apply to the API.
- The URL translator should work correctly & match these github file URLs.
  - [ ] The raw version of the URL should be in the url column, while the original should be moved to original_url
  - [ ] The correct file size, matching the file on github should be recorded in the database


ryscher:
Thanks Alin. This seems to be working as intended.

PR #1721, opened 2024-06-20
Temporariliy set dev DB to the duplicate of production

ryscher:


ahamelers:
@ryscher do we still want to merge this?

ryscher:
Only if you need it for testing

PR #1737, opened 2024-06-27
Speed up load and display of the My datasets page

ahamelers:
The My datasets page uses a lot of slow methods and models unnecessarily and loads a lot of secondary associations for each result in individual SQL commands. This speeds everything up by switching to already loaded models for some checks, removing several instances of circular model loading (loading an association which then reloads the original model), only loading what is needed in the original sql query, and preloading associations to reduce sql queries overall.

### DIFF
--- a/app/views/stash_engine/dashboard/_user_datasets.html.erb
+++ b/app/views/stash_engine/dashboard/_user_datasets.html.erb
@@ -1,10 +1,10 @@
 <% @datasets.each_with_index do |dataset, i| %>
   <% resource = dataset&.latest_resource %>
-  <% display_r = @display_resources[i]%>
+  <% title = resource.title.blank? ? '[No title supplied]' : resource.title %>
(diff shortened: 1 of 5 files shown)


app/views/stash_engine/dashboard/_user_datasets.html.erb
```diff
@@ -1,10 +1,10 @@
 <% @datasets.each_with_index do |dataset, i| %>
   <% resource = dataset&.latest_resource %>
-  <% display_r = @display_resources[i]%>
+  <% title = resource.title.blank? ? '[No title supplied]' : resource.title %>
```
alinvetian, line 3: You can write this as 
```
title = title.presence || '[No title supplied]'
```

ahamelers: @alinvetian there's probably a couple hundred places in our code where that could be changed...

PR #1757, opened 2024-07-08
fixed rspec exiting from rake tasks specs

alinvetian:
Rspec did not run all specs, but random numbers of specs due to `exit` commands in rake tasks.


ahamelers:
Thanks Alin!

PR #1827, opened 2024-09-02
Remove pub date from new copied resources

ahamelers:
I was trying to see if any other resources were affected by the issue where published resources are set to file_view=false (https://github.com/datadryad/dryad-product-roadmap/issues/3186, https://github.com/datadryad/dryad-product-roadmap/issues/3206)

I'll have to figure out a more complex query for checking this, because doing this query there were tons more than I was expecting (around 8800):

`StashEngine::Resource.where('publication_date < now()').where(file_view: false)`

It turns out when a resource is copied to make a new version from one that has been published, it still has the previous publication_date set, even though the new version has never been published. This removes that!

### DIFF
--- a/app/models/stash_engine/resource.rb
+++ b/app/models/stash_engine/resource.rb
@@ -127,6 +127,7 @@ module StashEngine
                   # let init_version do its job
                   new_resource.current_resource_state_id = nil
                   # do not mark these resources for public view until they've been re-curated and embargoed/published again
+                  new_resource.publication_date = nil
                   new_resource.meta_view = false
                   new_resource.file_view = false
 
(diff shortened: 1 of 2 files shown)


ahamelers:
It turns out there's a test checking that this is copied! Can anyone explain why we would want this copied? @ryscher 

It seems counterintuitive

ahamelers:
I tried to investigate the reasoning behind the copying of the pub date to new versions. The test was copied from another source, and I can't seem to track that source down to check the description of the PR or tickets.

In any case, the copying of the is now definitely counterintuitive to our setup, since each resource and the identifier itself now all have publication date columns.

PR #1833, opened 2024-09-10
Fix admin dashboard author list limit

ahamelers:
The admin dashboard author column is listing all authors when it is meant to limit to the first 6 in that view.

### DIFF
--- a/app/controllers/stash_engine/admin_dashboard_controller.rb
+++ b/app/controllers/stash_engine/admin_dashboard_controller.rb
@@ -177,9 +177,9 @@ module StashEngine
 
     def author_field
       @datasets = @datasets.select("(
-        select GROUP_CONCAT(CONCAT_WS(', ', sea.author_last_name, sea.author_first_name) ORDER BY sea.author_order, sea.id separator '; ')
-          from (select sa.author_last_name, sa.author_first_name, sa.author_order, sa.id
-          from stash_engine_authors sa where sa.resource_id = stash_engine_resources.id limit 6) sea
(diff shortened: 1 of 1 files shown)


app/controllers/stash_engine/admin_dashboard_controller.rb
```diff
@@ -176,6 +160,29 @@ def add_fields
...
+          select stash_engine_users.* from stash_engine_users
+          inner join stash_engine_roles on stash_engine_users.id = stash_engine_roles.user_id
+            and role in ('curator', 'superuser')
+        ) curator on curator.id = stash_engine_resources.current_editor_id")
+        .select("CONCAT_WS(' ', curator.first_name, curator.last_name) as curator_name")
+    end
+
+    def author_field
+      @datasets = @datasets.select("(
+        select GROUP_CONCAT(CONCAT_WS(', ', sea.author_last_name, sea.author_first_name) separator '; ')
```
ahamelers, line 180: This is the only actual change, all the other methods are just moving things around to make a big method shorter

app/controllers/stash_engine/admin_dashboard_controller.rb
```diff
@@ -176,6 +160,29 @@ def add_fields
...
+          from (select sa.author_last_name, sa.author_first_name
+          from stash_engine_authors sa where sa.resource_id = stash_engine_resources.id ORDER BY sa.author_order, sa.id LIMIT 6) sea
```
ahamelers, line 182: This is the only actual change, all the other bits are just moving things around to make a long method shorter

PR #1836, opened 2024-09-10
Exclude throttling on file preview_check

ryscher:
We were seeing legitimate users blocked due to the speed of requests for the preview_check function. This excludes it from the throttling process.


ryscher:
I manually added this to the current release on production. Not sure whether it is the optimal solution, but it's working well enough until Audrey can give it more thought.

PR #1840, opened 2024-09-11
Fix citation format for works citing

ahamelers:
For https://github.com/datadryad/dryad-product-roadmap/issues/3640

I notice we save the citation string we create, and not the retrieved raw metadata itself, making it hard to regenerate/correct the citation formatting for existing records. I've added a db column and we will save the raw JSON retrieved going into the future to make any future changes to the citation string easy to regenerate.

We will need to build a way to retrieve and refresh everything where the JSON has not previously been saved, once we've agreed on a method for this with DataCite.

### DIFF
--- a/lib/stash/datacite_metadata.rb
+++ b/lib/stash/datacite_metadata.rb
@@ -38,7 +38,7 @@ module Stash
       return '' if names.blank?
 
       names = names.map { |i| name_finder(i) }
-      return "#{names.first} et al." if names.length > 4
+      return "#{names.first(3).join('; ')} et al." if names.length > 4
 
       names.join('; ')
     end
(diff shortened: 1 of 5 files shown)


lib/stash/datacite_metadata.rb
```diff
@@ -84,8 +86,13 @@ def html_citation
...
       # html_safe when concatenated with other stuff makes non-html-safe escaped
-      ''.html_safe + "#{author_names} (#{year_published}), #{title}, #{journal}, #{resource_type}, " +
-          "<a href=\"#{doi_link}\" target=\"_blank\">#{doi_link}</a>".html_safe
+      citation_array = []
+      citation_array << "#{author_names}#{year_published ? " (#{year_published})" : ''}"
+      citation_array << "#{title.html_safe}#{resource_type}"
+      citation_array << journal
+      citation_array << publisher unless raw_metadata['type'].include?('journal')
+      citation_array << "<a href=\"#{doi_link}\" target=\"_blank\">#{doi_link}</a>".html_safe
+      citation_array.reject(&:blank?).join('. ')
```
alinvetian, line 95: Did you intended to replace `,` with `.` s separator?

ahamelers: Yes, absolutely I did, separating everything with a comma isn't done in any citation standard! Thanks for checking though.

PR #1844, opened 2024-09-14
Include README and all-file list in reviewer view

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3631


alinvetian:
@ahamelers what about point `4. Download option (same as dataset landing page); users can either select individual files for download or ALL files for download 5) Incorporate tabular data preview option (?)`

ahamelers:
> what about point `4. Download option (same as dataset landing page); users can either select individual files for download or ALL files for download 5) Incorporate tabular data preview option (?)`

@alinvetian these are done, the dataset display is now the same for the landing page and the reviewer (share) page, including the file download options available and any previews.

PR #1846, opened 2024-09-17
Allow multiple user roles on resources

ahamelers:
For https://github.com/datadryad/dryad-product-roadmap/issues/2711

This is doing the prep work for allowing multiple editors per submission—while changing nothing visible for most users, this changes the back end of how ownership of a submission is indicated and migrates existing data to reflect this. This is in preparation for the future goal of having 3 distinct roles (creator, collaborator, submitter) with access to a submission and all the UI changes involved in applying and indicating those roles.

- Adds resource roles and applies both "creator" and "submitter" user roles for each resource
- Uses these roles throughout the code to check user permissions and display user datasets
- Repurposes the user_id column, which is no longer used for dataset ownership, to indicate the assigned curator (more stickily than using the current_editor_id or having to search through the history of curation activities)
- Only changes the current_editor_id when a dataset is actually edited

What is visible for some users?
- Curators, once assigned to and not unassigned from a dataset, will be recorded much more stickily for the purposes of the admin dashboard and reporting (see https://github.com/datadryad/dryad-product-roadmap/issues/3477)
- Both creators and submitters will see the same datasets listed on the "My datasets" page, if these are 2 separate users (this should affect very few users, mainly API submitters).

Incidentally, this also:
- Deletes some unused methods
- Removes some noisy test log outputs
- Fixes the sorting and filtering of items on the user admin profile table of datasets


ahamelers:
I considered putting the roles on the identifier instead of each individual resource, since 1) users universally want to set or change roles for the entire dataset, not just a version, and 2) it would create vastly fewer database rows. But, we've always assigned ownership on the resource level previously. Let me know your thoughts and if I should change this to the identifier level, @ryscher

ryscher:
This looks good, and works with the scenarios I tested. I do want to test a little more with the ACS workflow once this goes live.

PR #1848, opened 2024-09-19
Related works style fix

ahamelers:
Curators have pointed out the related works on the landing page can break weird on chrome, specifically, putting the icon on its own line. This fixes that and spruces the display up a bit across browsers.

### DIFF
--- a/app/assets/stylesheets/scss/_landing.scss
+++ b/app/assets/stylesheets/scss/_landing.scss
@@ -358,10 +358,14 @@
 
     li {
       margin:.25ch 0;
-      padding: 0;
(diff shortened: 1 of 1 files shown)


alinvetian:
@ahamelers the page also has this issue, is this something you left intended?
image

ahamelers:
@alinvetian Yes, that's not an issue, the warning-color symbols are special indicators for admins that the values in those fields are unmatched in our database (for institutions/journals) or unverified links. You can hover over to see what is indicated. The symbols do not appear publicly or for non-admin users.

PR #1858, opened 2024-09-23
Include error messages in API logs

ryscher:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3665

To aid in debugging, log more of the errors thrown by the API.


app/controllers/stash_api/files_controller.rb
```diff
@@ -23,6 +23,10 @@ class FilesController < ApiApplicationController
     before_action :require_viewable_file, only: %i[show download]
     before_action -> { require_viewable_resource(resource_id: params[:version_id]) }, only: :index
 
+    def api_logger
```
alinvetian, line 26: This method can be removed since it is defined in parent controller

app/controllers/api_application_controller.rb
```diff
@@ -35,25 +35,35 @@ def require_json_headers
...
     render json: { error: UNACCEPTABLE_MSG }.to_json, status: 406
   end
 
   def require_stash_identifier(doi:)
     # check to see if the identifier is actually an id and not a DOI first
     @stash_identifier = StashEngine::Identifier.where(id: doi).first
-    @stash_identifier = StashEngine::Identifier.find_with_id(doi) if @stash_identifier.blank?
-    render json: { error: 'not-found' }.to_json, status: 404 if @stash_identifier.blank?
+
+    return unless @stash_identifier.blank?
```
alinvetian, line 46: Try rewriting these returns to used `if` so they will be easier to read

PR #1864, opened 2024-09-25
Move tenant logos to their own table

ahamelers:
Using the data URI for tenant logo images works well, but these can be huge in the database, so storing them in the tenants table itself can make it annoying and unwieldy to work with tenants in the rails console. It was driving me a little crazy. This PR moves the logos to their own table (which could be used for other member logos as well, if desired)


ahamelers:
We could also consider moving cedar_json to its own table

alinvetian:
This LGTM.
Please note that once this is released we need to updated the field type for `logo_id` from text to `integer` as it is a FK

alinvetian (approved):

PR #1870, opened 2024-10-07
Change downloadable files check

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3615

I figured out the permissions checks use this one method, and nothing else does, so changed it to check previous resources as well


ryscher:
I updated this method on the prod servers, and it appears to be working correctly now.

PR #1881, opened 2024-10-14
Remove blackout check

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3535


alinvetian:
@ahamelers Currently in the database we have no Journal with `allow_blackout: true` and 187 with `allow_blackout: nil`
Do we still want to keep this flag? or we can do a code cleanup on this?

ahamelers:
@alinvetian yes, I updated the journals in the database so these would no longer cause the curators problems while waiting for this PR. We don't usually delete now-unused fields from the database entirely (I'm not sure why not, exactly, and I've suggested it before. We can discuss in a dev meeting as there are a lot of them floating around!)

PR #1888, opened 2024-10-17
added random submitters report

alinvetian:
Closes: https://github.com/datadryad/dryad-product-roadmap/issues/3680

I am not sure this is still needed, but since I created the report generator, we should keep it in case we'll need it in future


ahamelers:
Sarah was just asking for this, I'll just check with her that this output is good

ahamelers:
I'll just call it good and we'll update it if she wants 😄

PR #1898, opened 2024-10-29
added support for peer review email custom text

alinvetian:
Closes: https://github.com/datadryad/dryad-product-roadmap/issues/3734

I also removed user_mailer views footer text since it is now in mailer layout


ahamelers:
@alinvetian I don't think this should hold this from being merged or anything, but it might be better to make like a 'custom_email_text' table and link each journal to that—I learned with the tenant logos how annoying it could become to have potentially very long content like this on the main table.

Also, @ryscher is there any potential that other member types than journals would want to add custom email text in the future?

alinvetian:
@ahamelers I know having a separate table might be better, but at the moment we only have this simple customization. If we decide to do more custom emails in future I will update this field too, but there is no point at the moment

PR #1911, opened 2024-11-06
Reduce error report emails

ahamelers:
Various exception fixes

### DIFF
--- a/app/controllers/stash_engine/landing_controller.rb
+++ b/app/controllers/stash_engine/landing_controller.rb
@@ -82,7 +82,7 @@ module StashEngine
     end
 
     def identifier_from(params)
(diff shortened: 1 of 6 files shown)


app/controllers/stash_engine/landing_controller.rb
```diff
@@ -82,6 +82,8 @@ def ensure_has_geolocation!
     end
 
     def identifier_from(params)
+      logger.error("Can't parse identifier from nil id param'") && return unless params[:id].present?
```
alinvetian, line 85: Should be 
```
logger.error("Can't parse identifier from nil id param") && return unless params[:id].present?
```

alinvetian (approved):

PR #1912, opened 2024-11-06
Add UI for creating members

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3653

### DIFF
--- a/app/assets/stylesheets/scss/_admin-layout.scss
+++ b/app/assets/stylesheets/scss/_admin-layout.scss
@@ -70,6 +70,20 @@
   align-items: center;
 }
 
+#admin_search {
+  select {
+    flex-shrink: 1;
+    flex-grow: 1;
+    min-width: 0;
+    max-width: 300px;
+    text-overflow: ellipsis;
+  }
+  input[type="submit"] {
+    padding-right: $spacing-md;
+    padding-left: $spacing-md;
+  }
+}
+
 .o-admin-form-pair {
   display: inline-flex;
   gap: $spacing-base/2;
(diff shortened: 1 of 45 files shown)


alinvetian (commented):
1. Please show `Add new` buttons on all 3 pages only if the user has permissions to create
2. Shouldn't there be a `delete` action?
3. Please make at least the publisher name required. Currently the user can add a record without adding any info
4. Models validation would be nice

**Suggestion:**
- The table layout is too crowded in case the user can edit the records, too many pencils. I think it would be nicer if we add an edit link as last column in case the user has permissions which will open the same for as `new` action.
- Rename `_new.html.erb` to `_form.html.erb` 
- Also this method that you used generated too many partial files. `new` and `edit` templates, rendering the `_form` partial, CRUD way

ahamelers:
> The table layout is too crowded in case the user can edit the records, too many pencils. I think it would be nicer if we add an edit link as last column in case the user has permissions which will open the same for as new action.

This is a very reasonable suggestion but would change a lot more code and would change the mechanism users are currently used to

alinvetian (changes requested):
Changes requested

ahamelers:
> Also this method that you used generated too many partial files. new and edit templates, rendering the _form partial, CRUD way
While true, this is also very funny, as these admin pages used to have entirely separate controller methods, javascript partials, and html partials for *every single editable field*, and I reduced these a lot in an earlier PR!

> The table layout is too crowded in case the user can edit the records, too many pencils.
Looking again I think this is only the case for Members and Journals, although we might want to have the same mechanism in any case across all 4 account types. Reducing these to just a single edit form would also reduce all those extra files more. However there are things that should not be editable after creation, like the tenant ID, which will need to be considered. I think we'll want to ask the users who use these edit buttons (basically Jess, Maria) what they prefer and then make any consolidation change in another PR.

PR #1918, opened 2024-11-07
Adjust role association & preloading

ahamelers:
May help with https://github.com/datadryad/dryad-product-roadmap/issues/3755


ahamelers:
This is deployed and it did in fact speed things up!

PR #1937, opened 2024-11-25
Fix problem with calculating curation stats

ryscher:
Allow the loop to process correctly. `find_each` does not work over ranges.


ryscher:
(This fix is already deployed on prod-1)

PR #1939, opened 2024-11-26
2544 allow superusers to withdraw data submissions from any status

alinvetian:
Closes: https://github.com/datadryad/dryad-product-roadmap/issues/2544
Closes: https://github.com/datadryad/dryad-product-roadmap/issues/1959


app/views/stash_engine/admin_dashboard/_curation_form.html.erb
ahamelers, line 1: I do think this permissions check should be moved to the policies. The easiest thing would probably be to make something in the resource policy (https://github.com/datadryad/dryad-app/blob/main/app/policies/stash_engine/resource_policy.rb) like
```ruby
def change_status?
  @resource.curatable? || @user.superuser?
end
```
Then change this check to `<% if policy(@resource).change_status?  && filter_status_select(@resource.current_curation_status) %>`

ahamelers (changes requested):

ahamelers (approved):

PR #1940, opened 2024-11-29
added option for superuser to delete a dataset

alinvetian:
Closes: https://github.com/datadryad/dryad-product-roadmap/issues/2032


app/helpers/stash_engine/identifiers_helper.rb
ahamelers, line 1: This should not be in a new helper but should be moved to the relevant policy and combined with the existing check for "authorize". It might make sense to move it out of the admin_datasets policy and into the identifier pundit policy, above the Scope: https://github.com/datadryad/dryad-app/blob/main/app/policies/stash_engine/identifier_policy.rb

The check in the policy could be like:
```ruby
module StashEngine
  class IdentifierPolicy < ApplicationPolicy
    def initialize(user, identifier)
      @user = user
      @identifier = identifier
    end

    def destroy?
       @user.superuser?
       @identifier.resources.count == 1 &&
          @identifier.resources.first.curation_activities.pluck(:status).uniq == ['in_progress']
    end
```
(there's an existing all-application policy initialization, but it calls every model `@resource` which I feel might be confusing in this case 😆 )

You would check it in `app/controllers/stash_engine/admin_datasets_controller.rb` very simply, like
```ruby
def destroy
      identifier = authorize Identifier.find(params[:id])
```
And you would delete the extraneous "IdentifiersHelper" and the method from the admin datasets policy

app/controllers/stash_engine/admin_datasets_controller.rb
```diff
@@ -58,6 +58,21 @@ def create_salesforce_case
       redirect_to(sf_url, allow_other_host: true)
     end
 
+    def destroy
+      authorize %i[stash_engine admin_datasets], :delete_dataset?
+
+      identifier = Identifier.find(params[:id])
+      unless helpers.can_user_delete_identifier?(identifier)
+        redirect_to activity_log_path(identifier.id), alert: 'This dataset can not be deleted.' and return
```
ahamelers, line 66: See the other comment. With the pundit policy/authorization correctly implemented, this will automatically redirect and show this error, which I think is sufficient: https://github.com/datadryad/dryad-app/blob/6e9d00c4f605737819e3e8dffa456cb088ff0015/app/controllers/stash_engine/application_controller.rb#L30

But with the check of the policy corrected in the view no one should get that far anyway

app/views/stash_engine/admin_datasets/activity_log.html.erb
```diff
@@ -156,6 +156,18 @@
           Please only do this if you know the author is unable/unwilling to submit it.</p>
       </div>
 
+      <% if policy([:stash_engine, :admin_datasets]).delete_dataset? && can_user_delete_identifier?(@identifier) %>
```
ahamelers, line 159: If you follow the correction to the pundit policy this would just become:
`<% if policy(@identifier).destroy? %>`

ahamelers (changes requested):

app/helpers/stash_engine/identifiers_helper.rb
ahamelers: This makes sense to me

ahamelers (approved):

PR #1956, opened 2024-12-17
Add choice of files to package for download for datasets over zip maximum

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3325

I think this solution is pretty elegant as it uses the same mechanisms for all datasets—it adds the ability to select files for packaging and download, then for datasets under the zip max, it hides the (checked) checkboxes and changes some text (so the experience remains as it currently is on production).

For datasets over the zip max, it allows you to select any files under the maximum, and as you select it disables checkboxes for files that, if you added them, would put you over the max.

While the ability to select/deselect all was requested, I have not implemented this as I can't think of a way to do it that makes sense while not allowing users to select more than the size limit—it would have to do something like start selecting all starting from the top but stop once the limit is reached, which I think might be frustrating and confusing.

The easiest way to test this is to adjust your dev or local `APP_CONFIG.maximums.zip_size` value way down (after doing this and restarting rails you may also need to hard refresh your browser to clear your javascript cache of the previous stored value)

### DIFF
--- a/app/assets/stylesheets/scss/_download.scss
+++ b/app/assets/stylesheets/scss/_download.scss
@@ -3,6 +3,10 @@
 .o-download {
 	margin-top: $spacing-sm;
 	text-align: right;
+
+	#download-select-label {
+		font-size: .98rem; margin-left: 1ch;
+	}
 }
 
 *[class*="o-download__"] {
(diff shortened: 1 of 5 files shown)


alinvetian (approved):
LGTM

PR #1960, opened 2024-12-30
Added ruby filename sanitizer

alinvetian:
Closes: https://github.com/datadryad/dryad-product-roadmap/issues/3733

### DIFF
--- a/app/controllers/stash_api/files_controller.rb
+++ b/app/controllers/stash_api/files_controller.rb
@@ -87,8 +87,9 @@ module StashApi
     # prevent people from sending bad filenames
     def sanitize_filename(filename)
       @original_filename = filename
+      file_name = StashEngine::FilenameSanitizer.new(filename).process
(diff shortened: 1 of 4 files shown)


ahamelers:
@alinvetian is it possible we could change when file names are sanitized during upload, so that they can be sanitized using this new code, so that we can remove the JS sanitizer and only maintain this ruby one?

alinvetian:
@ahamelers this is possible, but this means that on upload page, we need to make an AJAX call so that we can sanitize the file names in ruby. Do you want to do this?

ahamelers:
@alinvetian The upload files page is already making a bunch of calls to controllers with axios. Couldn't this piggyback on one of those?

alinvetian:
@ahamelers unfortunately none of those can be used, this call needs to be triggered on the moment the user selects a file, but before showing the file name in the table.

PR #1975, opened 2025-01-13
Update scripts to upload CSV reports to S3

alinvetian:
Closes: https://github.com/datadryad/dryad-product-roadmap/issues/3863

### DIFF
--- a/config/app_config.yml
+++ b/config/app_config.yml
@@ -198,6 +198,7 @@ defaults: &DEFAULTS
     region: us-west-2
     bucket: dryad-s3-dev
     merritt_bucket: dryad-assetstore-merritt-dev
+    reports_bucket: dryad-reports
     key: AKIA2KERHV5E3OITXZXC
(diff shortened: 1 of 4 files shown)


ahamelers:
@alinvetian It looks like there are some other CSV reports you haven't adjusted:

https://github.com/datadryad/dryad-app/blob/b93833db5483e25364710b183227d2bc60f8145d/lib/tasks/stash_engine_tasks.rake#L426
https://github.com/datadryad/dryad-app/blob/b93833db5483e25364710b183227d2bc60f8145d/lib/tasks/stash_engine_tasks.rake#L456

Also, there's an API reports endpoint, that you may need to adjust to check this new bucket:

https://github.com/datadryad/dryad-app/blob/b93833db5483e25364710b183227d2bc60f8145d/app/controllers/stash_api/api_controller.rb#L25

Possibly we should move ALL reports, including PDF, to use this new bucket instead of the `REPORTS_DIR` variable. @ryscher ?

PR #1961, opened 2025-01-03
Newrelic integration

alinvetian:
Closes: https://github.com/datadryad/dryad-product-roadmap/issues/3856

PR #1980, opened 2025-01-16
Add Newrelic integration

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3856

Redo of #1961 with hidden key


alinvetian:
`Newrelic` and `rack-mini-profiler` gems seem not to like each other so removing mini-profiler for the moment

PR #1983, opened 2025-01-22
Account editing cleanup

ahamelers:
Suggested from https://github.com/datadryad/dryad-app/pull/1912

Closes https://github.com/datadryad/dryad-product-roadmap/issues/3881

Blocking https://github.com/datadryad/dryad-product-roadmap/issues/3549


alinvetian (commented):
Maybe in the future we can add type validation on email and ISSN fields.
Currently if you enter incorrect value format there is no error and the values are not saved

ahamelers:
> Maybe in the future we can add type validation on email and ISSN fields. Currently if you enter incorrect value format there is no error and the values are not saved

I think we have type validation for these on the model, do you mean we need client side validation as well?

alinvetian:
Yes, I was talking about adding the validations on the forms too.
Currently it behaves like in production so we can make this a future story.

ahamelers:
@alinvetian I've added basic validation for user email and journal issns

ahamelers:
I've added quite complicated validation, including on creation, for ISSNs and for all email addresses including those in arrays in the non-user models

app/models/stash_engine/tenant.rb
```diff
@@ -28,11 +28,13 @@ class Tenant < ApplicationRecord
     validates :id, presence: true, uniqueness: true
     validates :short_name, presence: true
     validates :long_name, presence: true
+    EMAIL_REGEX = /\A[\w+\-.]+@[a-z\d-]+(\.[a-z\d-]+)*\.[a-z]+\z/i
```
alinvetiansage, line 31: Please move this to `constants.rb` since it is repeated in 4 places

alinvetiansage (approved):

alinvetian (approved):

PR #1989, opened 2025-01-30
Add ability for admins to flag accounts and datasets

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3549


app/controllers/stash_engine/journal_admin_controller.rb
```diff
@@ -76,13 +76,19 @@ def setup_sponsors
...
 
     def update_hash
-      valid = %i[title default_to_ppr payment_plan_type sponsor_id]
+      valid = %i[title default_to_ppr]
       update = edit_params.slice(*valid).to_h
-      update[:sponsor_id] = nil if edit_params.key?(:sponsor_id) && edit_params[:sponsor_id].blank?
-      update[:payment_plan_type] = nil if edit_params.key?(:payment_plan_type) && edit_params[:payment_plan_type].blank?
-      update[:notify_contacts] = edit_params[:notify_contacts].split("\n").map(&:strip).to_json if edit_params.key?(:notify_contacts)
-      update[:review_contacts] = edit_params[:review_contacts].split("\n").map(&:strip).to_json if edit_params.key?(:review_contacts)
+      update[:sponsor_id] = edit_params[:sponsor_id].presence || nil
```
ryscher, line 81: The `|| nil` is redundant. That's the purpose of `presence` -- it returns `nil` if there is any empty value.

ryscher: This construction appears several times throughout the PR.

ryscher:
Minor comment above, but otherwise this is good for me. Can merge whenever the curators approve.

app/controllers/stash_engine/journal_admin_controller.rb
ahamelers: I think this came from a rubocop -A struggle to reduce the ABC size or something like that, that I eventually gave up on and added a disable comment for. Will fix

PR #2009, opened 2025-02-18
Exception fixes

ahamelers:
Fixing exceptions
```
  undefined method `squish' for nil
  app/controllers/stash_datacite/contributors_controller.rb:115:in `find_or_initialize'
```
```
  uninitialized constant ApiApplicationController::UNACCEPTABLE_MSG
  app/controllers/api_application_controller.rb:41:in `require_json_headers'
```
```
  undefined method `to_i' for an instance of ActionController::Parameters
  app/controllers/api_application_controller.rb:18:in `page'
```
```
  undefined local variable or method `publication' for an instance of #<Class:0x00007f2e8e352a00>
  app/views/stash_engine/admin_datasets/_publication_popup.html.erb:2
```

### DIFF
--- a/app/controllers/api_application_controller.rb
+++ b/app/controllers/api_application_controller.rb
@@ -15,11 +15,11 @@ class ApiApplicationController < StashEngine::ApplicationController
   skip_before_action :verify_authenticity_token
 
   def page
-    @page ||= (params[:page].to_i.positive? ? params[:page].to_i : 1)
+    @page ||= (params[:page].respond_to?(:to_i) && params[:page].to_i.positive? ? params[:page].to_i : 1)
   end
 
   def per_page
-    [params['per_page']&.to_i || DEFAULT_PAGE_SIZE, 100].min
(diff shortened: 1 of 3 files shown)


app/controllers/api_application_controller.rb
```diff
@@ -15,11 +15,11 @@ class ApiApplicationController < StashEngine::ApplicationController
   skip_before_action :verify_authenticity_token
 
   def page
-    @page ||= (params[:page].to_i.positive? ? params[:page].to_i : 1)
+    @page ||= (params[:page].respond_to?(:to_i) && params[:page].to_i.positive? ? params[:page].to_i : 1)
```
alinvetian, line 18: you can use 
```
@page ||= [params[:page].to_s.to_i, 1].max
```

ahamelers: I feel like this is less clear to read, even though it is shorter

PR #2017, opened 2025-02-25
Soft delete resources

alinvetian:
Closes: https://github.com/datadryad/dryad-product-roadmap/issues/3953

Also removed code related to old/custom enum implementation


ryscher:
Bug: When a dataset has two resources, and the most recent resource is destroyed, the Activity Log page no longer displays. 

Concern: When a resource is destroyed, the associated objects have different behaviors:
- stash_engine_versions and stash_engine_curation_activities have a deleted_at field set the same as the resource (this is what I would expect)
-  stash_engine_resource_states and stash_engine_authors are actually deleted (is this intentional?)
- dcs_subjects_stash_engine_resources are left unchanged (is this intentional?)

alinvetian:
@ryscher related to your concerns:
1. Each record has it's own deleted_at value, as in time the object got deleted. This way we know when the record was deleted, and not the parent resource.
2. and 3. Since this is the initial implementation of soft delete, I added only a few models that i thought are more important. We can add the deleted_at to more models as we get more info on what we need to track/show after being deleted.

ryscher:
Error still exists. See new example on dev at https://v3-dev.datadryad.org/ds_admin/3579

Error in log:
```
Information for: ActionView::Template::Error (No route matches {:action=>"dupe_check", :admin=>true, :controller=>"stash_engine/resources", :format=>:js, :id=>nil}, possible unmatched constraints: [:id]):
     6:   <h1 class="o-heading__level1"><%= r&.title %></h1>
     7:   <%= @identifier.identifier %>
     8: </div>
     9: <div id="dupe_check" class="callout warn" data-load="<%= stash_url_helpers.dupe_check_resource_path(format: :js, id: r&.id, admin: true) %>"></div>
    10: <div class="callout alt">
    11:   <div id="activity-at-a-glance">
    12:     <div class="input-stack">
  
app/views/stash_engine/admin_datasets/index.html.erb:9
app/controllers/application_controller.rb:3:in `process_action'
```

alinvetian:
The error appears due to using `resource .destroy` instead of `StashEngine::DeleteDatasetsService.new(StashEngine::Resource.find(<ID>), add_delete_note: false).call`
which does not always update `latest_resource_id`

PR #2027, opened 2025-02-28
Add CC0 questions and additional HSI information to submission validation

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3946

Also change some headings to questions to reinforce user attention (Jen suggestion)


alinvetian (approved):

alinvetian (changes requested):
@ahamelers As per @ryscher comment, the CC0 license checkbox needs to be unchecked by default

ahamelers:
@alinvetian It is, but only on new dataset submissions.

ahamelers:
@alinvetian Also, please don't merge this until it's been deployed to staging and the curators have had a change to approve

jeremywalter (changes requested):
Two things: Firstly we adjusted the text for the HSI declaration to hopefully head off users entering 1. and 2. Text is in the Validation doc: https://docs.google.com/document/d/1fPFWbTwulf3HatWh9hMoRP8dAVoIymtBFDSINLexHvE/edit?usp=sharing (not sure if this got lost). The UX for the HSI text box is a little fiddly. We might want to have a 'submit' button or similar so users can indicate they are done entering text? 

Here is the HSI text prompt: 

**Please explain how you de-identified the data AND confirm that you received explicit consent from your participants to publish the de-identified data in the public domain.**

ahamelers:
I've altered the prompt a bit—the all caps and bold isn't really in house style or easy to read—but got rid of the numbering as requested

> Second: The UX for the HSI text box is a little fiddly. We might want to have a 'submit' button or similar so users can indicate they are done entering text?

It's the same UI as for every other text input, adding a button to submit would make it very different from the rest of the inputs

jeremywalter:
The bold formatting wasn't so necessary as avoiding the numbering which we think would result in some pretty awkward responses (authors will use the numbering when they enter information and it wouldn't have context on the README display - unless the display is designed that way?). Text update looks okay to me, and ready to see how it goes on my other comment. Excited to see this in action!

jeremywalter (approved):

PR #2028, opened 2025-02-28
Open and close main menu items with hover

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3891

### DIFF
--- a/public/javascript/interactions.js
+++ b/public/javascript/interactions.js
@@ -57,24 +57,28 @@ if (timezone) {
 var navButtons = Array.from(document.getElementsByClassName('c-header_nav-button'));
 navButtons.forEach(button => {
   button.parentElement.addEventListener('mouseenter', (e) => {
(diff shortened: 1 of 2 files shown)


alinvetian:
@ahamelers this works OK on web, but now, on mobile devices yoou have to click the menu item twice in order for it to open

alinvetian:
@ahamelers the issue is now fixed on mobile view, but it still persists for larger mobile devices like iPad Pro
Here is the resolution:
image

ahamelers:
@alinvetian This is not true, in fact now that I see this, I believe the earlier change was also not necessary. I believe it is a problem with the chrome testing environment. Actual mobile devices do not have a hover state and shouldn't respond to mouseenter and mouseleave unless the user is dragging their finger. The dev environment in firefox with touch enabled seems to react correctly, it's only on chrome devtools that this happens, which makes me think it's an issue there. I guess we'd need to test it on a real device of that size to be absolutely, 100% positive.

PR #2034, opened 2025-03-05
Update resource destroy

alinvetian:
Given a published dataset where user created a ne in progress version.
Currently when the user deleted the in progress version, the identifier was not updated with the ID of the previous resource.
The results are:
- the published dataset does not appear in the admin dashboard and in admin search
- activity log page is not loading due to errors

Right now, there is no identifier in production that has a bad `latest_resource_id`, so no data migration is needed


ahamelers:
I'm concerned this isn't resolving the problem in every deletion scenario. Made a ticket: https://github.com/datadryad/dryad-product-roadmap/issues/3978

PR #2035, opened 2025-03-05
Add admin functionality to the dataset activity page

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3943
Closes https://github.com/datadryad/dryad-product-roadmap/issues/2471
Closes https://github.com/datadryad/dryad-product-roadmap/issues/2228
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3945
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3975
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3979


alinvetian:
Bugs:
- `Notification date` saving empty value has an error and does not show the error message
- `Funders` 
  - You should not be allowed to save empty rows
  - Already saved rows are searching for contributors instead of affiliations
  - `Close dialog` does not work after clicking any save ore delete in the modal
- `Curation state` just open and save will create an log empty entry without any meaning

You can add a left border the these 3 columns so it is more intuitive that the data from them is in presented vertically, not horizontally like the rest  
image

Please also address this https://github.com/datadryad/dryad-product-roadmap/issues/3975#issuecomment-2704451464

alinvetian (changes requested):
Bugs:
- `Notification date` saving empty value has an error and does not show the error message
- `Funders` 
  - You should not be allowed to save empty rows
  - Already saved rows are searching for contributors instead of affiliations
  - `Close dialog` does not work after clicking any save ore delete in the modal
- `Curation state` just open and save will create an log empty entry without any meaning

You can add a left border the these 3 columns so it is more intuitive that the data from them is in presented vertically, not horizontally like the rest  
image

Please also address this https://github.com/datadryad/dryad-product-roadmap/issues/3975#issuecomment-2704451464

PR #2044, opened 2025-03-10
Lower throttling limits for downloads

ryscher:
More work towards controlling [S3 fees](https://github.com/datadryad/dryad-product-roadmap/issues/3980)

We may need to reverse these if legitimate users complain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # Dryad
-
+ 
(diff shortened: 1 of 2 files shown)


ahamelers:
@ryscher are the rate limits different for logged in users?

ryscher:
Hmmm, no. There are different limits for API users, because we can easily distinguish them using the HTTP headers. But RackAttack doesn't have access to know whether a GUI user is logged in.

ryscher:
Making this a draft until we are certain it is needed.

ryscher:
Will not use -- investigating other bot mitigation

PR #2077, opened 2025-03-20
Do not send invoice for re-publishing when not stepping over 10GB limit

alinvetian:
Closes: https://github.com/datadryad/dryad-product-roadmap/issues/4031

### DIFF
--- a/lib/stash/payments/invoicer.rb
+++ b/lib/stash/payments/invoicer.rb
@@ -55,7 +55,8 @@ module Stash
         return unless customer_id.present?
 
         lfs = APP_CONFIG.payments.large_file_size
-        return unless ds_size > lfs && ds_size > prev_size && (ds_size / 10).floor != (prev_size / 10).floor
+        overage_step = 10_000_000_000 # 10GB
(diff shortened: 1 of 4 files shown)


lib/stash/payments/invoicer.rb
```diff
@@ -47,20 +48,25 @@ def charge_user_via_invoice
...
+        res
       end
 
       def check_new_overages(prev_size)
         customer_id = stripe_user_customer_id
         return unless customer_id.present?
 
         lfs = APP_CONFIG.payments.large_file_size
-        return unless ds_size > lfs && ds_size > prev_size && (ds_size / 10).floor != (prev_size / 10).floor
+        overage_step = 10_000_000_000 # 10GB
```
ahamelers, line 62: I think this should be `APP_CONFIG.payments.additional_storage_chunk_size` rather than hard coded 10GB

ahamelers:
@alinvetian Added one comment, and a commit where I used the new column and fixed the calculation in the front end, too. As discussed, with the new fee system work, we will scrap this stuff and use a calculation service API call instead

ahamelers (changes requested):

ahamelers (approved):

PR #2082, opened 2025-03-21
Cleanup unused files

ahamelers:
Unused views cleanup


ahamelers:
routes and controllers could also be cleaned up

PR #2110, opened 2025-03-28
Create fee calculator UI

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3989

I also added the `covers_storage_fee` option for the calculator for institutions, as we were directed to do.

The oversized storage percentage selector is, by default, set to 100% at tier 0 (under 10GB). The calculator service controller is using a `get_tier_by_value` method that apparently defaults to tier 1, ignoring tier 0, so it returns an incorrect total (treating the tier 0 items, which should have no charge, as if they should be charged at tier 1).

I figure @alinvetian will fix this bug? Probably in https://github.com/datadryad/dryad-app/pull/2106 ? But let me know if I should fix it instead.

### DIFF
--- a/app/views/fee_calculator/_institution.html.erb
+++ b/app/views/fee_calculator/_institution.html.erb
@@ -40,7 +40,7 @@
   </p>
   <div class="input-line spaced" style="justify-content: stretch;">
     <h3 style="font-weight: bold">Annual partner fee estimate:</h3>
-    <p id="total_estimate">5000</p>
+    <output id="total_estimate" for="service_tier dpc_tier">5000</output>
     <p><%= form.submit 'Recalculate', class: 'o-button__plain-text2' %></p>
   </div>
 <% end %>
--- a/app/views/fee_calculator/_institution.html.erb
+++ b/app/views/fee_calculator/_institution.html.erb
@@ -10,22 +10,22 @@
   </p>
   <p class="input-line center" id="service_tier0">
     <%= form.label :service_tier, 'Your annual research expenditure:' %>
-    <%= form.select :service_tier, options_for_select(FeeCalculator::InstitutionService.service_fee_tiers.map {|t| [t[:range].to_s.split('..').map {|n| n.to_i.to_s == n ? (n.to_i > 999999 ? number_to_currency(number_to_human(n), precision: 0) : number_to_currency(n, precision: 0)) : nil}.join('-').gsub(/\-$/, '+').downcase, t[:tier]]}), {}, {class: 'c-input__select'} %>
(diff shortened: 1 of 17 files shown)


public/javascript/calculator.js
```diff
@@ -0,0 +1,30 @@
...
+          or.nextElementSibling.value = newVal;
+        }
+      })
+    }
+  })
+})
+$(document).on("ajax:complete", function(status, response){
+  if (response.status === 200) {
+    const {fees: {total}} = response.responseJSON;
+    document.getElementById('total_estimate').innerHTML = total;
```
alinvetian, line 28: Please use currency format in here:
```
(total).toLocaleString('en-US', {style: 'currency', currency: 'USD'})  /* $2,500.00 */
```

app/views/fee_calculator/_publisher.html.erb
```diff
@@ -0,0 +1,36 @@
...
+  <p class="input-line center" style="margin-top: 2rem;">
+    <span id="legend2">Will you cover large data storage fees for your authors?</span>
+    <span role="group" aria-labelledby="legend2" class="radio_choice">
+      <label><input type="radio" name="cover_storage_fee" value="1" checked/>Yes</label>
+      <label><input type="radio" name="cover_storage_fee" value="0"/>No</label>
+    </span>
+  </p>
+  <div class="input-line spaced" style="justify-content: stretch;">
+    <h3 style="font-weight: bold">Annual partner fee estimate:</h3>
+    <output id="total_estimate" for="service_tier dpc_tier">1000</output>
```
alinvetian, line 32: Do not hardcode this value

app/views/fee_calculator/_institution.html.erb
```diff
@@ -0,0 +1,62 @@
...
+  <p class="input-line center" style="margin-top: 2rem;">
+    <span id="legend2">Will you cover large data storage fees for your authors?</span>
+    <span role="group" aria-labelledby="legend2" class="radio_choice">
+      <label><input type="radio" name="cover_storage_fee" value="1" checked/>Yes</label>
+      <label><input type="radio" name="cover_storage_fee" value="0"/>No</label>
+    </span>
+  </p>
+  <div class="input-line spaced" style="justify-content: stretch;">
+    <h3 style="font-weight: bold">Annual partner fee estimate:</h3>
+    <output id="total_estimate" for="service_tier dpc_tier">5000</output>
```
alinvetian, line 43: Do not hardcode this value.

alinvetian (changes requested):
@ahamelers Please trigger a calculation call for default values on page load, so that if the values change for default tiers, we do not need to update the HTML files.

app/views/fee_calculator/_publisher.html.erb
ahamelers: I'm fine with doing this change, but unless we also make all the tables based on some call to an API, we will need to update the HTML files whenever the tiers change to update all the tables.

app/views/fee_calculator/_institution.html.erb
ahamelers: I mainly hardcoded it because the initial response was wrong.

public/javascript/calculator.js
ahamelers: Having decimals is silly

PR #2112, opened 2025-03-31
Fix exception

ahamelers:
```
A TypeError occurred in datasets#create:

  no implicit conversion of nil into Array
  app/controllers/stash_api/datasets_controller.rb:515:in `intersect?'
```

### DIFF
--- a/app/controllers/stash_api/datasets_controller.rb
+++ b/app/controllers/stash_api/datasets_controller.rb
@@ -512,7 +512,7 @@ module StashApi
         end
         # admin restrictions
         if %w[skipEmails preserveCurationStatus].include?(attr) && item_value.instance_of?(TrueClass) &&
-          !(@user.min_curator? || @user.journals_as_admin.intersect?(@resource&.journals))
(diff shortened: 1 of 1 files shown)


app/controllers/stash_api/datasets_controller.rb
```diff
@@ -512,7 +512,7 @@ def check_restricted_params
         end
         # admin restrictions
         if %w[skipEmails preserveCurationStatus].include?(attr) && item_value.instance_of?(TrueClass) &&
-          !(@user.min_curator? || @user.journals_as_admin.intersect?(@resource&.journals))
+          !(@user.min_curator? || @user.journals_as_admin?.intersect?(@resource&.journals))
```
alinvetian, line 515: I think the issue in here is `@resource&.journals`, it return s nil instead of an array
I would use as a fix:
```
!(@user.min_curator? || @user.journals_as_admin?.intersect?(@resource&.journals || []))
```

ahamelers: Also I used javascript `?` instead of ruby `&`! too much switching between languages recently

PR #2126, opened 2025-04-07
Fix too-wide file list table

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/4089

Breaks long filenames to prevent the file list table from extending beyond page width. Also fixes a couple weird things in the table (The "Actions" column only ever contained "Remove", so it was silly to call it "Actions"; the text "Removing..." would appear under the "Tabular data check" table header, which made no sense)

### DIFF
--- a/app/assets/stylesheets/scss/_upload-table.scss
+++ b/app/assets/stylesheets/scss/_upload-table.scss
@@ -49,6 +49,8 @@
 
   a {
     @extend %o-link__primary;
+    word-break: normal;
+    overflow-wrap: anywhere;
   }
 
   button {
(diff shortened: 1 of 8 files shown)


alinvetian:
This does not fixed the preview page
image

alinvetian (changes requested):
See above comment

ahamelers:
@alinvetian that's not what the closed issue was about at all, it was specifically about the file table and pushing Remove buttons off screen. If you want to report that, please create a new issue.

(Also, nobody has ever made a filename THAT long)

PR #2143, opened 2025-04-11
Improve duplicate check for curators

ahamelers:
1. Remove modify_permission check from dupe_check so all curators can see the results, not just the curator assigned to the submission
2. Make the check for bad strings (`NA, N/A, TBD, unknown`) case insensitive (the submission in question had 'n/a', lowercase)

Closes https://github.com/datadryad/dryad-product-roadmap/issues/4107

### DIFF
--- a/app/controllers/stash_engine/resources_controller.rb
+++ b/app/controllers/stash_engine/resources_controller.rb
@@ -4,6 +4,7 @@ module StashEngine
     include StashEngine::LandingHelper
 
     before_action :require_login
+    before_action :assign_resource, only: %i[logout display_readme dupe_check]
     before_action :require_modify_permission, except: %i[index new logout display_readme dupe_check]
     before_action :require_in_progress, only: %i[upload review upload_manifest up_code up_code_manifest]
     # before_action :lockout_incompatible_uploads, only: %i[upload upload_manifest]
@@ -108,7 +109,6 @@ module StashEngine
     end
 
     def logout
-      @resource = resource
       @resource.update_columns(current_editor_id: nil)
       respond_to do |format|
         format.html { redirect_to dashboard_path }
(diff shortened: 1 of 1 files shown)


app/controllers/stash_engine/resources_controller.rb
```diff
@@ -184,12 +183,12 @@ def dupe_check
         primary_article = @resource.related_identifiers.find_by(work_type: 'primary_article')&.related_identifier
         manuscript = @resource.resource_publication.manuscript_number
         dupes = other_submissions.where(title: @resource.title)&.select(:id, :title, :identifier_id).to_a
-        if primary_article.present? && !['NA', 'N/A', 'TBD', 'unknown'].include?(primary_article)
+        if primary_article.present? && ['NA', 'N/A', 'TBD', 'unknown'].none? { |s| s.casecmp(primary_article) == 0 }
```
alinvetian, line 186: we should be using `casecmp?()` which return true or false
`casecm()?` and return -1, 0, 1, or nil

app/controllers/stash_engine/resources_controller.rb
```diff
@@ -184,12 +183,12 @@ def dupe_check
...
           dupes.concat(other_submissions.joins(:related_identifiers)
               .where(related_identifiers: { work_type: 'primary_article', related_identifier: primary_article })
               &.select(:id, :title, :identifier_id).to_a)
         end
-        if manuscript.present? && !['NA', 'N/A', 'TBD', 'unknown'].include?(manuscript)
+        if manuscript.present? && ['NA', 'N/A', 'TBD', 'unknown'].none? { |s| s.casecmp(manuscript) == 0 }
```
alinvetian, line 191: plaase use `casecmp?()`, same as above

PR #2158, opened 2025-04-24
3999 verify ability to handle large files 2tb

alinvetian:
Closes: https://github.com/datadryad/dryad-product-roadmap/issues/3999


alinvetian:
Do we also want to change the `:merritt_size` value to 2TB?

ahamelers:
> Do we also want to change the `:merritt_size` value to 2TB?

Yes, you'll need to

ahamelers:
@alinvetian Leave this to me actually, I think we may need to have separate sizes for the old and new payment systems for the time being.

ryscher:
Merging this, though there are other activities required to fully close the ticket.

PR #2159, opened 2025-04-24
Adjust sentence caser

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/4135

Changed the handling to lowercase unknown words when the whole sentence is all caps, and added a bunch of comments. Also fixed plural proper noun handling.

Titles previously used for testing can be found in https://github.com/datadryad/dryad-app/pull/2081#issuecomment-2744031254

### DIFF
--- a/app/javascript/lib/sentence-case.js
+++ b/app/javascript/lib/sentence-case.js
@@ -74,11 +74,11 @@ export function sentenceCase(input, options = {}) {
       result += whiteSpace;
       continue;
     }
-    // Ignore URLs, email addresses, acronyms, etc.
+    // ignore URLs, email addresses, acronyms with dots, etc.
     if (IS_SPECIAL_CASE.test(token)) {
       const acronym = token.match(IS_ACRONYM);
-      // The period at the end of an acronym is not a new sentence,
-      // but we should uppercase first for i.e., e.g., etc.
+      // the period at the end of an acronym is not a new sentence,
+      // but uppercase first letter for i.e., e.g., etc.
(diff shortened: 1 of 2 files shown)


alinvetian:
@ahamelers `DE` and `de` is still translated to `De` is this the right behavior?
Screenshot 2025-04-25 at 15 11 30

Also i think it would be nice to have a test with all the titles used for testing from https://github.com/datadryad/dryad-app/pull/2081#issuecomment-2744031254

ahamelers:
> DE and de is still translated to De is this the right behavior?

The dictionary probably believes this is part of a surname, which is most often the case

> Here is another strange case with the same words
THE TEST FOR WORDS THAT START WITH DE LIKE DEICE, DECARBOXYLATION, DEMAGNETIZATIOn

I mean, you've got a lowercase letter in there, that means it explicitly fails the "if the sentence is in all caps" check. The check isn't "if every letter except one or two is all caps".

ahamelers:
> Also i think it would be nice to have a test with all the titles used for testing from https://github.com/datadryad/dryad-app/pull/2081#issuecomment-2744031254

Sure a test could be built, but I really don't think there's any need to _run_ tests on this except on explicit demand, since this is a standalone javascript that relies on and is affected by nothing else in the system. No one is going to make a change somewhere else and have this break unexpectedly.

alinvetian:
The test would be for us so when we make any change to the script, we do not break any of the other cases we previous fixed, since in that list are a lot of test cases

PR #2169, opened 2025-04-29
Load files separately for submission editing

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/4145

As a first step, we'll just load file information separately from the rest of the submission data for editing. If this is not sufficient for the problem of large numbers of files, we'll move to possibly loading the files in chunks, and any previous submission files even more separately.

### DIFF
--- a/app/controllers/stash_datacite/metadata_entry_pages_controller.rb
+++ b/app/controllers/stash_datacite/metadata_entry_pages_controller.rb
@@ -19,7 +19,6 @@ module StashDatacite
                     } }]
       )
       @submission[:users] = @resource.users.select('stash_engine_users.*', 'stash_engine_roles.role')
-      find_files
       @submission = @submission.to_json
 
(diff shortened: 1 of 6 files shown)


app/controllers/stash_datacite/metadata_entry_pages_controller.rb
```diff
@@ -34,22 +33,24 @@ def find_or_create
...
-
-      @submission['previous_curated_resource'][:generic_files] = @resource.previous_curated_resource.generic_files
-        .includes(:frictionless_report).validated_table.as_json(
+      if @resource.previous_curated_resource.present?
+        files[:previous_files] = @resource.previous_curated_resource.generic_files.includes(:frictionless_report).validated_table.as_json(
           methods: :type, include: { frictionless_report: { only: %i[report status] } }
         )
+      end
+
+      render json: files
```
alinvetian, line 47: As a note for future, do not return the entire list of attributes, return only the attributes that front-end needs

PR #2183, opened 2025-05-06
Fix help page typo

ahamelers:


config/initializers/assets.rb
```diff
@@ -10,7 +10,7 @@
 Rails.application.config.assets.raise_runtime_errors = true
 
 # Version of your assets, change this if you want to expire all your assets.
-Rails.application.config.assets.version = '20250409'
+Rails.application.config.assets.version = '20250506'
```
alinvetian, line 13: Maybe using the bellow would handle this permanently
```
Rails.application.config.assets.version = Time.current.to_i.to_s
```

alinvetian (approved):

config/initializers/assets.rb
ahamelers: I don't think we want to fully recompile all assets every single day

PR #2189, opened 2025-05-08
Fix contact form exception

ahamelers:
This should fix the exception where people were able to leave off the email body. The recaptcha was also not fully set up and should be now.

I also added a new Case 'Origin', because with what I'm able to set up as far as auto-response emails, I can get something working but then the ones that truly come through email receive the auto-response twice 🤦 . Having a new origin allows me to distinguish them.


alinvetian (approved):
LGTM

PR #2217, opened 2025-05-27
Fix file_view flag

alinvetian:
Closes: https://github.com/datadryad/dryad-product-roadmap/issues/4153

I added 2 edge cases tests please verify and see if they are correct or not


ahamelers (approved):
This looks good to me (good catch on that test!) but I'd like @ryscher to double check in case there's some weird historic reason for any of this

PR #2218, opened 2025-05-27
Make file uploaded check more robust for front end

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/4211

This check now more closely matches the backend (DatasetValidation) check

### DIFF
--- a/app/models/stash_engine/generic_file.rb
+++ b/app/models/stash_engine/generic_file.rb
@@ -237,14 +237,14 @@ module StashEngine
     end
 
     def uploaded_success_url
-      dl_url = s3_staged_presigned_url if file_state.nil? || file_state == 'created'
+      dl_url = s3_staged_presigned_url unless digest? || storage_version_id.present?
       dl_url ||= public_download_url
       dl_url ||= url
       dl_url
     end
 
     def uploaded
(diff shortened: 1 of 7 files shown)


app/models/stash_engine/generic_file.rb
```diff
@@ -237,13 +237,15 @@ def clean_frictionless_json(in_str)
     end
 
     def uploaded_success_url
-      dl_url = s3_staged_presigned_url if file_state.nil? || file_state == 'created'
+      dl_url = s3_staged_presigned_url unless digest? || storage_version_id.present?
```
alinvetian, line 240: Please update this `unless` to `if` so they are easier to read
```
if !digest? && storage_version_id.blank?
```

app/models/stash_engine/generic_file.rb
```diff
@@ -237,13 +237,15 @@ def clean_frictionless_json(in_str)
...
       dl_url ||= public_download_url
       dl_url ||= url
       dl_url
     end
 
     def uploaded
+      return Stash::Aws::S3.new.exists?(s3_key: s3_staged_path) unless digest? || storage_version_id.present?
```
alinvetian, line 247: Same here

PR #2222, opened 2025-05-28
Reauthenticate users with partner institutions

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3337

### DIFF
--- a/app/controllers/stash_engine/sessions_controller.rb
+++ b/app/controllers/stash_engine/sessions_controller.rb
@@ -15,7 +15,7 @@ module StashEngine
     # this is the place omniauth calls back for shibboleth logins
     def callback
       current_user.roles.tenant_roles.delete_all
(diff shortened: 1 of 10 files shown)


ryscher:
When starting a new dataset, I'm getting the error:

```
AbstractController::DoubleRenderError (Render and/or redirect were called multiple times in this action. Please note that you may only call render OR redirect, and at most once per action. Also note that neither redirect nor render terminate execution of the action, so if you want to exit an action after redirecting, you need to do something like "redirect_to(...); return".):
  
app/controllers/stash_engine/shared_security_controller.rb:41:in `require_login'
```

I think all of the `redirect_to` statements in this method need  `and return`

ryscher:
In my testing, adding the `and return` works correctly, so I will add that and then merge this PR.

PR #2236, opened 2025-06-05
Don't block uploads over 300GB

ahamelers:
For https://github.com/datadryad/dryad-product-roadmap/issues/4214


ahamelers:
@alinvetian All users SHOULD be able to upload more than 300GB, and then only super users should be able to submit such datasets. We were never meant to block uploads of more than 300GB. Doing so was introducing a bug. This is a bug fix.

PR #2245, opened 2025-06-09
Release from PPR when accepted publication added

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/4239


app/controllers/stash_datacite/related_identifiers_controller.rb
```diff
@@ -38,7 +40,9 @@ def create
     def update
       respond_to do |format|
         if @related_identifier.update(calc_related_identifier_params)
-          @related_identifier.update(verified: @related_identifier.live_url_valid?)
+          @related_identifier.update(verified: @related_identifier.live_url_valid?) if @related_indentifier.verified.blank?
```
alinvetian, line 43: Please use `unless @related_indentifier.verified?`, `verified` is a boolean field in DB
Why are we updating only if the link is not verified? Maybe the link is became not valid since last update.

PR #2246, opened 2025-06-09
Add autocomplete for manuscript number

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/4226

Autocomplete is limited to manuscripts from the selected journal, and shows a truncated title and the first 3 author surnames


ahamelers:
I have changed this to do the following:

- For users, automatically fill in a prefix if the journal uses one for its manuscript numbers
- For curators (from the activity log page), provide a manuscript lookup/search

PR #2253, opened 2025-06-11
Allow file renaming after upload

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/3963

To rename files without the time and S3-intensive need to copy files to a new S3 location and delete the originals, I separated the s3 key from the filename visible to the user. To do this I added a new column, `download_filename`, which is user editable. `upload_file_name` will, from release on, be a UUID for loading and accessing the file (regardless of user-facing filename) from S3.

Make sure the data migration runs along with the database migration when testing (if your database has null values for `download_filename`, things won't work)


ryscher:
Everything else looks good. All of the actual file naming works correctly.

PR #2262, opened 2025-06-18
4019 add sidekiq for background jobs processing

alinvetian:
Closes: https://github.com/datadryad/dryad-product-roadmap/issues/4019


ryscher:
Oh, I see there is documentation in #2489 -- will have to check that later.

PR #2322, opened 2025-07-18
Create public file change log

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/623

Approved by curation team


ryscher:
Please move the textbox for notes to the curator below the required textbox for the public change long, and note that the notes to the curator are private.

jeremywalter:
When I attempted this in Sandbox, I was not able to fulfill the requirements:
image

jeremywalter:
When entering text into the version prompt in Sandbox I was not able to get past the requirements flag.

jeremywalter:
When I change the files the file change log allows me to submit without changing the text. If I erase the preset date, it does throw an error and stop me from resubmitting. I guess it needs a multi-word minimum entry?

ahamelers:
> This looks good. I would like to get a final ok from Dan before we merge this, due to the pause.

Got this on slack, that's why I took it out of draft.

> I did find it a little odd that the auto-filled text is "Changes after..." instead of "Changes on...", but if that's ok with the curation team, I don't mind it.

Everyone minds this, but the dates are the previous file publication dates, not the upcoming publication date of the revision, which is unknown. At the time the log is being written, we have no idea what the publication date of the in-progress version is going to be (it could take several days for curation, and many back and forths through AAR, between when the log is written and when it is ultimately published)

PR #2325, opened 2025-07-20
Limit helpdesk spam

ryscher:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/4265


ahamelers:
Blocking at 3 form submissions in a month may be too stringent. If someone has multiple submissions, they may have a separate form submission for each one, and that would be fine.

ahamelers:
@ryscher In curation meeting we decided this should trigger if 10 form submissions are attempted in a day and block the user from submitting the form for 2 weeks

PR #2357, opened 2025-07-30
Allow partners to limit sponsored data storage size

alinvetian:
Closes: https://github.com/datadryad/dryad-product-roadmap/issues/4302

**NOTE**: when deploying/testing this, please run `rails db:migrate:with_data`, do not run the migration commands separate

Refactored payment configuration:
- extract common fields into a new polymorphic model
- update related models
- update forms
- updated usage
- updated specs
- added payment plan selector for institutions


ryscher:
In my testing, I was unable to see the journal covering paymetns at all. My test data may be old and messy, but let's find a time to discuss.

ryscher:
Further complication, since the migrations are not reversible, now that this PR cannot be merged, I have to manually fix my database.

ryscher:
Bug: 
1. Create a dataset and upload some files.
2. Go to the agreements section and see the overage fee. 
3. Go back to the files section, and change the files so the dataset is in a different payment tier (higher or lower). 
4. When you go back to the agreements tab, the overage fee will be the same as before (incorrect). Refreshing the page will show the correct fee.

ahamelers (changes requested):
Looks good, except you need to fix the front end check in the React UI on the Connect screen, when the journal and a msid or DOI is added, for journal sponsorship. See the related failing tests.

PR #2366, opened 2025-08-06
Stop duplicates created with journal links

ahamelers:
Also, whenever someone is creating a dataset, they should start on the first submission system page even if the dataset has imported content.

Closes https://github.com/datadryad/dryad-product-roadmap/issues/4385


ahamelers:
> Maybe this will be better when the Title and Connect screens are separate?

It will, but that indicates the import_info isn't being set when manuscript info is imported like this. The new design doesn't care about that so much, but I'll fix it anyway.

PR #2373, opened 2025-08-07
Markdown editor packages and code update

ahamelers:
- Updated NPM packages
- Set up basic jest tests for the README editor and the markdown editor itself
- Moved button code around

(Getting ready to add more buttons)


alinvetian:
Subscription form is not loading, complains about not finding `@milkdown/kit/core` 
After replacing  `@milkdown/kit/core`  with `@milkdown/kit` the form loads.

Readme does not load 
image

PR #2400, opened 2025-08-19
Add additional system role

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/4420


ahamelers:
@alinvetian would this be sensible as they need to be different for user and proxy_user?

ahamelers:
I want us to reconsider what the heck proxyuser is for and maybe get rid of it anyway (see dev meeting notes)

PR #2434, opened 2025-09-03
Change partners ID

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/4470


ryscher:
This looks good, but I want to discuss whether we should leave an anchor with the old name in case people are using it.

ahamelers:
> This looks good, but I want to discuss whether we should leave an anchor with the old name in case people are using it.

Only one anchor will automatically open the section, so I don't know how valuable this would be.

PR #2448, opened 2025-09-09
Downgrade milkdown version

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/4484

I tried many versions until I got to the latest one where table pasting works. The only functionality that seems to be broken is the ability to use a button to select the entirety of either a table row or column. I have commented out these buttons, since I think if there must be a choice, the copying and pasting of tables is preferred.


ryscher:
This is weird. But the new version works correctly. Thanks for tracking it down.

PR #2475, opened 2025-09-18
Make manuscript number parsing consistent and stop duplicates

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/4522


ryscher:
Great. This tests fine with all of the combinations I can think of. Thanks!

PR #2896, opened 2026-04-14
Implement image uploads for the markdown editor

ahamelers:
From https://github.com/datadryad/dryad-product-roadmap/issues/4991


alinvetian:
Is there a way to limit the image size?

ahamelers:
@alinvetian certainly, we could put in a check for file sizes and add an error message if something is too large, but that wasn't requested in the original ticket, and I'd need to know what file size.

ahamelers:
Addition of image upload to the text editor needs further organizational discussion.

ahamelers:
This can be merged as it now creates the menu, but does not add the menu to the editor. Should be merged to prevent future merge conflicts.

PR #2916, opened 2026-05-01
Update UI display of sponsorship information

### LINKED ISSUE #4986: Expose more payment details on the relevant tables
Curators need to be able to look up whether sponsors cover large data fees, various limits, etc. Currently this information is only present in the editing mode for these accounts, which are privileged. Expose the information on the account tables themselves, which are visible to all system users.

ahamelers:
To account for new sponsorship inheritance. Also removes payment information from the journal level, as discussed.

Closes https://github.com/datadryad/dryad-product-roadmap/issues/4986
Closes https://github.com/datadryad/dryad-product-roadmap/issues/5013


ryscher:
I'm getting errors about missing method `JournalOrganization.payment_logs` -- I see this method defined for `Journal` and `Funder`, but not `JournalOrganization`.

app/services/payment_limits_service.rb
```diff
@@ -42,7 +42,7 @@ def exceeds_sponsor_yearly_limit?(storage_fee)
     return false if sponsor.nil?
     return false if payment_configuration&.yearly_ldf_limit.nil?
 
-    paid_amount = SponsoredPaymentLog.for_current_year.where(sponsor_id: sponsor.id).sum(:ldf)
+    paid_amount = SponsoredPaymentLog.for_current_year.where(payer: sponsor.payers_sponsored).sum(:ldf)
```
alinvetian: This change should not be done in here, instead:

- add a new method in class PayersService called `limits_sponsor` which should return `payment_sponsor` for tenants and `sponsor` for journals
- Use this method on the initializer to set the sponsor

The sponsor in here should be the one providing the limits.
This is currently a bug.

PR #2945, opened 2026-05-12
Fix flaky tests

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/5032


ahamelers:
We have a new error that is happening _everywhere_ on github:

```
Selenium::WebDriver::Error::UnknownError:
       unknown error: unhandled inspector error: {"code":-32000,"message":"Node with given id does not belong to the document"}
         (Session info: chrome=147.0.7727.55)
```

This seems to be due to the version of the chrome driver used by Selenium on github. The stable version is now 148 and locally with the upgraded driver this issue seems resolved. I can't see how to force a driver update on github; hopefully this happens automatically soon?

PR #2989, opened 2026-06-10
Payment debugging pages

alinvetian:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/5086

Please pay extra attention to the 3 `def sponsored_identifiers` methods and le me know if there is something that i missed


ryscher:
Some old test datasets get 
```
ActionView::Template::Error (undefined method `payment_configuration' for nil):
    20:   <br />
    21:   <div class="details-box">
    22:     <h3>Payment limits</h3>
    23:     <% if @limits_sponsor.payment_configuration&.payment_plan != '2025' %>
    24:       <p class="c-alert--error">Not 2025 payer!!!</p>
    25:     <% end %>
app/views/hiddens/identifier_payment_details.html.erb:23
```

ryscher (approved):

ryscher:
Fine to merge as-is, since it's for superusers only. But I'll let you decide if you want to fix the little error first.

app/models/stash_engine/tenant.rb
```diff
@@ -133,6 +133,10 @@ def payers_sponsored
       [self] + sponsored_deep
     end
 
+    def sponsored_identifiers
+      StashEngine::Identifier.joins(:resources).where(resources: { tenant_id: id }).distinct
```
ahamelers: This should also use the identifier.payment_id (not the resource.tenant_id). Funder sponsors supersede institutions, so it can't be assumed that if the resource.tenant_id matches an institution that they are also paying. Plus, what if the tenant_id changed on a later resource for the same identifier? This does not account for that.

app/models/stash_engine/journal.rb
```diff
@@ -197,6 +197,10 @@ def as_json(_options = {})
       )
     end
 
+    def sponsored_identifiers
+      StashEngine::Identifier.where("payment_type like 'journal-%'").where(payment_id: issn_ids).distinct
```
ahamelers: `issn_ids` has not been defined for this class, here you should be using `issn_array`

ahamelers (changes requested):

app/models/stash_engine/journal.rb
alinvetian: Done

app/models/stash_engine/tenant.rb
alinvetian: Done

ahamelers (approved):

PR #3020, opened 2026-06-25
Deleted dataset search

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/5063


app/policies/stash_engine/admin_datasets_policy.rb
```diff
@@ -36,5 +36,9 @@ def add_concern?
     def notification_date?
       @user.min_manager?
     end
+
+    def deleted?
+      @user.system_user? && @user.min_curator?
```
alinvetian, line 41: this translates to `@user.system_user?`

alinvetian (changes requested):

alinvetian (approved):

app/policies/stash_engine/admin_datasets_policy.rb
ahamelers: No, it's for system users who are at least curators (excluding tenant curators). I could write a new role check like:

```
roles.any? { |r| r.role_object_id.nil? && %w[superuser manager curator].include?(r.role) }
```

But that seemed superfluous when this combination basically does that.

PR #3029, opened 2026-07-02
Change LDF notification boolean to string and add for all sponsors

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/5107

### DIFF
--- a/spec/services/cost_reporting_service_spec.rb
+++ b/spec/services/cost_reporting_service_spec.rb
@@ -14,15 +14,15 @@ RSpec.shared_examples('sends ldf notification') do
 end
 
 describe CostReportingService do
-  let(:tenant) { create(:tenant, campus_contacts: ['some@email.com'].to_json) }
   let(:identifier) { create(:identifier) }
-  let(:org) { create(:journal_organization) }
-  let!(:journal) { create(:journal, sponsor: org) }
-  let!(:payment_conf) { create(:payment_configuration, partner: org, payment_plan: '2025', covers_ldf: true, ldf_limit_notification: true) }
+  let!(:tenant) { create(:tenant) }
+  let!(:payment_conf) do
+    create(:payment_configuration, partner: tenant, payment_plan: '2025', covers_dpc: true, covers_ldf: true,
+                                   ldf_limit_notification: 'some@email.com')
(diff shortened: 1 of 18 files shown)


spec/services/cost_reporting_service_spec.rb
```diff
@@ -74,24 +69,12 @@
...
 
         context 'when non 2025 plan institution is the payer' do
-          let!(:payment_conf) { create(:payment_configuration, partner: journal, payment_plan: 'TIERED') }
+          let!(:payment_conf) { create(:payment_configuration, partner: tenant, payment_plan: 'TIERED') }
           let!(:resource) { create(:resource, identifier: identifier, total_file_size: resource_files_size) }
 
           include_examples 'does not send ldf notification'
         end
 
-        context 'where there is no tenant' do
```
alinvetian, line 83: why did you remove these tests?

alinvetian (changes requested):
1. The new field is enabled by default on all forms, even if the covers LDF is not checked.
2. If you have a saved value on the new field, disable LDF and save, you get the error that old value is not valid email address

spec/services/cost_reporting_service_spec.rb
ahamelers: It doesn't matter if there's a tenant or not if the journal is the sponsor.

ahamelers: These tests and the function itself were initially set up very strangely, basically it seemed like they would send an LDF notification email to any tenant emails, regardless of whether the tenant was the sponsor. With these changes as well as changing the field to a string, the email is sent to the sponsor specifically regardless of type.

ahamelers:
> 1. The new field is enabled by default on all forms, even if the covers LDF is not checked.

There's just not a disabled style for textareas (I guess I can add one!)

ahamelers:
>2. If you have a saved value on the new field, disable LDF and save, you get the error that old value is not valid email address

This is because by disabling these fields with javascript, you actually make it so the key for each is not sent and the column is not updated in the database via the form. To remove the set values something different will need to be done!

alinvetian (approved):

PR #3033, opened 2026-07-03
Use earliest of the 3 dates for admin first submission date search

alinvetian:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/5137


alinvetian:
Does not work as expected and this is not a real bug

PR #3041, opened 2026-07-09
Index file types in compressed files for search

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/5141


alinvetian:
@ahamelers please keep in mind that we need to trigger the full reindexing after deploy

PR #3049, opened 2026-07-14
Remove sponsored payment logs on dataset withdrawal

alinvetian:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/5147


app/services/sponsored_payments_service.rb
```diff
@@ -48,6 +48,14 @@ def log_payment
     end
   end
 
+  def remove_logs
+    logs_to_delete = identifier.sponsored_payment_logs
+    published_activity = identifier.last_published_status
+    logs_to_delete = logs_to_delete.where(created_at: [published_activity.created_at..]) if published_activity.present?
```
ryscher: We no longer allow published datasets to be withdrawn, they must be retracted. So this line should never have an effect.

ryscher (approved):

app/services/sponsored_payments_service.rb
ryscher: @ahamelers Can you confirm my logic before merging this?

ahamelers: I will look at this, but in the meantime, more accurate publication date info:

Date first published: `identifier.publication_date`
Most recently published: `identifier.latest_resource_with_public_metadata.publication_date`
Curation activity (with note and everything) from first publication: `identifier.first_published_status`
Curation activity (with note and everything) from last publication: `identifier.latest_resource_with_public_metadata.first_published_status`

ahamelers: Okay, I've checked all the changes and @alinvetian , Ryan is correct. This line can be removed, as later versions of published datasets are not withdrawn, even by automatic abandoned datasets processes. We just ensured they cannot be withdrawn through the UI either.

ahamelers (changes requested):

ahamelers:
SponsoredPaymentLogs are already removed when the parent resource is 'destroyed', correct @alinvetian ?

app/services/sponsored_payments_service.rb
alinvetian: Can we withdraw a retracted dataset? of new versions of it?

alinvetian:
> SponsoredPaymentLogs are already removed when the parent resource is 'destroyed', correct @alinvetian ?

yes

app/services/sponsored_payments_service.rb
ahamelers: We should not be able to, no, it should have the same settings for that as published and embargoed wherever possible.

ahamelers (approved):

PR #3064, opened 2026-07-21
Check payment on manuscript acceptance

alinvetian:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/5183

### DIFF
--- a/app/models/stash_engine/manuscript.rb
+++ b/app/models/stash_engine/manuscript.rb
@@ -103,7 +103,6 @@ module StashEngine
         target_status = 'withdrawn'
         resource.resource_publication.update(publication_name: nil, publication_issn: nil)
         check_resource_payment(resource)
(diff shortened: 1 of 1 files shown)


ahamelers (approved):

ahamelers (changes requested):
I approved this and then set it back! I think we need to send an automatic email to users for this case, like the pubupdater case. Otherwise they will not be informed that the dataset is awaiting payment for a month.

app/models/stash_engine/manuscript.rb
ahamelers: This is the correct email to sent out just here, a few lines down! I feel like this is the wrong place for this entirely, actually; it should probably be sent only if we know the status is NOT ultimately going be 'withdrawn'.

Maybe remove the if statement here and move this email sending down after line 118, like:

```ruby
return unless target_status

StashEngine::UserMailer.peer_review_payment_needed(@new_res).deliver_now if target_status == 'awaiting_payment'
```

ahamelers (changes requested):

ahamelers (approved):

PR #3072, opened 2026-07-22
Fix resource paper trail versions duplication issue, add cleanup script

alinvetian:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/5104

The script will delete `330,528` resource records and `1,781` author records
I will run the script manually since it takes a while and did not want to keep the deploy waiting on data migrations


ryscher (approved):
Looks good to me, but always useful to have a second set of eyes when applying deletions.

ahamelers (approved):

PR #3117, opened 2026-08-05
5144 update stripe invoice flow

alinvetian:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/5144

### DIFF
--- a/app/controllers/payments_controller.rb
+++ b/app/controllers/payments_controller.rb
@@ -4,7 +4,7 @@ class PaymentsController < ApplicationController
   include StashEngine::SharedSecurityController
 
   skip_before_action :verify_authenticity_token
-  before_action :resource
+  before_action :resource, except: %i[invoice_callback]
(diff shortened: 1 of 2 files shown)


app/controllers/payments_controller.rb
```diff
@@ -64,6 +64,25 @@ def callback
     update_payment_details(payment)
   end
 
+  def invoice_callback
+    render json: {}, status: :ok and return unless params[:type] == 'invoice.paid'
+
+    invoice_id = params[:data][:object][:id]
+    payment = ResourcePayment.where(pay_with_invoice: true, invoice_id: invoice_id).last
+    if payment
+      payment.update(
```
ahamelers, line 73: Is there a reason not to use the existing `invoicer.mark_invoice_paid!` function here?

alinvetian: That one queries Stripe API again and pulls the info from there, while in here we already have the info coming form Stripe

PR #3142, opened 2026-08-25
Slightly better payment details pages

### LINKED ISSUE #5171: Reformat and open up payment details pages
Data managers are always asking questions that are answered in these payment details pages. The pages use all dev language and are very inaccessible and missing lots of functionality that would make them more useful. They need fixing up if they are to be shown to other admins.

ahamelers:
Still not beautiful or anything, but open to admins, accessible, and with more features

Closes https://github.com/datadryad/dryad-product-roadmap/issues/5171


alinvetian:
When there is no sponsor, the page looks bad
image

PR #3164, opened 2026-09-07
Add multiple Stripe callbacks handling

### LINKED ISSUE #5238: Update ResourcePayment status for voided/refunded payments
For tracking payment history accurately, we need to update ResourcePayment statuses for (permanently) voided invoices and for refunded payments.

The `invoice_paid?` method checks for voided invoices with replacement revisions and updates the invoice_id, but if any invoices are voided with _no_ replacement, we should update the ResourcePayment status to 'void'

If a non-invoice payment is refunded, we should also update that status to 'void' or 'refunded' (whatever matches the stripe payment status)

These should be done going forward, with callbacks etc., and these statuses also need to be checked for ResourcePayments where these statuses may have already been updated.

alinvetian:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/5238


ahamelers:
@alinvetian, this looks good, but one method of invoice voiding on the stripe side is not a permanent marker of something not being paid, but immediately replaces one invoice with another. It's not associated with a new resource or resource_payment on our end, but is an action done entirely on the stripe side. So, I would expect the voided invoices handler to check for a replacement invoice ID (`latest_revision` in the stripe API), and then update the resource_payment invoice ID and status to match the new one if one is found, rather than marking the status immediately voided (we have no other way to find and update invoice IDs with their replacements—unless you can also add a callback for that!)

ahamelers:
@alinvetian Also, this looks great for updating future payment status changes using webhooks, but how will we check and update the status for everything that may have been voided or refunded before this is released? Will there be a script that also hooks into these handlers, or will it be entirely separate?

alinvetian:
@ahamelers the code is updated now to handle voided invoice revisions.
I also added 2 rake tasks to check historical refunded/voided payments.

PR #3167, opened 2026-09-08
Save retrievable fee calculator receipt record

### LINKED ISSUE #5165: Payment tracking: Submitter receipt
Starting with the information on the "payment details" page, make a page that provides easy access to the information about what portion of payment the Sponsor is covering, and what the Submitter is responsible for. This should appear both in the submission system and as a link from the My datasets page. This information should be viewable by Submitters, Partners, and Dryad staff.

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/5165

Saves a fee calculator record per resource, and marks it as a receipt when paid. Displayed on the user datasets page and in the DPC paid by column of the admin dashboard (for now)

### DIFF
--- a/app/controllers/fee_calculator_controller.rb
+++ b/app/controllers/fee_calculator_controller.rb
@@ -17,10 +17,13 @@ class FeeCalculatorController < ApplicationController
 
   def calculate_resource_fee
     resource = StashEngine::Resource.find(params[:id])
+    fees = ResourceFeeCalculatorService.new(resource).calculate(resource_options)
+    record = resource.fee_record || resource.build_fee_record
+    record.update(fees: fees.to_json)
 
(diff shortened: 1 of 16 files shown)


ryscher:
I created a new dataset on dev, and paid with a test card. The receipt had two problems:
1. "Dryad fees" are shown as 0, even though the total is correct
2. "Submitter total" has extra characters appended.
Screenshot 2026-09-08 at 5 58 55 PM

ryscher:
Would it be better to omit the Large Data Fee column when it does not contain any information?

ahamelers:
@ryscher that was my intention, but I forgot about the heading row 🤦

This table was originally created to display only to sponsored users

PR #3176, opened 2026-09-15
Integration of ScholarOne API

### LINKED ISSUE #5240: ScholarOne integration testing
Test the APIs and tools in ScholarOne, as described in [ScholarOne dev notes](https://docs.google.com/document/d/1tXm_M6CztzTooixpRw1yGlx1eyNpaDDXpbPMmt1pXbk/edit?usp=sharing). Propose a method for journal integration using these APIs.

alinvetian:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/5240


lib/integrations/scholar_one.rb
```diff
@@ -0,0 +1,46 @@
+module Integrations
+  class ScholarOne
+    BASE_URL = APP_CONFIG.scholar_one.api_base_url
+    SITE_NAME = 'supportdemoplus'.freeze
```
ryscher, line 4: This will be different for each journal, so I don't think we want to define it as a constant.

ryscher (changes requested):

PR #3186, opened 2026-09-18
Add publication status field and filter to admin dashboard

### LINKED ISSUE #5267: Published dataset filter for admin dashboard
Currently, to retrieve datasets that have ever been published, admins must do something like setting the First Published filter to a far past date. Retrieving never-published datasets is much tricker. Both of these ways of limiting are desired for better reporting. Add something like a pub_state filter, so users can quickly retrieve only published or only unpublished datasets.

ahamelers:
Closes https://github.com/datadryad/dryad-product-roadmap/issues/5267


app/models/stash_engine/identifier.rb
```diff
@@ -79,14 +79,7 @@ class Identifier < ApplicationRecord
...
 
-    # This makes the setting of the "preliminary information" and how something was imported explicit.  Default is other.
-    # There is only one import per dataset and it overwrites info after getting valid information.
-    enum :import_info, {
-      other: 0,
-      manuscript: 1,
-      published: 2,
-      preprint: 3
-    }
+    enum :pub_state, { unpublished: 'unpublished', published: 'published', embargoed: 'embargoed', retracted: 'retracted', withdrawn: 'withdrawn' }
```
alinvetian: This should look like:
```
enum :pub_state, %w[unpublished published embargoed retracted withdrawn].index_by(&:to_sym)
```

alinvetian (approved):

